PR #13, opened 2021-08-17
SCRUM-132 load disease annotation and create endpoint to retrieve all…

cmpich:
… DAs


src/main/java/org/alliancegenome/curation_api/base/BaseAnnotation.java
```diff
@@ -0,0 +1,33 @@
...
+	@CreationTimestamp
+	@JsonView({View.FieldsOnly.class})
+	private LocalDateTime created;
+
+	@GenericField
+	@UpdateTimestamp
+	@JsonView({View.FieldsOnly.class})
+	private LocalDateTime lastUpdated;
+
+}
```
oblodgett, line 33: Do we need this class considering Association Extends BaseGeneratedEntity which is the same as this class?

oblodgett: Plus this class is not used?

src/main/java/org/alliancegenome/curation_api/bulk/controllers/DiseaseAnnotationBulkController.java
```diff
@@ -0,0 +1,66 @@
...
+            if (gene != null) {
+                g.setSubject(gene);
+                g.setObject(disease);
+                Reference publication = new Reference();
+                publication.setCurie(annotation.getEvidence().getPublication().getPublicationId());
+                if (referenceService.get(publication.getCurie()) == null) {
+                    referenceService.create(publication);
+                }
+                g.setReferenceList(List.of(publication));
+                diseaseService.create(g);
```
oblodgett, line 58: What happens the next time we run the annotations file? Its just going to create another set... so far we have been doing incremental updates.

oblodgett: Also this code should be moved to the service... I know the examples are otherwise but in the latest code these have been moved.

oblodgett (approved):

PR #18, opened 2021-08-18
Refactored DTO's and added FMS Component and AGM bulk import

oblodgett:


cmpich (commented):
Why did you add an AffectedGenomicModel loader?

adamgibs (approved):

oblodgett:
I wouldn't call this story done, but got most of the infrastructure code done.

PR #25, opened 2021-08-31
Added fix for secondary ids

oblodgett:
PS The database doesn't have any right now so this code checks and doesn't import the new ones.


abecerra (commented):
I used a set to avoid duplication of sec Ids in the case of data reload

oblodgett:
Yes but in this case we are setting the list each time so there will be no duplicates just what we are getting from the file.

PR #27, opened 2021-09-01
Handle no synonyms in DB and no Synonyms in incoming stream for BGI

cmpich:


src/main/java/org/alliancegenome/curation_api/base/BaseSQLDAO.java
```diff
@@ -47,6 +47,18 @@ public E find(String id) {
...
+        if(id != null) {
+            E entity = entityManager.find(myClass, id);
+            log.debug("Entity Found: " + entity);
+            return entity;
+        } else {
+            log.debug("Input Param is null: " + id);
+            return null;
+        }
+    }
+
```
oblodgett, line 61: @cmpich good catch I didn't update these after we made the GeneratedCurie base class

src/main/java/org/alliancegenome/curation_api/base/BaseSQLDAO.java
```diff
@@ -84,6 +96,13 @@ public E remove(String id) {
         return entity;
     }
 
+    public E remove(Long id) {
+        log.debug("SqliteDAO: remove: " + id);
+        E entity = find(id);
+        entityManager.remove(entity);
+        return entity;
+    }
+
```
oblodgett, line 105: @cmpich good catch I didn't update these after we made the GeneratedCurie base class

src/main/java/org/alliancegenome/curation_api/base/BaseService.java
```diff
@@ -38,6 +38,11 @@ public E delete(String id) {
         return dao.remove(id);
     }
 
+    @Transactional
+    public E delete(Long id) {
+        return dao.remove(id);
+    }
+
```
oblodgett, line 45: @cmpich good catch I didn't update these after we made the GeneratedCurie base class

oblodgett (approved):

PR #54, opened 2021-09-23
SCRUM-321

adamgibs:


src/main/java/org/alliancegenome/curation_api/model/ingest/json/dto/ExperimentalConditionDTO.java
```diff
@@ -13,6 +14,8 @@
     private String conditionQuantity;
     private String anatomicalOntologyId;
     private String geneOntologyId;
+
+    @JsonProperty("NCBITaxonId")
```
adamgibs, line 18: This was added because I kept running into an issue while loading ZFIN DAF files. I would get a message "Unrecognized field "NCBITaxonId"; (class org.alliancegenome.curation_api.model.ingest.json.dto.ExperimentalConditionDTO), not marked as ignorable". This seemed to fix it.

oblodgett: This is only an issue with the Zfin file. We have a new Zfin file where we won't need this... I would suggest not added it.

cmpich: I should give you the most recent file, Adam.

oblodgett (approved):

PR #59, opened 2021-09-28
SCRUM-321 modified DAComponent null check defaulted null values to fa…

adamgibs:
…lse in DB

I've also removed the search bars from creation date and negation since currently they are not searchable.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/DiseaseAnnotationCurie.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/DiseaseAnnotationCurie.java
@@ -24,7 +24,13 @@ public abstract class DiseaseAnnotationCurie {
 
     public String getEvidenceCurie(EvidenceDTO dto) {
         CurieGenerator curie = new CurieGenerator();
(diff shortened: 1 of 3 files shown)


src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/DiseaseAnnotationCurie.java
```diff
@@ -24,7 +24,13 @@ public String getExperimentalConditionCurie(ExperimentalConditionDTO dto) {
...
     public String getEvidenceCurie(EvidenceDTO dto) {
         CurieGenerator curie = new CurieGenerator();
-        curie.add(dto.getPublication().getCrossReference().getCurie());
+
+        if(dto.getPublication().getCrossReference() != null){
+            curie.add(dto.getPublication().getCrossReference().getCurie());
+        }else{
+            curie.add(dto.getPublication().getPublicationId());
+        }
+
```
oblodgett, line 33: @cmpich the publication -> crossreference is optional. Not all mods have those so the logic is use the ModId if it exists otherwise use the PubMedId.

oblodgett (approved):

PR #64, opened 2021-10-04
Added Biological Entity Endpoints and Refactored / Reorgged some of the code.

oblodgett:


src/main/java/org/alliancegenome/curation_api/services/GeneService.java
```diff
@@ -156,15 +157,15 @@ private void handleCrossReferences(GeneDTO geneDTO, Gene gene) {
     
     private void handleNewSynonyms(GeneDTO gene, Gene g) {
         if (CollectionUtils.isNotEmpty(gene.getBasicGeneticEntity().getSynonyms())) {
-            List<Synonym> synonyms = DtoConverter.getSynonyms(gene);
+            List<Synonym> synonyms = DtoConverterHelper.getSynonyms(gene);
```
abecerra, line 160: Could you think in a more specific name for this class DtoConverterHelper? 
Or maybe just make it a method in the service?

oblodgett: This code already existed. I was just changing the name to indicate its a helper vs a service.

abecerra (approved):

PR #70, opened 2021-10-06
SCRUM-475

adamgibs:
Do not merge! Draft is here for code sharing purposes. Code is not ready for review.


src/main/cliapp/src/service/DiseaseAnnotationService.js
```diff
@@ -1,6 +1,10 @@
 import axios from 'axios';
 
 export class DiseaseAnnotationService {
+
+    saveDiseaseAnnotation(newAnnotation){
+        return axios.put(`api/disease-annotation`, newAnnotation)
+    }
```
oblodgett, line 7: Sorry to be picky... but wouldn't name this variable "newAnnotation" as its not "new" vs a post which would be "new"

PR #83, opened 2021-10-11
SCRUM-459 release display

mluypaert:
I've made a few updates to the Makefile which cause the pom.xml file included in the compiled API jar file to contain the version as returned by `git describe`. This way the alpha environment will more accurately display/return the code version deployed to it (while it should make no difference for beta or production).

### DIFF
--- a/src/main/cliapp/src/components/DiseaseAnnotationsComponent.js
+++ b/src/main/cliapp/src/components/DiseaseAnnotationsComponent.js
@@ -4,7 +4,7 @@ import {Column} from 'primereact/column';
 import {Button} from 'primereact/button';
 import {DiseaseAnnotationService} from '../service/DiseaseAnnotationService'
 import {useMutation, useQuery} from 'react-query';
-import {Messages} from "primereact/messages";
+import {Message} from "primereact/message";
 import {AutoComplete} from "primereact/autocomplete";
 import {BiologicalEntityService} from "../service/BiologicalEntityService";
 import { Toast } from 'primereact/toast';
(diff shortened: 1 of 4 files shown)


mluypaert:
> 
> 
> @mluypaert I guess my mind was going in a very different idea of the this where we would fix: `.github/workflows/post-merge-image-build.yml:78` instead of "alpha" would be `git describe` and then inside java we could pull this variable and display it via the: `src/main/java/org/alliancegenome/curation_api/crud/controllers/APIVersionInfoController.java` we could add a another variable:
> 
> ```java
>    @ConfigProperty(name = "agr.curation.release")
>    String curationReleaseVersion;
> ```
> 
> This would all achieve the same goal while not having to deal with maven commands.

@oblodgett While I agree that the main branch images should be tagged with the `git describe` output rather than just "alpha" (I've now made the necessary code changes for that), I'm not so convinced on using an environment variable for version reporting. As the code version to report in an inherent part of the code that is being compiled, it makes more sense to me to bake the version into the compiled jar than to provide it through an environment variable at runtime, as baking it in makes the jar always report the same (correct) version regardless of its deployment. So unless you want to start to re-use images built at main for (pre-)release deployment (in which case the release to report is not known when compiling the jar and building the image), and thus you want to stop using `mvn release`, I don't see the advantage of using environment variables over baking it in.

oblodgett:
Ya, the issue is if the `mvn ${OPTS}` fails it will not run the following command of `mvn versions:revert` which will leave the developer a change to the pom file unawares. I don't know what a second `mvn versions:set -DnewVersion=${GIT_VERSION}` would do, but might fix the change once the `mvn ${OPTS}` passes. To me it leaves too much room for error on the part of someone that doesn't know it's happening. But pulling them out into there own targets is fine.

PR #88, opened 2021-10-14
Added ui start proxy options

adamgibs:
Making a draft pull request for feedback (probably won't see until Monday). This seems to work on my local system. I haven't worked with make files much so I'm sure there is a better way to do this.

### DIFF
--- a/src/main/cliapp/Makefile
+++ b/src/main/cliapp/Makefile
@@ -4,6 +4,12 @@ all:
 run:
 	npm start
 
+run-alpha:
+	npm run start-alpha
+
+run-local:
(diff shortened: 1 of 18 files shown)


src/main/cliapp/Makefile
```diff
@@ -4,6 +4,12 @@ all:
 run:
 	npm start
 
+run-alpha:
+	npm run start-alpha
+
+run-local:
+	npm run start-local
+
```
oblodgett, line 12: @adamgibs probably want to remove both these lines also as the scripts are now removed.

PR #101, opened 2021-10-20
Fix lazy serialization via JsonView (SCRUM-493)

oblodgett:


cmpich (commented):
Looks good, though, most changes are the import statements!

PR #117, opened 2021-11-09
SCRUM-601 Removing leading & trailing extra spaces from input string

kthorat-prog:


src/main/cliapp/src/components/DiseaseAnnotationsComponent.js
```diff
@@ -163,6 +165,7 @@ export const DiseaseAnnotationsComponent = () => {
             itemTemplate={subjectItemTemplate}
             completeMethod={searchSubject}
             onChange={(e) => onSubjectEditorValueChange(props, e)}
+            forceSelection={true}
```
adamgibs, line 168: Setting forceSelection as a property leads to a bug where if you tab away from an input, it causes the cell to go blank.

adamgibs (commented):
This didn't seem to work for me. When I paste in an item with whitespace and try to submit, it still get an error. Maybe it also needs to be trimmed somewhere in the onRowEditSave function?

src/main/cliapp/src/components/DiseaseAnnotationsComponent.js
kthorat-prog: Ohh yes, I accidently added that while fixing other issue. Only the .replace lines (2 times) is the change.

kthorat-prog:
> This didn't seem to work for me. When I paste in an item with whitespace and try to submit, it still get an error. Maybe it also needs to be trimmed somewhere in the onRowEditSave function?

adamgibs:
I just get the "not a valid entry" message when I try to submit after pasting.

kthorat-prog:
Ohh...I forgot to take care of that one... Thanks. Will fix it and then create a PR.

PR #118, opened 2021-11-12
Initial commit of Molecule entity class

markquintontulloch:
I'm unsure of how to test this, so it is untested but I thought I'd create the PR anyway so you can see if I've done anything obviously wrong!


cmpich (commented):
From the setup I don't see anything jumping out. However, you should abide by the usual conventions for indentation of lines: 4 white spaces per level. Lines 54 and 59 should be indented just like the line above it. It's the same level of scope.

oblodgett (approved):

PR #122, opened 2021-11-16
SCRUM-788  and SCRUM-789 molecule api

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/entities/Synonym.java
```diff
@@ -17,7 +17,7 @@
 @Data
 @NoArgsConstructor
 @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
-@ToString(exclude = {"genomicEntities"})
+@ToString(exclude = {"genomicEntities", "molecules"})
```
oblodgett, line 20: Might want to remove this considering the field isn't there anymore.

src/main/java/org/alliancegenome/curation_api/model/entities/Molecule.java
```diff
@@ -47,12 +49,14 @@
...
+	@Column(columnDefinition="TEXT")
 	private String smiles;
 	
-	@ManyToMany
+	@ElementCollection
+	@JsonView(View.FieldsAndLists.class)
 	@JoinTable(indexes = @Index( columnList = "molecule_curie"))
-	@JsonView({View.FieldsAndLists.class})
-	private List<Synonym> synonyms;
+	@Column(columnDefinition="TEXT")
```
oblodgett, line 58: These `@Column(columnDefinition="TEXT")` are only needed if the field is going to be longer than 255 charactors, otherwise it's better to leave it out.

src/main/java/org/alliancegenome/curation_api/services/helpers/DtoConverterHelper.java
```diff
@@ -19,4 +20,15 @@
                     return syn;
                 }).collect(toList());
     }
+    
+    public static List<Synonym> getSynonyms(MoleculeDTO molecule) {
```
oblodgett, line 24: This method is not needed if you are going to use the `List<String>` for synonyms?

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/model/entities/Molecule.java
markquintontulloch: Yeah, I had removed that but there are indeed synonyms > 255 chars

src/main/java/org/alliancegenome/curation_api/model/entities/Synonym.java
markquintontulloch: removed

src/main/java/org/alliancegenome/curation_api/services/helpers/DtoConverterHelper.java
markquintontulloch: removed

oblodgett (approved):

PR #125, opened 2021-11-18
Validate name field is present and not empty

markquintontulloch:
Added quarkus-hibernate-validator dependency to pom.xml and added @NotNull and @NotEmpty constraints to name field of molecule class.


markquintontulloch:
Changing how required fields are dealt with so response codes will change, closing PR.

PR #126, opened 2021-11-18
Add molecule consumer

markquintontulloch:
Also contains changes from #125 as I forgot to switch back to master before branching.

Molecule bulk upload to alpha was timing out (there are more molecules that I thought!), so reimplemented consumer.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -82,6 +82,10 @@
 		</dependency>
 		<dependency>
 			<groupId>io.quarkus</groupId>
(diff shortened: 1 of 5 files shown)


oblodgett (changes requested):
Didn't see this commit... we don't need the Consumer for this endpoint

markquintontulloch:
Timeout not caused by size of file and resubmission works, so abandoning changes.

PR #127, opened 2021-11-18
Initial commit of bulk upload tests for molecule

markquintontulloch:
!!! Not sure how to actually run these tests, so haven't - can someone advise before merging!!!

Tests for missing id and name should (I believe) return a 400 error code, but swagger UI shows 500 - quarkus terminal displays correct error message from @NotNull and @NotEmpty annotations.

### DIFF
--- a/src/test/java/org/alliancegenome/curation_api/bulkupload/MoleculeBulkUploadITCase.java
+++ b/src/test/java/org/alliancegenome/curation_api/bulkupload/MoleculeBulkUploadITCase.java
@@ -1,0 +1,387 @@
+package org.alliancegenome.curation_api.bulkupload;
+
+import java.io.IOException;
+import java.nio.file.*;
+
+import org.alliancegenome.curation_api.resources.TestElasticSearchResource;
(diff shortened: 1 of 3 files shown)


src/test/java/org/alliancegenome/curation_api/bulkupload/MoleculeBulkUploadITCase.java
```diff
@@ -0,0 +1,387 @@
...
+    @Order(1)
+    public void moleculeBulkUploadCheckModExamples() throws Exception {
+        String content = Files.readString(Path.of("src/test/resources/bulk/05_molecule/00_mod_examples.json"));
+
+        // upload file
+        RestAssured.given().
+            contentType("application/json").
+            body(content).
+            when().
+            post("/api/molecule/bulk/moleculefile?async=false").
```
oblodgett, line 42: I believe we removed the async param in the API so the `?async=false` could be removed from all these URL's?

oblodgett (changes requested):

PR #128, opened 2021-11-19
Molecule constraints and testing

markquintontulloch:
Added logic to molecule service to skip entries with no/empty name field.

Integration tests for molecules added and passing.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/base/BaseCrudRESTInterface.java
+++ b/src/main/java/org/alliancegenome/curation_api/base/BaseCrudRESTInterface.java
@@ -37,7 +37,7 @@ public interface BaseCrudRESTInterface<E extends BaseEntity> {
 
     @POST
     @Path("/find")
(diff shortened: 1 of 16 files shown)


abecerra (approved):

src/main/java/org/alliancegenome/curation_api/base/BaseCrudRESTInterface.java
```diff
@@ -37,7 +37,7 @@
 
     @POST
     @Path("/find")
-    @JsonView(View.FieldsOnly.class)
+    @JsonView(View.FieldsAndLists.class)
```
oblodgett, line 40: This may be problematic in the future. For objects that references back and forth with each other.

oblodgett: It would be better to override this method in the interface and change the JsonView there.

PR #130, opened 2021-11-22
Scrum 600 - Added Analyzer and sort normalizer to all fields for all datatables.

kthorat-prog:


oblodgett (approved):
This change is going to require the indexes to be regenerated. So lets hold off on bringing the code in until we can coordinate that.

kthorat-prog:
sure. We can discuss in Scrum Meet today.

PR #133, opened 2021-11-23
SCRUM-280 create integration tests for DA

cmpich:


src/main/java/org/alliancegenome/curation_api/interfaces/rest/DiseaseAnnotationRESTInterfaceRescu.java
```diff
@@ -0,0 +1,58 @@
+package org.alliancegenome.curation_api.interfaces.rest;
```
oblodgett, line 1: Was this class suppose to be a ZFIN only class?

cmpich: I am using the rescu library to retrieve disease annotations. For that I also need to create this interface as we discussed a few days ago.

abecerra: @cmpich in the following line there is an alternative way to fetch a disease annotation without a library:
https://github.com/alliance-genome/agr_curation/blob/c0a5df06d20287f3c18660e65adada63e0b1dd0e/src/test/java/org/alliancegenome/curation_api/crud/controllers/DiseaseAnnotationITCase.java#L124

It depends on:

https://github.com/alliance-genome/agr_curation/blob/c0a5df06d20287f3c18660e65adada63e0b1dd0e/src/test/java/org/alliancegenome/curation_api/crud/controllers/DiseaseAnnotationITCase.java#L32

cmpich: Good point, Andres.
I do want to entertain a Interface that can be used with rescu so MODs can use the interface directly to call our endpoints.

oblodgett: I guess my question is why is the DA's failing with the time issue when the other tests don't? I would rather not use another library to pull items.

PR #138, opened 2021-11-26
Disease annotation bulk integration tests

markquintontulloch:
A whole bunch of changes related to getting bulk loads of AGMs to work properly, with validation of required fields, and integration tests.


abecerra (approved):

src/main/java/org/alliancegenome/curation_api/model/ingest/json/dto/ExperimentalConditionDTO.java
```diff
@@ -13,7 +15,8 @@
     private String conditionQuantity;
     private String anatomicalOntologyId;
     private String geneOntologyId;
-    private String ncbitaxonId;
+    @JsonProperty("NCBITaxonId")
```
oblodgett, line 18: I don't think this works for every mod? This works only for ZFIN and ZFIN has produced a new file so that we don't need the json property? @cmpich ?

markquintontulloch: Only ZFIN is submitting that field, but it is in the capitalised form in the frozen JSON schema and the latest file was failing to load.

PR #139, opened 2021-11-29
Add 'with' field to disease annotation load

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
```diff
@@ -78,6 +80,15 @@ private DiseaseAnnotation upsertAnnotation(DiseaseModelAnnotationDTO annotationD
         }
         annotation.setNegated(annotationDTO.getNegation() == DiseaseModelAnnotationDTO.Negation.not);
 
+        if (CollectionUtils.isNotEmpty(annotationDTO.getWith())) {
+            List <Gene> withGenes = new ArrayList<>();
+            annotationDTO.getWith().forEach(with -> {
+                Gene withGene = geneDAO.find(with);
```
oblodgett, line 86: find() is ment to work with the primary key on the table. So in this case with.getCurie().

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
markquintontulloch: `with` is just a string, so I've switched it to `geneDAO.getByIdOrCurie(with)`

oblodgett: I'm sorry.. I didn't catch that... the getWith from the DTO is a string... ok.

oblodgett (approved):

PR #151, opened 2021-12-01
SCRUM-868 load zfa ontology

cmpich:


src/main/cliapp/src/components/Dashboard.js
```diff
@@ -52,6 +53,10 @@ export const Dashboard = () => {
           setMACount(results.totalResults);
         });
 
+        searchService.search('zfaTerm', 0, 0).then(results => {
```
oblodgett, line 56: All the other endpoints this is lowercase can we make it `zfaterm` to be consistant with the rest of them?

src/main/cliapp/src/components/ontologies/ZFAOntologyComponent.js
```diff
@@ -0,0 +1,85 @@
...
+  const [page, setPage] = useState(0);
+  const [first, setFirst] = useState(0);
+  const [rows, setRows] = useState(50);
+  const [totalRecords, setTotalRecords] = useState(0);
+  const [isEnabled, setIsEnabled] = useState(true);
+  const searchService = new SearchService();
+  const errorMessage = useRef(null);
+
+  useQuery(['terms', rows, page, multiSortMeta, filters],
+    () => searchService.search('zfaTerm', rows, page, multiSortMeta, filters), {
```
oblodgett, line 25: All the other endpoints this is lowercase can we make it `zfaterm` to be consistant with the rest of them?

src/main/java/org/alliancegenome/curation_api/interfaces/crud/ontology/ZfaTermRESTInterface.java
```diff
@@ -0,0 +1,18 @@
...
+import org.alliancegenome.curation_api.base.BaseCrudRESTInterface;
+import org.alliancegenome.curation_api.model.entities.ontology.ZfaTerm;
+import org.eclipse.microprofile.openapi.annotations.tags.Tag;
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("/zfaTerm")
```
oblodgett, line 12: All the other endpoints this is lowercase can we make it `zfaterm` to be consistant with the rest of them?

src/main/java/org/alliancegenome/curation_api/interfaces/bulk/ontology/ZfaTermBulkRESTInterface.java
```diff
@@ -0,0 +1,17 @@
+package org.alliancegenome.curation_api.interfaces.bulk.ontology;
+
+import org.alliancegenome.curation_api.base.BaseOntologyTermBulkRESTInterface;
+import org.eclipse.microprofile.openapi.annotations.tags.Tag;
+
+import javax.ws.rs.Path;
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+@Path("/zfaTerm/bulk")
```
oblodgett, line 10: All the other endpoints this is lowercase can we make it `zfaterm` to be consistant with the rest of them?

src/main/cliapp/package.json
```diff
@@ -27,8 +28,7 @@
         "react-query": "^3.19.6",
         "react-router-dom": "^5.2.0",
         "react-scripts": "4.0.0",
-        "react-transition-group": "^4.4.1",
-        "http-proxy-middleware": "^2.0.1"
+        "react-transition-group": "^4.4.1"
```
oblodgett, line 31: Also we try not to commit this file unless making changes to the UI.

oblodgett (changes requested):

src/main/cliapp/src/components/Dashboard.js
cmpich: I'd rather change all the other endpoints to use camel case as it is easier to read.

src/main/cliapp/package.json
cmpich: Sure. I don't know why this file got mixed around...

src/main/cliapp/src/components/Dashboard.js
oblodgett: https://datatracker.ietf.org/doc/html/rfc3986

"Although schemes are case-
   insensitive, the canonical form is lowercase and documents that
   specify schemes must do so with lowercase letters.  An implementation
   should accept uppercase letters as equivalent to lowercase in scheme
   names (e.g., allow "HTTP" as well as "http") for the sake of
   robustness but should only produce lowercase scheme names for
   consistency."

oblodgett: I would be fine with: `term/mp`, `mp_term`, or `mp-term`... just not uppercase in URL's.

cmpich: Ok, I'll change them all into a schema:
term/zfa
term/ma
... 
Will do that in a subsequent PR. Please let this one then go through for testing purposes.

oblodgett: So can we go the other direction, as in, lowercase this. So that its consistent with ALL the other ontology endpoints. And add the other PR later.

oblodgett (approved):

PR #152, opened 2021-12-01
SCRUM-491: edit disease annotation - integration test

abecerra:
Done without getting all the disease annotations in the DB to avoid running order execution issues with bulk upload integration tests merged by Mark.


markquintontulloch (approved):
Looks good to me

PR #155, opened 2021-12-02
Ontology terms without ids (SCRUM-101)

markquintontulloch:
Allows ZECO terms to be loaded.
ZECO UI components added.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
@@ -178,11 +178,11 @@ public class GenericOntologyLoadHelper<T extends OntologyTerm> implements OWLObj
 
         T term = clazz.getDeclaredConstructor().newInstance();
         term.setObsolete(false);
-
+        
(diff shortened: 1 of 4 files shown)


src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
```diff
@@ -239,6 +239,10 @@ else if(key.equals("inSubset")) {
             }
 
         });
+        
+        if (EntitySearcher.getAnnotationObjects(node, ontology).count() > 0 && term.getCurie() == null) {
```
oblodgett, line 243: I would swap the logic on this put the `term.getCurie() == null` first that way it will only check the other condition when terms don't have a curie. (Maybe more efficient)

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
markquintontulloch: done

PR #156, opened 2021-12-03
SCRUM-897 GO ontology

markquintontulloch:
This includes code to allow multiple namespaces to be specified in the config and loaded in a single load into a single table.  I think this is what we want to do for GO (at least for now), but if not I can retract this PR and push just the code for multiple namespaces (I think this will be useful for other ontologies).

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
@@ -91,13 +91,15 @@ public class GenericOntologyLoadHelper<T extends OntologyTerm> implements OWLObj
 
             termParent = getOntologyTerm(parent);
 
-            String requiredNamespace = config.getAltNameSpace();
-            if (requiredNamespace == null) {
-                requiredNamespace = defaultNamespace;
+            ArrayList<String> requiredNamespaces = config.getAltNameSpaces();
+            if (requiredNamespaces.isEmpty()) {
+                if (defaultNamespace != null) {
+                    requiredNamespaces.add(defaultNamespace);
+                }
             }
             
             if(
-                (termParent.getNamespace() != null && requiredNamespace != null && termParent.getNamespace().equals(requiredNamespace)) ||
+                (termParent.getNamespace() != null && !requiredNamespaces.isEmpty() && requiredNamespaces.contains(termParent.getNamespace())) ||
                 config.isLoadWithoutDefaultNameSpace()
             ) {
                 //System.out.println(termParent);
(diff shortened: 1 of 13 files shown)


src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
```diff
@@ -91,13 +91,15 @@ public T traverse(OWLClass parent, int depth) throws Exception {
...
             termParent = getOntologyTerm(parent);
 
-            String requiredNamespace = config.getAltNameSpace();
-            if (requiredNamespace == null) {
-                requiredNamespace = defaultNamespace;
+            ArrayList<String> requiredNamespaces = config.getAltNameSpaces();
+            if (requiredNamespaces.isEmpty()) {
+                if (defaultNamespace != null) {
+                    requiredNamespaces.add(defaultNamespace);
+                }
```
oblodgett, line 98: This block should happen once outside the traverse method, as it will get called on every node. Should go around line 69 or so.

This is an interesting idea with the list of alt name spaces. I just wonder if it could get confusing, where like in the ZFA file there are two ontologies (ZFA, ZFS), and two namespaces, vs a single ontology with multiple namespaces (GO). I believe this will work, I guess lets see what happens.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
```diff
@@ -91,13 +91,15 @@ public T traverse(OWLClass parent, int depth) throws Exception {
...
             }
             
             if(
-                (termParent.getNamespace() != null && requiredNamespace != null && termParent.getNamespace().equals(requiredNamespace)) ||
+                (termParent.getNamespace() != null && !requiredNamespaces.isEmpty() && requiredNamespaces.contains(termParent.getNamespace())) ||
```
oblodgett, line 102: Also when using a list you wouldn't need to do the `!requiredNamespaces.isEmpty()` check as the contains on an empty list will work correctly.

markquintontulloch: done

src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
markquintontulloch: done

oblodgett (approved):

PR #162, opened 2021-12-06
SCRUM-881 edit evidence codes

markquintontulloch:
Existing obsolete evidence codes can be kept in an annotation being updated, but annotations cannot be updated or created with obsolete evidence codes that were not pre-existing in the annotation.  Bulk upload unaffected (i.e. entries can be created with obsolete evidence codes).  Auto-suggest limited to non-obsolete terms.

### DIFF
--- a/src/main/cliapp/src/components/EvidenceEditor.js
+++ b/src/main/cliapp/src/components/EvidenceEditor.js
@@ -1,0 +1,52 @@
+import React, {useState} from 'react';
+import {AutoComplete} from "primereact/autocomplete";
+
+   export const EvidenceEditor = ({ rowProps, searchService, setDiseaseAnnotations, autocompleteFields }) => { 
+        const [filteredEvidenceCodes, setFilteredEvidenceCodes] = useState([]);
+
+        const searchEvidenceCodes = (event) => {
+            console.log(event);
+            let evidenceFilter = {};
+            autocompleteFields.forEach( field => {
+                evidenceFilter[field] = event.query;
+            });
+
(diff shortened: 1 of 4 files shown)


src/main/cliapp/src/components/EvidenceEditor.js
```diff
@@ -0,0 +1,52 @@
...
+        const [filteredEvidenceCodes, setFilteredEvidenceCodes] = useState([]);
+
+        const searchEvidenceCodes = (event) => {
+            console.log(event);
+            let evidenceFilter = {};
+            autocompleteFields.forEach( field => {
+                evidenceFilter[field] = event.query;
+            });
+
+            searchService.search("ecoterm", 15, 0, null, {"evidenceFilter":evidenceFilter})
```
oblodgett, line 14: These can be filtered on the backend via the filter. Just add a isObsolete: false

markquintontulloch: Ah, ok - I think I see what you mean.  Something along the lines of :
`let evidenceFilter = {obsolete: false};`
If I do that though, I get the following error:
`ERROR [org.all.cur.exc.RuntimeExceptionMapper] (executor-thread-7) org.hibernate.search.util.common.SearchException: HSEARCH400614: Cannot use 'predicate:simple-query-string' on field 'obsolete'. Make sure the field is marked as searchable/sortable/projectable/aggregable (whichever is relevant). If it already is, then 'predicate:simple-query-string' is not available for fields of this type.`
We could change that in the OntologyTerm (or just for EcoTerm), but do we actually want to do that?

oblodgett (approved):

PR #167, opened 2021-12-06
Fixed booleans

oblodgett:


oblodgett:
@markquintontulloch we have the booleans fixed so you should be able to use the "obsolete": true filter now.

adamgibs (approved):

PR #168, opened 2021-12-07
Fixes error where getLoadOnlyIRIPrefix is null

markquintontulloch:
Was getting the following error:
java.lang.NullPointerException: Cannot invoke "String.length()"
because "prefix" is null

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
@@ -100,7 +100,7 @@ public class GenericOntologyLoadHelper<T extends OntologyTerm> implements OWLObj
     
             if(
                 (termParent.getNamespace() != null && requiredNamespaces.contains(termParent.getNamespace())) ||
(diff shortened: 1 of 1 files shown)


src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
```diff
@@ -100,7 +100,7 @@ public T traverse(OWLClass parent, int depth, ArrayList<String> requiredNamespac
     
             if(
                 (termParent.getNamespace() != null && requiredNamespaces.contains(termParent.getNamespace())) ||
-                (parent.getIRI().getShortForm().startsWith(config.getLoadOnlyIRIPrefix()))
+                (config.getLoadOnlyIRIPrefix() != null && parent.getIRI().getShortForm().startsWith(config.getLoadOnlyIRIPrefix()))
```
oblodgett, line 103: Oh... i guess that produces a null pointer. Sorry about that.

oblodgett (approved):

markquintontulloch:
@oblodgett - I think the above commit addresses most of the issues.  The refactor makes things easier to understand, to my mind at least, and I think I'm now doing the correct thing with regards to checking entities that are referred to are in the database and retrieving/setting the full objects.

PR #169, opened 2021-12-07
SCRUM-594 prevent update with obsolete DO terms

markquintontulloch:
Exclude obsolete DO terms from autocomplete and prevent updating with obsolete terms (unless annotation updated on another field but existing obsolete DO term unchanged).

Also brings in updated logic for excluding obsolete ECO terms from autocomplete, doing this in the backend search rather than a filter of returned terms.


src/main/cliapp/src/components/EvidenceEditor.js
```diff
@@ -10,20 +10,16 @@ import {AutoComplete} from "primereact/autocomplete";
...
+            searchService.search("ecoterm", 15, 0, null, {"evidenceFilter":evidenceFilter, "obsoleteFilter:":obsoleteFilter})
                 .then((data) => {
-                    console.log(data);
-                    if (data.results) {
-                        setFilteredEvidenceCodes(data.results.filter((ecoterm) => Boolean(!ecoterm.obsolete)));
-                    }
-                    else {
-                        setFilteredEvidenceCodes([]);
-                    }
+                    console.log(data)
```
oblodgett, line 17: Can we comment this console.log out? we try to get rid of those unless we are diagnosing an error or something.

src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
```diff
@@ -212,6 +213,7 @@ public void validateAnnotation(DiseaseAnnotation entity, boolean isUpdate) {
                 throw new ApiErrorException(response);
                 // do not continue validation for update if Disease Annotation ID has not been found
             }       
+            originalDoTerm = entity.getCurie();
```
oblodgett, line 216: This is not the DoTerm this is the auto generated Curie. you might be looking for entity.getObject().getCurie()?

oblodgett (changes requested):
I think we need to go over this PR... there are several things that need to change.

oblodgett (approved):

PR #170, opened 2021-12-07
SCRUM-953: Breaking change increasing name for Ontology Term

abecerra:
There are 8364 names in CheBI larger than 255 chars.
The largest name has 1741 chars
This pr implies to drop and reload the ontology term related tables

Tested by:

- Loading zeco ontology without errors
- Checking table definition:

  Table "public.ontologyterm"
   Column    |            Type             | Collation | Nullable | Default | Storage  | Stats target | Description 
-------------+-----------------------------+-----------+----------+---------+----------+--------------+-------------
 curie       | character varying(255)      |           | not null |         | extended |              | 
 created     | timestamp without time zone |           |          |         | plain    |              | 
 lastupdated | timestamp without time zone |           |          |         | plain    |              | 
 definition  | text                        |           |          |         | extended |              | 
 name        | character varying(2000)     |           |          |         | extended |              | 
`


oblodgett:
@abecerra this actually can be handled in the DB before we deploy code via doing a modification to the column.

abecerra:
> @abecerra this actually can be handled in the DB before we deploy code via doing a modification to the column.

Yes, as we are increasing size, maybe there is no problem trying in alpha

oblodgett:
@abecerra do you have a query that can be run to update the column? So that we don't have to drop the table?

abecerra:
> @abecerra do you have a query that can be run to update the column? So that we don't have to drop the table?

I've just tested succesfully: 

`ALTER TABLE ontologyterm ALTER COLUMN name TYPE varchar(2000);`

oblodgett (approved):

PR #174, opened 2021-12-10
SCRUM-879 evidence code abbreviations

markquintontulloch:
Uses a hard-coded list of abbreviations in a helper, so that when the ECO ontology is loaded the abbreviations get added.  This seemed better than having a separate script or running separate loads as it will be more manageable.


oblodgett:
Ok... this one needs some help.

markquintontulloch:
thought it might!

markquintontulloch:
Closing - will use a different approach of having ECOTerm abbreviations stored as a controlled vocabulary and use an API endpoint to allow updating from that table.

PR #178, opened 2021-12-13
Make Negated field editable with dropdown menu

kthorat-prog:


src/main/cliapp/src/components/datatables/DiseaseAnnotationsComponent.js
```diff
@@ -43,7 +44,10 @@ export const DiseaseAnnotationsComponent = () => {
     const toast_topleft = useRef(null);
     const toast_topright = useRef(null);
 
-
+    const negatedTerms = [
+        { name: 'True' },
+        { name: 'False' }
+    ];
```
oblodgett, line 50: These terms should come from the `ControlledVocabularyService` just setup a list in there and give it a name.

kthorat-prog: exactly 
exactly like this one - useQuery(['diseaseRelationTerms'],
        () => controlledVocabularyService.getTerms('disease_relation_terms'), {
            onSuccess: (data) => {
                setDiseaseRelationTerms(data)
            }
        }
    )

oblodgett: yes but the name is: `generic_boolean_terms`

kthorat-prog: yea..got it...thanks

PR #179, opened 2021-12-13
Scrum 499

kthorat-prog:


src/main/cliapp/src/components/TrueFalseDropDownSelector.js
```diff
@@ -0,0 +1,23 @@
+import React, {useState} from 'react';
+import { Dropdown } from "primereact/dropdown"
+
+export function TrueFalseDropdown({ options, editorChange, props }) {
+    const [selectedValue, setSelectedValue] = useState();
```
adamgibs, line 5: I may be wrong, but I think ControlledVocabularySelector could be reused instead of creating a new component. There would just need to be some logic added to turn the placeholder text into a string. Somthing like this: 
```

    let placeholderValue; 
    
    if(typeof props.rowData[props.field] !== 'string'){
        placeholderValue = JSON.stringify(props.rowData[props.field]);
    } else{
        placeholderValue = props.rowData[props.field];
    }
```
   and then just pass the value as a prop like this:

`placeholder={placeholderValue}`

kthorat-prog: I did think of this one but, I just thought its better to keep Boolean drop down separate than the string one for future because in future many more table columns are gonna be there.

adamgibs: That's a good point. Either way seems reasonable to me.

adamgibs (approved):

PR #181, opened 2021-12-14
SCRUM-879 evidence code abbreviations mk2

markquintontulloch:
Also includes loading of xrefs for ontologies (which fixes SCRUM-715), but not in strict accordance with the LinkML model as the required field 'page_areas' is not populated.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/dao/VocabularyDAO.java
+++ b/src/main/java/org/alliancegenome/curation_api/dao/VocabularyDAO.java
@@ -1,6 +1,7 @@
 package org.alliancegenome.curation_api.dao;
 
 import javax.enterprise.context.ApplicationScoped;
+import javax.persistence.Query;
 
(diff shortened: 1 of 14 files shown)


src/main/java/org/alliancegenome/curation_api/dao/VocabularyDAO.java
```diff
@@ -11,4 +12,10 @@
     protected VocabularyDAO() {
         super(Vocabulary.class);
     }
+    
+    public Long getIdFromName(String name) {
+        Query jpqlQuery = entityManager.createQuery("SELECT vocabulary.id FROM Vocabulary vocabulary WHERE vocabulary.name=:name");
+        jpqlQuery.setParameter("name", name);
+        return(Long) jpqlQuery.getSingleResult();
+    }
```
oblodgett, line 20: There is already a "find" framework in place so that we don't need this extra code, its similar to the "search" methods.

oblodgett (approved):

PR #183, opened 2021-12-14
SCRUM-257: Added integration testing to PR validation

mluypaert:
This is the first attempt to run the integration tests as part of the PR validation.
This PR serves as a testing platform until further notice, so should not be merged yet. I will update this PR once ready for merging.


mluypaert:
@oblodgett The integration tests succesfully ran as part of the [latest PR validation run](https://github.com/alliance-genome/agr_curation/runs/4522999598?check_suite_focus=true) on this PR and this was the result:
> Tests run: 86, Failures: 10, Errors: 0, Skipped: 0

As a consequence, the PR validation failed (as it should). Do we intend to fix these 10 failing integration tests or disable them going forward?

oblodgett:
Intend to fix. Also what is the need for the external script? vs just running directly?

mluypaert:
> Intend to fix. Also what is the need for the external script? vs just running directly?

I moved the API package command in an external script so it could be called in both gh-actions and in the Dockerfile, to avoid duplicate code to do this (without needing `make` to be installed as part of the docker container image building).

oblodgett (approved):

PR #187, opened 2021-12-14
SCRUM-962: replace hard-coded disease relation terms for DB results

abecerra:


src/main/cliapp/src/components/datatables/DiseaseAnnotationsComponent.js
```diff
@@ -76,11 +76,10 @@ export const DiseaseAnnotationsComponent = () => {
...
     useQuery(['diseaseRelationTerms'],
-        () => controlledVocabularyService.getTerms('disease_relation_terms'), {
-            onSuccess: (data) => {
-                setDiseaseRelationTerms(data)
-            }
-        }
+        () => searchService.search("vocabularyterm", 15, 0, null, {"vocabFilter": {"vocabulary.name": "Disease Relation Vocabulary"}})
+        .then((data) => {
+            setDiseaseRelationTerms(data.results);
+        })
```
oblodgett, line 82: @abecerra this works. However some vocab's can't be selected this way. So it would be better to use the controlledVocabularyService and pass in "Disease Relation Vocabulary" to the `getTerms` method.

oblodgett (changes requested):

src/main/cliapp/src/components/datatables/DiseaseAnnotationsComponent.js
abecerra: Ok!

adamgibs: @oblodgett I helped andreas with this, I thought we talked about switching out the controlledVocabService for the searchService? Or do you mean put a call to the search service within the ConrolledVocabService?

src/main/cliapp/src/service/ControlledVocabularyService.js
```diff
@@ -42,9 +42,16 @@ export class ControlledVocabularyService {
...
-        return this.terms[termType].terms;
+        if (termType in this.terms) {
+            return this.terms[termType]['terms'];
+        }
+        else {
+            SearchService.search("vocabularyterm", 15, 0, null, {"vocabFilter": {"vocabulary.name": termType}})
+            .then((data) => {
+                return data.results;
+            })
+        }
```
oblodgett, line 53: Not sure this is going to work with promises etc.

src/main/cliapp/src/components/datatables/DiseaseAnnotationsComponent.js
oblodgett: Ya... you are right.. I think the issue with promises etc is going to be a problem.

oblodgett (approved):

PR #204, opened 2022-01-06
SCRUM-547 docker logging

mluypaert:
(Re-)enabled application docker logging to be sent to the central log server.


mluypaert:
Currently, the application logs show up in the central logging server with this change (search for "quarkus"), but the container name fails to show up in the systems list (an issue that was noticed with other systems before as well).

mluypaert:
Discovered incorrect ticket nr usage in branch name and commit messaged. Will re-open with rebased commits.

PR #205, opened 2022-01-06
SCRUM-547 docker logging

mluypaert:
(Re-)enabled application docker logging to be sent to the central log server.

Currently, the application logs show up in the central logging server with this change (search for "quarkus"), but the container name fails to show up in the systems list (an issue that was noticed with other systems before as well).


docker-compose.yml
```diff
@@ -3,8 +3,12 @@ version: "3.2"
 services:
 
   curation.app:
-    container_name: agr.curation.api.server
+    container_name: agr.curation.${NET}.api.server
     image: 100225593120.dkr.ecr.us-east-1.amazonaws.com/agr_curation:${AGR_CURATION_RELEASE}
+    logging:
+      driver: gelf
+      options:
+        gelf-address: 'udp://logs.alliancegenome.internal:12201'
```
oblodgett, line 11: We don't need to use the .internal anymore we can use the .org

oblodgett, line 11: PS Plus I have tried these gelf lines in the past and for some reason the docker on the EBS instance does not pass the log data on to the log server.

mluypaert: This worked fine for me (see logtrail and search for "quarkus"), however the container name does not show up in the system list for some reason (something we've noticed before with some other containers).

Makefile
```diff
@@ -75,7 +76,8 @@ eb-init:
 	eb init --region us-east-1 -p Docker curation-app
 
 eb-create:
-	@eb create ${ENV_NAME} --region=us-east-1 --cname="${ENV_NAME}" --elb-type application
+	@eb create ${ENV_NAME} --region=us-east-1 --cname="${ENV_NAME}" --elb-type application \
+	  --envvars QUARKUS_QPID_JMS_URL=${QUARKUS_QPID_JMS_URL},QUARKUS_QPID_JMS_USERNAME=${QUARKUS_QPID_JMS_USERNAME},QUARKUS_HIBERNATE_SEARCH_ORM_ELASTICSEARCH_HOSTS=${QUARKUS_HIBERNATE_SEARCH_ORM_ELASTICSEARCH_HOSTS},QUARKUS_DATASOURCE_USERNAME=${QUARKUS_DATASOURCE_USERNAME},BULK_DATA_LOADS_S3SECRETKEY=${BULK_DATA_LOADS_S3SECRETKEY},BULK_DATA_LOADS_S3ACCESSKEY=${BULK_DATA_LOADS_S3ACCESSKEY},QUARKUS_DATASOURCE_JDBC_URL=${QUARKUS_DATASOURCE_JDBC_URL},QUARKUS_QPID_JMS_PASSWORD=${QUARKUS_QPID_JMS_PASSWORD},BULK_DATA_LOADS_S3PATHPREFIX=${BULK_DATA_LOADS_S3PATHPREFIX},QUARKUS_DATASOURCE_PASSWORD=${QUARKUS_DATASOURCE_PASSWORD},QUARKUS_HIBERNATE_SEARCH_ORM_ELASTICSEARCH_PROTOCOL=${QUARKUS_HIBERNATE_SEARCH_ORM_ELASTICSEARCH_PROTOCOL},NET=${NET}
```
oblodgett, line 80: I think it would be better to not have these variable in here as they change. Like now the JMS variables are no longer used.j

oblodgett (approved):

Makefile
mluypaert: What commit was the use of those JMS variables removed in? They're still defined in the Dockerfile and in the EB properties (of the alpha environment)?
That's exactly why I want them defined in the Makefile, it makes (re)creating an EB environment much more reliable, as keeping the Makefile up-to-date then allows anyone to see what variables need to be set to create an environment to support a specific version of the code, and doesn't require reliance on looking at current configurations of existing environments as much, as those might be in a different state (requiring different variables) than the code needing deployment to a new environment.

PR #208, opened 2022-01-07
SCRUM-1006 Link Genes to SOTerms

markquintontulloch:
Can populate the new column using the following SQL query:
`UPDATE gene SET genetype_curie = type;`


oblodgett (approved):

oblodgett:
Did you test a SO Term Load?

markquintontulloch:
@oblodgett - Yep, SOTerms loaded on alpha already.  Linking of genes to SOTerms tested locally.

PR #213, opened 2022-01-07
Update molecule column properties

markquintontulloch:


oblodgett (approved):
This will not be an automatic change as the fields are TEXT and now they are going to a varchar(size)

markquintontulloch:
The database was still on varchar(255) - I've changed the column properties manually on alpha

oblodgett:
Ahh ok... a change to the size will automatically go the database... changing a type wont.

PR #214, opened 2022-01-07
automated_loads disabled some form fields on edit

adamgibs:


oblodgett:
Maybe I missed something but where is the true / false values for each specific form?

Also I think I like the props idea vs the long lists of params coming into the method... what do you think?

oblodgett (approved):

PR #218, opened 2022-01-11
SCRUM-948 Load all terms with MP prefix

markquintontulloch:
@oblodgett - When I test this locally, if I load the ontology via the API then ~14,000 terms load (as expected with the changes), but if I load it via the Data Loads page then only the original number of terms (~12,000) get loaded.  Any thoughts?


oblodgett:
Ahh yes... the loader needs to be configured the same as the controller see `BulkLoadJobExecutor.java:166` you will need to add a ` config.setLoadOnlyIRIPrefix("MP");`

oblodgett:
PS Its not good having the configuration like that duplicated in two places. I'll come up with a way to only have it in one place.

markquintontulloch:
Done - thanks, I thought it would be something like that but that must have been the one file I didn't look at!

oblodgett (approved):

PR #219, opened 2022-01-11
SCRUM-948 Add underscore to prefix in IRI search

markquintontulloch:
Avoids loading of ontology terms with prefixes for which the
desired prefix is a substring at the start of the actual prefix
- e.g. avoids MPATH terms being loaded when MP prefix is
specified in config

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
@@ -99,7 +99,7 @@ public class GenericOntologyLoadHelper<T extends OntologyTerm> implements OWLObj
     
             if(
                 (termParent.getNamespace() != null && requiredNamespaces.contains(termParent.getNamespace())) ||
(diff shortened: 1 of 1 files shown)


src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
```diff
@@ -99,7 +99,7 @@ public T traverse(OWLClass parent, int depth, ArrayList<String> requiredNamespac
     
             if(
                 (termParent.getNamespace() != null && requiredNamespaces.contains(termParent.getNamespace())) ||
-                (config.getLoadOnlyIRIPrefix() != null && parent.getIRI().getShortForm().startsWith(config.getLoadOnlyIRIPrefix()))
+                (config.getLoadOnlyIRIPrefix() != null && parent.getIRI().getShortForm().startsWith(config.getLoadOnlyIRIPrefix() + "_"))
```
oblodgett, line 102: Ahh good catch "MP" vs "MPATH"

oblodgett (approved):

PR #220, opened 2022-01-11
SCRUM-1028

adamgibs:


oblodgett:
This looks good... just want to make sure this doesn't hinder us from using facets on the filters?

oblodgett (approved):

adamgibs:
@oblodgett I wouldn't _think_ so. If we need a specialized filter type, then I think that would just mean  a specialized filterComponentTemplate that is then set to the value of the filterElement field in the column array.

PR #226, opened 2022-01-14
SCRUM-254 Allele testing

markquintontulloch:
Allele bulk and CRUD validation and integration tests.  A bunch of gene stuff came along for the ride too.


src/main/java/org/alliancegenome/curation_api/services/AffectedGenomicModelService.java
```diff
@@ -197,14 +199,22 @@ private void handleSecondaryIds(AffectedGenomicModelDTO agm, AffectedGenomicMode
...
         }
         
+        // Validate taxon ID
+        Pattern taxonIdPattern = Pattern.compile("^NCBITaxon:\\d+$");
+        Matcher taxonIdMatcher = taxonIdPattern.matcher(agm.getTaxonId());
+        if (!taxonIdMatcher.find()) {
+            log.debug("Invalid taxon ID for AGM " + agm.getPrimaryID() + " - skipping");
+            return false;
+        }
+
```
oblodgett, line 217: A better way to do this might be to do a looking of the taxon in a HashSet<String> we are going to need taxons at some point... it will also get rid of the list thats in the Job Executer also.

markquintontulloch: Yeah, this is only meant to be in the interim until we have taxons loaded.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/services/AffectedGenomicModelService.java
markquintontulloch: Ah, I see what you mean - I was forgetting that the taxons here are limited to the MOD taxons, unlike those in the experimental conditions.

oblodgett: Either way... its fine... we need to get the taxons in the DB at some point.

PR #228, opened 2022-01-17
SCRUM-338 Taxon check

markquintontulloch:
Possible solution to issue of loading taxons.

The NCBITaxon OWL file is large and the format is not recognised by our OWL parsing library.  We do not require the majority of taxon entities to be loaded, but we do want to be able to validate taxon IDs that are entered through the curation interface or via bulk loads.  This proposed solution carries out a validation check when a taxon ID is encountered using the following logic:

- The taxon ID is used to search the curation database.  If an entry is found then the validation passes.
- If no entry is found then the NCBI eUtils API is queried using the taxon ID.
- If the response does not contain any results then validation fails, if it does then the curation database is updated with that term.


oblodgett (approved):

oblodgett:
@markquintontulloch did know if you wanted to get the tests in with this PR?

markquintontulloch:
@oblodgett - I've not managed to get the integration tests to work (getting a null pointer exception on the NcbiTaxonTermDAO.  Neither myself nor Andreas have currently got a solution for this.

My thoughts are that we should merge this branch as is (with tests commented out) and address them later - what do you think?

abecerra:
> @oblodgett - I've not managed to get the integration tests to work (getting a null pointer exception on the NcbiTaxonTermDAO). Neither myself nor Andreas have currently got a solution for this.
> 
> My thoughts are that we should merge this branch as is (with tests commented out) and address them later - what do you think?

I think leaving the comments is a good approach, at some point quarkus Devs will include injection for integration tests, I guess

PR #233, opened 2022-01-18
SCRUM-984

cmpich:


src/main/java/org/alliancegenome/curation_api/model/entities/Condition.java
```diff
@@ -0,0 +1,41 @@
...
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonView;
+import lombok.Getter;
+import lombok.Setter;
+import org.alliancegenome.curation_api.view.View;
+
+@Setter
+@Getter
+public class Condition {
```
oblodgett, line 11: @cmpich is this suppose to be an Entity?

src/main/java/org/alliancegenome/curation_api/model/entities/ConditionRelation.java
```diff
@@ -0,0 +1,31 @@
...
+@EqualsAndHashCode(onlyExplicitlyIncluded = true)
+@Schema(name = "Disease_Relation", description = "Condition class ")
+public class ConditionRelation extends BaseGeneratedAndUniqueIdEntity  {
+
+    @JsonView({View.FieldsOnly.class})
+    @JsonProperty("condition_relation_type")
+    private String relationType;
+
+    @JsonView({View.FieldsAndLists.class})
+    private List<Condition> conditions;
```
oblodgett, line 25: @cmpich are these relations in the DB? if so they need @ManytoOne or @OnetoMany annotations on them.

abecerra (approved):
looks good

PR #236, opened 2022-01-19
SCRUM-1068 & SCRUM-1069 Replace taxon curies with taxon objects - DO NOT MERGE YET?

markquintontulloch:
Update code to link to NCBITaxonTerm objects from Genes, Alleles, and AGMs.  Once the code has been merged, the following SQL statements will need to be executed on alpha:
`UPDATE biologicalentity SET taxon_curie = taxon;`
`ALTER TABLE biologicalentity DROP COLUMN taxon CASCADE;`
Genes, Alleles, and AGMs will also need to be reindexed.

Given the above, and in case there are any unforseen consequences, it probably makes sense to hold off merging this branch until after Sprint Review.


oblodgett (approved):
Looks good... but might want to wait till after the demo?

markquintontulloch:
@oblodgett - Agreed, will hold off merging until after demo

PR #239, opened 2022-01-20
Validation and integration testing updates

markquintontulloch:
Test order added to all integration tests - doesn't solve the random test failure issue, but rules it out at least.
Added validation for xref IDs - avoids 500 response and just skips loading of entity if xref ID missing.
Allow multiple connection attempts to NCBI REST API to retrieve taxon details if call initially fails.


src/main/java/org/alliancegenome/curation_api/dao/ontology/NcbiTaxonTermDAO.java
```diff
@@ -29,9 +30,21 @@ public NCBITaxonTerm downloadAndSave(String taxonCurie) {
             return null;
         }
         
-        NCBITaxonResponseDTO resp = api.getTaxonFromNCBI("taxonomy", "json", taxonIdMatcher.group(1));
-        HashMap<String, Object> result = resp.getResult();
-        HashMap<String, Object> taxonMap = (HashMap<String, Object>)result.get(taxonIdMatcher.group(1));
+        HashMap<String, Object> taxonMap = null;
+        int attempts = 5;
```
abecerra, line 34: Could you name it something lik MAX_ATTEMPTS and make it const ?

abecerra (approved):

src/main/java/org/alliancegenome/curation_api/dao/ontology/NcbiTaxonTermDAO.java
markquintontulloch: Done

PR #244, opened 2022-01-25
SCRUM-1067 Add appropriate file extension to s3 URLs

markquintontulloch:
No idea if this is a good way of doing this, but works and should cope with more file types in the future.

Determine filetype from source URL or datatype and store in bulkloadfile table.
Add appropriate extension when saving to s3;

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadFileProcessor.java
+++ b/src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadFileProcessor.java
@@ -2,6 +2,8 @@ package org.alliancegenome.curation_api.jobs;
 
 import java.io.*;
 import java.util.*;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 import java.util.zip.GZIPInputStream;
 
 import javax.enterprise.context.ApplicationScoped;
@@ -60,7 +62,8 @@ public class BulkLoadFileProcessor {
             String s3Url = processFMS(bulkFMSLoad.getDataType(), bulkFMSLoad.getDataSubType());
             String filePath = fileHelper.saveIncomingURLFile(s3Url);
             String localFilePath = fileHelper.compressInputFile(filePath);
-            processFilePath(bulkFMSLoad, localFilePath);
+            String fileType = parseURLFileType(s3Url);
+            processFilePath(bulkFMSLoad, localFilePath, fileType);
             bulkFMSLoad.setErrorMessage(null);
             bulkFMSLoad = bulkFMSLoadDAO.find(bulkFMSLoad.getId());
             bulkFMSLoad.setStatus(BulkLoadStatus.FINISHED);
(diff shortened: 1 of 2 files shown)


oblodgett:
@markquintontulloch my initial thoughts on this was to put this field on the BulkLoad so that it would be specified in the UI, and call it fileExtension.

src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadFileProcessor.java
```diff
@@ -244,6 +250,7 @@ private void processFilePath(BulkLoad bulkLoad, String localFilePath) {
             bulkLoadFile.setFileSize(inputFile.length());
             bulkLoadFile.setStatus(BulkLoadStatus.PENDING);
             bulkLoadFile.setLocalFilePath(localFilePath);
+            bulkLoadFile.setFileType(fileType);
```
oblodgett, line 253: @markquintontulloch for instance this line would be a `bulkLoadFile.setFileExtension(bulkLoad.getFileExtension());

markquintontulloch: @oblodgett - Makes sense, I'll revisit it tomorrow, thanks

oblodgett: And then `bulkLoad.fileExtension` would be set in the UI when creating the BulkLoad, plus we need to considering the files that have been uploaded already. Those may need to be renamed in S3 to that the URLs continue to work.

src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkLoadFile.java
```diff
@@ -56,6 +59,6 @@ public String getS3Url() {
     @JsonIgnore
     @JsonView({View.FieldsOnly.class})
     public String generateS3MD5Path() {
-        return md5Sum.charAt(0) + "/" + md5Sum.charAt(1) + "/" + md5Sum.charAt(2) + "/" + md5Sum.charAt(3) + "/" + md5Sum + ".gz";
+        return md5Sum.charAt(0) + "/" + md5Sum.charAt(1) + "/" + md5Sum.charAt(2) + "/" + md5Sum.charAt(3) + "/" + md5Sum + "." + fileType + ".gz";
```
oblodgett, line 62: Actually upon more investigation I think the only change that needs to happen is  `md5Sum + "." + fileType + ".gz"` -> `md5Sum + "." + bulkLoad.getFileExtension() + ".gz"` and the fileExtension field needs to be added to the BulkLoad object only and that will take care of everything. And the UI needs to set that field upon creation.

markquintontulloch:
Closing this PR and creating a new one as going with the approach suggested by @oblodgett above

PR #246, opened 2022-01-26
SCRUM-1067 mk2

markquintontulloch:
Add fileExtension to BulkLoad entity and select from dropdown in UI on creation of bulk load.

### DIFF
--- a/src/main/cliapp/src/containers/dataLoadsPage/NewBulkLoadForm.js
+++ b/src/main/cliapp/src/containers/dataLoadsPage/NewBulkLoadForm.js
@@ -23,7 +23,6 @@ export const NewBulkLoadForm = ({ bulkLoadDialog, setBulkLoadDialog, groups, new
 
     const [backendBulkLoadTypes, setBackendLoadTypes] = useState();
     const loadTypes = dataLoadService.getLoadTypes();
-	const fileExtensions = dataLoadService.getFileExtensions();
 
     const mutation = useMutation(bulkLoad => {
         if (bulkLoad.id) {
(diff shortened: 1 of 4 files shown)


oblodgett:
@markquintontulloch I think it would be better to not have it be an enum but rather just free text... this way other file types can be added without code changes. (gff, obo, vcf, gaf, etc, etc)

markquintontulloch:
@oblodgett - fair enough, done

src/main/cliapp/src/containers/dataLoadsPage/NewBulkLoadForm.js
```diff
@@ -170,6 +170,18 @@ export const NewBulkLoadForm = ({ bulkLoadDialog, setBulkLoadDialog, groups, new
...
 
+					<div className="p-field">
+						<label htmlFor="fileExtension">File Extension</label>
+						<InputText
+							id="bulkLoadFileExtension"
+							value={newBulkLoad.fileExtension}
+							onChange={onChange}
+							placeholder={"Select Bulk Load File Extension"}
+							name='fileExtension'
+							disabled={disableFormFields}
```
oblodgett, line 181: @markquintontulloch sorry last item... not sure if this field is specified somewhere else? Maybe disableFormFields is just a boolean.

oblodgett (approved):

src/main/cliapp/src/containers/dataLoadsPage/NewBulkLoadForm.js
oblodgett: This is used so that fileExtension can't be changed after files have been uploaded.

markquintontulloch: isn't that what we want, otherwise the filename won't match the saved file?

oblodgett: Correct... I assume that disableFormFields is a boolean and not a list of fields to disable. If it is a list, then this field would need to be specified in another place to "disable" after editing @adamgibs ? Maybe it's just late and I need to look at this in the morning.

markquintontulloch: It does disable the field from being edited on my local system, but I'll let @adamgibs take a look before merging in case I'm doing something funky!

oblodgett: Oh ok... then yes looks good

oblodgett (approved):

PR #249, opened 2022-01-27
Scrum 1045

oblodgett:


adamgibs (approved):

src/main/java/org/alliancegenome/curation_api/auth/AuthenticationFilter.java
```diff
@@ -0,0 +1,166 @@
...
+        } else {
+            log.warn("OKTA Authentication Disabled using Test Dev User");
+            SearchResponse<Person> res = personDAO.findPersonByEmail("test@alliancegenome.org");
+            if(res == null) {
+                Person person = new Person();
+                person.setApiToken(UUID.randomUUID().toString());
+                person.setEmail("test@alliancegenome.org");
+                person.setFirstName("Local");
+                person.setLastName("Dev User");
+                personDAO.persist(person);
```
abecerra, line 148: Does this person needs to persist?

oblodgett: Yes... that way it can be used multiple times through local development or integration testing.

PR #250, opened 2022-01-27
SCRUM-1082 Filter name fixes

markquintontulloch:


oblodgett:
This would work fine... but you might want to put the word "Filter" on the end of each of those as thats the "name" of the filter thats getting sent to the backend. The only requirement is that they are unique so they can be named whatever.

`filterElement: filterComponentTemplate("conditionTaxon", ["conditionTaxon.curie", "conditionTaxon.name"])`

would become

`filterElement: filterComponentTemplate("conditionTaxonFilter", ["conditionTaxon.curie", "conditionTaxon.name"])`

markquintontulloch:
@oblodgett - "Filter" suffixes added

oblodgett (approved):

PR #251, opened 2022-01-27
SCRUM-1083 persist experimental conditions for disease annotations

cmpich:
I changed the bulk load to persist experimentatl conditions. Both experimentalCondition and ConditionRelation do have a uniqueId which is a concatenation of uniqueIds. If I find an existing one I re-use that one.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/response/SearchResponse.java
+++ b/src/main/java/org/alliancegenome/curation_api/response/SearchResponse.java
@@ -1,29 +1,30 @@
 package org.alliancegenome.curation_api.response;
 
-import java.util.*;
-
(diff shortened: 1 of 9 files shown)


src/main/java/org/alliancegenome/curation_api/response/SearchResponse.java
```diff
@@ -36,4 +37,8 @@ public void setResults(List<E> results) {
             this.results = new ArrayList<E>();
         }
     }
+
+    public E getSingleResult() {
+        return (results == null || CollectionUtils.isEmpty(results)) ? null : results.get(0);
```
oblodgett, line 42: I like this... could have used it earlier today.

oblodgett (approved):

PR #252, opened 2022-01-28
SCRUM-725

markquintontulloch:
Some experimental condition entity fixes, change of EMAPATerm to extend AnatomicalTerm, various Experimental Condition table fixes, and make conditionGeneOntology editable.

Had to be make AnatomicalTerm ExperimentalConditionOntologyTerm and ChemicalTerm non-abstract in order to be able to create test experimental conditions via the API (possibly another way around this?).

Known bug: After editing, conditionGeneOntology only displays curie and not name until page is refreshed.


oblodgett:
Ya looks good... @adamgibs maybe you can weigh in on all the UI stuff.

markquintontulloch:
Along with the display (until refresh) bug described in the first comment, another known bug is that you cannot delete the conditionGeneOntology entry by entering an empty string - we should be able to do this as it is a non-required field.
@adamgibs  or @kthorat-prog - any thoughts on how to solve these two issues?

abecerra (approved):

PR #262, opened 2022-02-01
SCRUM-1154 Generalise editing of single-value ontology fields

markquintontulloch:
Create general method for editing of single-value ontology fields.
Create Java components for AnatomicalTerm, ChemicalTerm, and
ExperimentalConditionOntologyTerm to allow autocomplete.
Remove redundant ConditionGeneOntology code.

### DIFF
--- a/src/main/cliapp/src/containers/experimentalConditionsPage/ExperimentalConditionsTable.js
+++ b/src/main/cliapp/src/containers/experimentalConditionsPage/ExperimentalConditionsTable.js
@@ -247,11 +247,11 @@ export const ExperimentalConditionsTable = () => {
       }
     };
 
-  const singleOntologyEditorTemplate = (props, fieldname, endpoint) => {
+  const singleOntologyEditorTemplate = (props, fieldname, endpoint, autocomplete) => {
     return (
       <>
         <SingleOntologyEditor
-          autocompleteFields={["curie", "name", "crossReferences.curie", "secondaryIdentifiers", "synonyms"]}
+          autocompleteFields={autocomplete}
           rowProps={props}
           searchService={searchService}
           setExperimentalConditions={setExperimentalConditions}
@@ -265,6 +265,8 @@ export const ExperimentalConditionsTable = () => {
       </>
     );
   };
+  
+  const curieAutocompleteFields = ["curie", "name", "crossReferences.curie", "secondaryIdentifiers", "synonyms"];
 
   const columns = [
     {
@@ -290,7 +292,7 @@ export const ExperimentalConditionsTable = () => {
       body: conditionClassBodyTemplate,
       filter : true, 
       filterElement: filterComponentTemplate("conditionClassFilter", ["conditionClass.curie", "conditionClass.name"]),
-      editor: (props) => singleOntologyEditorTemplate(props, "conditionClass", "zecoterm")
+      editor: (props) => singleOntologyEditorTemplate(props, "conditionClass", "zecoterm", curieAutocompleteFields)
     },
     {
       field:"conditionId.name",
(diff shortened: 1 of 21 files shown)


oblodgett:
I cant take a better look at this until i get back to a computer. But i would like to add the endpoints for ontologyTerm also. So when indexing you only have to hit one endpoint.

src/main/cliapp/src/containers/experimentalConditionsPage/ExperimentalConditionsTable.js
```diff
@@ -240,23 +240,25 @@ export const ExperimentalConditionsTable = () => {
...
-                errorMessages={errorMessages[props.rowIndex]}
-                errorField={"conditionGeneOntology"}
-            />
-          </>
-      );
+  const singleOntologyEditorTemplate = (props, fieldname, endpoint) => {
+    return (
+      <>
+        <SingleOntologyEditor
+          autocompleteFields={["curie", "name", "crossReferences.curie", "secondaryIdentifiers", "synonyms"]}
```
oblodgett: The only issue I see with this is that these fields might want to be passed in rather than hardcoded here.

src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/DiseaseAnnotationCurie.java
```diff
@@ -55,7 +55,7 @@ public String getCurieID(
         return curie.getCurie();
     }
 
-    private static String getExperimentalConditionCurie(ExperimentalCondition cond) {
+    public static String getExperimentalConditionCurie(ExperimentalCondition cond) {
```
oblodgett, line 58: Curious on this change? Does this now get called from a place that wasn't before?

markquintontulloch: Yes, from the validator.  That way, when the fields get updated the compound unique_id also gets updated.

src/main/cliapp/src/containers/experimentalConditionsPage/ExperimentalConditionsTable.js
markquintontulloch: @oblodgett - done in latest commit

oblodgett:
@markquintontulloch I know it's crazy late for you. But I would like to get a PR into this PR that adds the OntologyTerm endpoints so we can index ALL the ontologies at the same time through the reindex endpoint, but it will be in like 2 hours.

src/main/cliapp/src/containers/experimentalConditionsPage/ExperimentalConditionsTable.js
```diff
@@ -119,18 +119,25 @@ export const ExperimentalConditionsTable = () => {
...
+      for (var ix = 0; ix < curieFields.length; ix++) {
+        if (event.data[curieFields[ix]] && Object.keys(event.data[curieFields[ix]]).length >= 1) {
+          event.data[curieFields[ix]].curie = trimWhitespace(event.data[curieFields[ix]].curie);
+          updatedRow[curieFields[ix]] = {};
+          updatedRow[curieFields[ix]] = event.data[curieFields[ix]];
+        }
       }
-
       mutation.mutate(updatedRow, {
           onSuccess: (data, variables, context) => {
```
oblodgett: Might be better here to change `data` to `response` that was later on `data.data` turns into `response.data`

markquintontulloch: @oblodgett - done, and OntologyTerm endpoints added

oblodgett:
@markquintontulloch looks great thanks.

oblodgett (approved):

PR #266, opened 2022-02-02
README-only PR validation

mluypaert:
Changes required to allow PR merging of PR only containing changes to currently ignored paths.


oblodgett (approved):

mluypaert:
@oblodgett The easiest solution would probably be to just assume every update for which a PR is opened must be validated, and remove the `paths-ignore` section of the PR validation?

PR #268, opened 2022-02-02
SCRUM-1083 add conditionId population / XCOTerm

cmpich:


abecerra (approved):

src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationFmsService.java
```diff
@@ -119,10 +121,22 @@ public DiseaseAnnotation upsert(DiseaseModelAnnotationFmsDTO annotationDTO) {
                         ChemicalTerm term = chemicalTermDAO.find(experimentalConditionDTO.getChemicalOntologyId());
                         experimentalCondition.setConditionChemical(term);
                     }
+                    if (experimentalConditionDTO.getConditionId() != null) {
+                        XcoTerm term = xcoTermDAO.find(experimentalConditionDTO.getConditionId());
```
markquintontulloch: This should be an ExperimentalConditionOntologyTerm rather than an XcoTerm

markquintontulloch (changes requested):
You also need to add loading of conditionGeneOntology

markquintontulloch (approved):

markquintontulloch (approved):

markquintontulloch (changes requested):
Just noticed that conditionQuantity is missing too

markquintontulloch (approved):

oblodgett:
@cmpich was this logic duplicated in another file somewhere? Just wondering if we should try to combine it?

cmpich:
I don't see a duplication. Though, there is code to create db ExperimentalCondition from a ui ExperimentalCondition. Different from what I am doing here.

oblodgett (approved):

PR #275, opened 2022-02-07
SCRUM-1194 Experimental condition validation and testing for FMS loads

markquintontulloch:
We're having to jump through a lot of hoops to load required entities to avoid 'duplicate key' errors.  @oblodgett @abecerra - any thoughts on the following?

The disease annotation bulk load validation checks for the presence of various entities in the database in order to load each disease annotation entry (GOTerm, ECOTerm, NCBITaxonTerm, etc.).  These need to be loaded into the test database before doing the disease annotation test loads, and this is done using a method that extracts the curies from the JSON and creates each one using a POST request.  However, we are having to keep track of loaded entities as trying to load the same entity twice is causing 'duplicate_key' errors.  

My understanding is that a POST request would either create a non-existing entry or update an existing one, but this doesn't appear to be the case.  

This is starting to get really complicated as it is not just the loads in question that are affected. e.g. I had to remove various entities from one of the molecule test files, as these had previously been loaded in the disease annotation tests.


oblodgett:
POST is always a create and PUT is always a UPDATE so we sometimes create a new endpoint that does a "UPSERT" which is update or insert.

markquintontulloch:
@oblodgett - As discussed, tests using MOD examples removed for all bulk load tests, avoiding the need for complicated entity loading requirements.

abecerra (approved):

PR #278, opened 2022-02-10
SCRUM-1203 Updates for DQM submission of DAs to new schema

markquintontulloch:


src/main/cliapp/src/containers/diseaseAnnotationsPage/DiseaseAnnotationsTable.js
```diff
@@ -445,11 +445,11 @@ export const DiseaseAnnotationsTable = () => {
       body: diseaseBodyTemplate
     },
     {
-      field: "reference.curie",
+      field: "singleReference.curie",
       header: "Reference",
       sortable: isEnabled,
       filter: true,
-      filterElement: filterComponentTemplate("reference", ["reference.curie"])
+      filterElement: filterComponentTemplate("singleReference", ["singleReference.curie"])
```
oblodgett: singleReference should be singleReferenceFilter just to differentiate from the field name.

src/main/java/org/alliancegenome/curation_api/interfaces/crud/AGMDiseaseAnnotationCrudInterface.java
```diff
@@ -34,4 +37,37 @@
...
+    @Path("/bulk/wbAnnotationFile")
+    public String updateWbAgmDiseaseAnnotations(List<DiseaseAnnotationDTO> annotationData);
+    
+    @POST @Secured
+    @Path("/bulk/humanAnnotationFile")
+    public String updateHumanAgmDiseaseAnnotations(List<DiseaseAnnotationDTO> annotationData);
+    
+    @POST @Secured
+    @Path("/bulk/sgdAnnotationFile")
+    public String updateSgdAgmDiseaseAnnotations(List<DiseaseAnnotationDTO> annotationData);
```
oblodgett: @markquintontulloch why do we need to split these out into each their own? Considering the definition hasn't changed?

src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadJobExecutor.java
```diff
@@ -53,6 +54,18 @@
...
+    private Map<String, String> modTaxons = Map.of(
+            "HUMAN", "NCBITaxon:9606",
+            "FB", "NCBITaxon:7227",
+            "MGI", "NCBITaxon:10090",
+            "RGD", "NCBITaxon:10116",
+            "SGD", "NCBITaxon:559292",
+            "WB", "NCBITaxon:6239",
+            "ZFIN", "NCBITaxon:7955"
+            );  
+
```
oblodgett, line 62: This should probably be split out into its own enum see:
https://github.com/alliance-genome/agr_java_software/blob/master/agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/SpeciesType.java
for reference.

src/main/java/org/alliancegenome/curation_api/model/entities/PaperHandle.java
```diff
@@ -21,7 +21,7 @@
     @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
     @ManyToOne
     @JsonView({View.FieldsOnly.class})
-    private Reference reference;
+    private Reference singleReference;
```
oblodgett: Did we discuss getting rid of this class? (Paperhandle that is)

src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
```diff
@@ -58,173 +78,93 @@ protected void init() {
...
         List<String> annotationsIdsBefore = new ArrayList<>();
-        annotationsIdsBefore.addAll(geneDiseaseAnnotationDAO.findAllAnnotationIds(taxonID));
-        annotationsIdsBefore.addAll(alleleDiseaseAnnotationDAO.findAllAnnotationIds(taxonID));
-        annotationsIdsBefore.addAll(agmDiseaseAnnotationDAO.findAllAnnotationIds(taxonID));
+        if (entityType.equals("agm")) {
+            annotationsIdsBefore.addAll(agmDiseaseAnnotationDAO.findAllAnnotationIds(taxonID));
+        } else if (entityType.equals("allele")) {
+            annotationsIdsBefore.addAll(alleleDiseaseAnnotationDAO.findAllAnnotationIds(taxonID));
+        } else if (entityType.equals("gene")) {
+            annotationsIdsBefore.addAll(geneDiseaseAnnotationDAO.findAllAnnotationIds(taxonID));
```
oblodgett: So... I think a better approach here would be to have the 3 different services... that extend DiseaseAnnotationService the common stuff put in the DiseaseAnnotationService and the specific stuff leave in each.

src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
```diff
@@ -250,59 +190,225 @@ public void runLoad(String taxonID, List<DiseaseAnnotationDTO> annotations) {
...
+        } else if (subjectEntity instanceof AffectedGenomicModel) {
+            SearchResponse<AGMDiseaseAnnotation> annotationList = agmDiseaseAnnotationDAO.findByField("uniqueId", annotationID);
+            if (annotationList == null || annotationList.getResults().size() == 0) {
+                AGMDiseaseAnnotation agmAnnotation = new AGMDiseaseAnnotation();
+                agmAnnotation.setUniqueId(annotationID);
+                agmAnnotation.setSubject((AffectedGenomicModel) subjectEntity);
+                annotation = agmAnnotation;
+            } else {
+                annotation = annotationList.getResults().get(0);
+            }
```
oblodgett: Splitting if up would prevent these huge if else if's.

src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
```diff
@@ -250,59 +190,225 @@ public void runLoad(String taxonID, List<DiseaseAnnotationDTO> annotations) {
...
+                withGenes.add(withGene);
+            }
+            annotation.setWith(withGenes);
         }
 
-        return true;
+        return annotation;
+    }
+    
+    private ExperimentalCondition validateExperimentalConditionDTO(ExperimentalConditionDTO dto) {
```
oblodgett: This whole method might be better served put in the ExperimentalConditionService class.

oblodgett:
@markquintontulloch please don't hate me.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/interfaces/crud/AGMDiseaseAnnotationCrudInterface.java
markquintontulloch: Given that the ingest class splits disease annotations into the separate entity classes (i.e. each gets submitted in its own array in the ingest file), in my mind it makes more sense to be able to submit each entity separately.  In this way a MOD can resubmit, for example, just gene disease annotations without removing all agm and allele disease annotations.

oblodgett: Ok... then to my later comments we might want to have the three different types also for `DiseaseAnnotationDTO`

src/main/cliapp/src/containers/diseaseAnnotationsPage/DiseaseAnnotationsTable.js
markquintontulloch: all filter names updated for clarity

src/main/java/org/alliancegenome/curation_api/interfaces/crud/AGMDiseaseAnnotationCrudInterface.java
markquintontulloch: done

src/main/java/org/alliancegenome/curation_api/model/entities/PaperHandle.java
markquintontulloch: removed

src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
markquintontulloch: agreed, have refactored

src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
markquintontulloch: done

src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadJobExecutor.java
```diff
@@ -111,21 +127,54 @@ public void process(BulkLoadFile bulkLoadFile) throws Exception {
...
-                    "SGD", "NCBITaxon:559292"
-                    );
-            log.info("Running with: " + fms.getDataSubType() + " " + map.get(fms.getDataSubType()));
+            log.info("Running with: " + fms.getDataSubType() + " " + modTaxons.get(fms.getDataSubType()));
 
             bulkLoadFile.setRecordCount(diseaseData.getData().size());
             bulkLoadFileDAO.merge(bulkLoadFile);
-            diseaseService.runLoad(map.get(fms.getDataSubType()), diseaseData);
+            diseaseFmsService.runLoad(modTaxons.get(fms.getDataSubType()), diseaseData);
+        } else if(bulkLoadFile.getBulkLoad().getBackendBulkLoadType() == BackendBulkLoadType.DISEASE_ANNOTATION) {
```
oblodgett: I guess I would just drop the Disease Annotation load ALL together.

markquintontulloch:
@oblodgett - the only slight gotcha with separate loads is that should the DQM want to put all disease annotation types in a single file (but load them separately), when the second entity type is loaded the md5sum check shows that the file exists and the load doesn't happen (and the file doesn't get associated with the load).  As such, I've kept the DISEASE_ANNOTATION BackendBulkLoadType in addition to the entity-specific ones.  This allows all 3 entity types to be loaded in a single load, which is what I imagine most people would want anyway if they have put all the entities in a single file.

src/main/java/org/alliancegenome/curation_api/services/AlleleDiseaseAnnotationCrudService.java
```diff
@@ -37,4 +54,77 @@ protected void init() {
...
+    @Transactional
+    public void runLoad(String taxonId, List<AlleleDiseaseAnnotationDTO> annotations) {
+        List<String> annotationIdsBefore = new ArrayList<>();
+        annotationIdsBefore.addAll(alleleDiseaseAnnotationDAO.findAllAnnotationIds(taxonId));
+        annotationIdsBefore.removeIf(Objects::isNull);
+        
+        log.debug("runLoad: Before: " + taxonId + " " + annotationIdsBefore.size());
+        List<String> annotationIdsAfter = new ArrayList<>();
+        ProcessDisplayHelper ph = new ProcessDisplayHelper(10000);
+        ph.startProcess("Disease Annotation Update " + taxonId, annotations.size());
```
oblodgett: Might want to change this to display the type that is running (AGM) and not just "Disease Annotation"

oblodgett (changes requested):

oblodgett:
But that could be true for adding in Gene's and Alleles and everything else that would fit in a "set"... So what I think might be better is to change DISEASE_ANNOATION -> FULL_INGEST then we will just go through ALL the sets and do the appropriate load for each one.

src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadJobExecutor.java
oblodgett: I see both the list and the enum now?

src/main/java/org/alliancegenome/curation_api/services/AlleleDiseaseAnnotationCrudService.java
markquintontulloch: Done

markquintontulloch:
> But that could be true for adding in Gene's and Alleles and everything else that would fit in a "set"... So what I think might be better is to change DISEASE_ANNOTATION -> FULL_INGEST then we will just go through ALL the sets and do the appropriate load for each one.

I've added FULL_INGEST but also kept DISEASE_ANNOTATION.  The reason being that DQMs are used to submitting all disease annotations as a single file, but other entities separately.  This would work with the WB workflow, for example, where it would be easy to generate all disease annotations together, but not so easy to generate all submissions in a single file.

src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadJobExecutor.java
markquintontulloch: Yes, not sure how to deal with this one.  The map is still there for the FMS load.  The datatype for this is a string, not an enum.  I don't know if we want to change this.  Is there a reason for this being different to the backendbulkloadtype used for the manual load?

src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkLoad.java
```diff
@@ -74,15 +75,32 @@
...
-        GENE_DTO, ALLELE_DTO, AGM_DTO, DISEASE_ANNOTATION_DTO,
-        GENE, ALLELE, AGM, DISEASE_ANNOTATION,
-        ONTOLOGY, MOLECULE
+        GENE_DTO, ALLELE_DTO, AGM_DTO, DISEASE_ANNOTATION_DTO, DISEASE_ANNOTATION,
+        GENE, ALLELE, AGM, AGM_DISEASE_ANNOTATION, ALLELE_DISEASE_ANNOTATION, GENE_DISEASE_ANNOTATION,
+        ONTOLOGY, MOLECULE, FULL_INGEST
         ;
     }
     
     public enum BackendBulkDataType {
```
oblodgett: I think pulling this out into its own file for the enum.

oblodgett (approved):

PR #286, opened 2022-02-14
Flyway investigation SCRUM-1054

oblodgett:


oblodgett:
@mluypaert do we want to get this in before or after the demo?

mluypaert:
> @mluypaert do we want to get this in before or after the demo?

What difference would it make? If none, then we maybe can include i right after the demo and include the baseline at that moment in this same PR?

oblodgett:
It just depends on whether you are going to take this to beta this afternoon / tomorrow or two weeks from now. As soon as it goes in we will need to start making migrations where appropriate, for alpha.

mluypaert:
Well, if we have the code version and the matching baseline schema, then we can deploy the version demoed today to beta any time in the coming two weeks I assume? I think we need to start using flywayDB as soon as possible.

mluypaert:
@oblodgett I think we want to use the DB schema for `v0.3.0-rc1` as base line, rather than [v0.2.0](https://github.com/alliance-genome/agr_curation/pull/286/files#diff-af2c33bb66f1c61c91b5ffa197d9a46ae633229f9399099b2b7974f62f4846e7) currently included?

oblodgett:
I don't think there are any changes... but I can have a look.

mluypaert:
> I don't think there are any changes... but I can have a look.

I know for a fact there are (we discovered some while working on the DB migration for `v0.3.0-rc1`). I can provide you the schema dump I took off alpha right after yesterday's sprint review if needed (but I don't think any DB changes have been introduced since yesterday's demo?)

oblodgett:
@mluypaert this file has to be named: vx.y.q__agr_curation_api.sql so these schema versions can't really match our deployment versions.

oblodgett:
@mluypaert generated the schema file from `v0.3.0-rc1` note this is not a database schema files directly this is a hibernate generated schema file.

mluypaert:
> @mluypaert this file has to be named: vx.y.q__agr_curation_api.sql so these schema versions can't really match our deployment versions.

@oblodgett If the migration files have to be named with a full release version in its name, and those files cannot be edited and reloaded after initial loading (as that could introduce error if not all queries are idempotent), how are we supposed to incrementally build up to a release then? Or is this naming limitation only applicable to the baseline schema file?

mluypaert:
> @mluypaert generated the schema file from `v0.3.0-rc1` note this is not a database schema files directly this is a hibernate generated schema file.

@oblodgett You'll need to generate the schema file from the current alpha instead (what will become `v0.3.0-rc2` and supposably `v0.3.0` on production later), as patch/migration files that would be required for deployment to beta cannot be applied to alpha anymore.

oblodgett:
@mluypaert the point here is to get a baseline of alpha. And start tracking any changes that happen in alpha from this point forward. On deploying this to beta, considering that there haven't been any DB changes to alpha this sprint then the next deployment to beta this might work. However considering this whole thing is still not in production it might be that we have to drop all three databases in order to get this in place, but as of right now this is the schema file for alpha and v0.3.0-rc1 as there is no difference. The other way to handle this is to have a blank baseline file and just assume the database the way it is, and only start making changes from this point forward.

mluypaert:
> @mluypaert the point here is to get a baseline of alpha. And start tracking any changes that happen in alpha from this point forward.

@oblodgett I get it that the baseline file must match the current alpha database, but that's exactly the problem: the current baseline file does not match the current alpha schema (see below).

> On deploying this to beta, considering that there haven't been any DB changes to alpha this sprint then the next deployment to beta this might work. However considering this whole thing is still not in production it might be that we have to drop all three databases in order to get this in place, but as of right now this is the schema file for alpha and v0.3.0-rc1 as there is no difference.

There have been database schema changes on alpha since v0.3.0-rc1, and the current [baseline src/main/resources/db/migration/v0.3.0__agr_curation_api.sql file](https://github.com/alliance-genome/agr_curation/pull/286/commits/277543ee5eb514388f00ddb6d3d57b550f0bb713#diff-d10de1452c9ca1af27666257f3f5b92fa887e2cc815d69897c98416564142128) file does not contain these changes. For example, the new `BulkLoadFileHistory` class you introduced [last week](https://github.com/alliance-genome/agr_curation/pull/289) is not represented here, while it has a table in the current alpha database, along with some other changes to the `gene` and `diseaseannotation` tables.

> The other way to handle this is to have a blank baseline file and just assume the database the way it is, and only start making changes from this point forward.

I think having a baseline file that clearly documents the state of the database schema at the start of using flyway is much prefered over a blank file, as this can then be used for future reference during future troubleshooting (the any DB schema weirdness apear) and during manual deployment of this version to beta and production (as it provides a stable snapshot of the state of the DB to be deployed while the schema of the live alpha DB can move on during deployment to beta and production).
But: if that file can be empty, then why does it have to be a hibernate generated file rather than a DB schema dump?

mluypaert (approved):

PR #289, opened 2022-02-16
Load reporting

oblodgett:
Some functionality changes...
1. When clicking on the upper level load, before it would go out grab the file and check if it was already loaded. Now it does that but "force" loads it anyway.
2. The lower play button still does the same thing and reruns that exact file, but not grabbing it from the source but rather from a backup on S3
3. All the Load URL's rather then returning the string "OK" now return a BulkLoadFileHistory object with much info on how the load went... (I would guess this is why integration tests are failing right now)
4. Removed the "/find" method that was defined in almost ALL the interfaces went back to the base one and change FieldsOnly to FieldsAndLists
5. Moved The file extension to not be in the middle of Load Type and Ontology Type
6. When running postgres locally now the data goes into a docker volume this way the local container does not run out of space. This will also help persist the data better. If needed `docker volume rm data` in order to get rid of it.


src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkLoadFileHistory.java
```diff
@@ -0,0 +1,73 @@
...
+import lombok.*;
+
+@Audited
+@Entity
+@Data
+@AllArgsConstructor @NoArgsConstructor
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
+@ToString(exclude = {"bulkLoadFile"}, callSuper = true)
+@TypeDef(name = JsonTypes.JSON_BIN, typeClass = JsonBinaryType.class)
+public class BulkLoadFileHistory extends BaseGeneratedEntity {
```
oblodgett, line 29: Need to add this to LinkML

adamgibs (approved):

PR #290, opened 2022-02-17
Fix for autocomplete fields SCRUM-1228

oblodgett:


oblodgett:
@mluypaert wondering why these checks are being skipped?

mluypaert:
> @mluypaert wondering why these checks are being skipped?

Updating the title triggered those skipped tests, but the first batch of tests ran and passed
image

oblodgett:
Oh weird ok...

kthorat-prog (approved):

mluypaert:
> Oh weird ok...

Ya, that's an unfortunate side-effect of enabling retriggering tests for PR that get their base branch changes (the only way is triggering on all updates and skip tests when the base branch wasn't changed)

PR #295, opened 2022-02-21
Evidence code filter changed to MultiSelect dropdown

kthorat-prog:


oblodgett (commented):
Do we not want the curators to search by name or curie?

oblodgett:
> Do we not want the curators to search by name or curie?

Oh never mind I see we are changing it to: FilterMultiSelectComponentTemplate

kthorat-prog:
I think search only by abrevation if we want the multiselect dropdown here.

oblodgett:
I haven't seen the ticket but this is going to imply that they can only search via abbreviation in the pull down... and when they see TAS they won't know which one to choose?

kthorat-prog:
Ohh yes, We can wait on this one then until Chris is back.

oblodgett:
@kthorat-prog what is happening with this PR?

kthorat-prog:
Need to think on all options and the best way to implement this. So, closing this PR for now.

PR #305, opened 2022-02-25
UI reskin

adamgibs:
This seems to be working to me. But it may make sense for somebody else to pull it down and test it before merging into alpha though just to be sure.


src/main/cliapp/gulpfile.js
```diff
@@ -1,17 +0,0 @@
...
-    .pipe(replace('.js"></script>', '.js" inline></script>'))
-    .pipe(replace('rel="stylesheet">', 'rel="stylesheet" inline>'))
-    .pipe(
-      inlinesource({
-        compress: false,
-        ignore: ["png"],
-      })
-    )
-    .pipe(gulp.dest("./build"));
-});
```
oblodgett, line 17: @adamgibs I do believe we are going to need this file for building the app... would need to verify.

src/main/cliapp/public/assets/demo/data/customers-large.json
```diff
@@ -1,3206 +1,3604 @@
-
```
oblodgett, line 1: This file probably needs to be deleted.

src/main/cliapp/public/assets/demo/data/customers-medium.json
```diff
@@ -1,56 +1,804 @@
-
```
oblodgett, line 1: This file probably needs to be deleted.

src/main/cliapp/public/assets/demo/data/events.json
```diff
@@ -3,66 +3,66 @@
     	{
             "id": 1,
     		"title": "All Day Event",
-    		"start": "2023-01-03"
```
oblodgett, line 6: This file probably needs to be deleted?

src/main/cliapp/public/assets/demo/data/photos.json
```diff
@@ -0,0 +1,94 @@
+{
```
oblodgett: This file probably needs to be deleted?

src/main/cliapp/public/index.html
```diff
@@ -2,13 +2,14 @@
 <html lang="en">
     <head>
         <meta charset="utf-8" />
-        <title>Alliance Curation Site</title>
+        <title>PrimeReact - Sakai</title>
```
oblodgett: Probably should fix this.

oblodgett (changes requested):
I think we need to go through the assets/demo directory and get rid of stuff that we don't need.... there are a lot of large pictures in there etc.

adamgibs:
@oblodgett Yeah, that makes sense to me. Do we need anything from that directory? Seems like public/asses/demo could just be totally wiped.

oblodgett (approved):

PR #307, opened 2022-03-03
SCRUM-1279

markquintontulloch:
Loading and validation code for additional fields for disease annotations (bulk upload + updates).
Test file updates.
Integration tests for disease annotations and experimental conditions.

Needs some database changes before merging - not sure where we are with Flyway
- Delete empty predicate column from genediseaseannotation, allelediseaseannotation, and agmdiseaseannotation
- data_provider changed from list to string
- add "Note types" vocabulary and "Disease annotation summary" term (+ presumably others)


src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkLoad.java
```diff
@@ -71,7 +71,7 @@
         FORCED_RUNNING,
         
         
-
+        RUNNING,
```
oblodgett: This seems like a random edit?

oblodgett: Otherwise everything looks good.

markquintontulloch: I was getting loads of errors in my terminal about RUNNING enum missing.  It's specified in the LinkML enum, so I added it - errors disappeared.

oblodgett: Ahh ok... this is where a database migration needs to come in. Chnage any bulkload status or bulkloadfile status from `RUNNING` to `FINISHED`

markquintontulloch: Hmm, removed enum again and no more errors.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmDiseaseAnnotationExecutor.java
```diff
@@ -52,6 +53,7 @@ public void runLoad(BulkLoadFile bulkLoadFile) {
     }
 
     // Gets called from the API directly
+    @Transactional
```
oblodgett: Generally speaking, you want the Service methods to be transactional not the caller of them.

src/main/java/org/alliancegenome/curation_api/jobs/executors/AlleleDiseaseAnnotationExecutor.java
```diff
@@ -51,6 +52,7 @@ public void runLoad(BulkLoadFile bulkLoadFile) {
     }
 
     // Gets called from the API directly
+    @Transactional
```
oblodgett: So like the upsert method should be transactional... otherwise what also happens is this WHOLE update turns into 1 transaction and I think it would be better to have partial?

src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmDiseaseAnnotationExecutor.java
markquintontulloch: Done

src/main/java/org/alliancegenome/curation_api/jobs/executors/AlleleDiseaseAnnotationExecutor.java
markquintontulloch: Done

PR #308, opened 2022-03-03
Fix file history SCRUM-1280 SCRUM-1281

oblodgett:
This should fix both bugs.


oblodgett:
@markquintontulloch @mluypaert This should fix both issues. Did test the double run issue on my machine and added the protections to prevent it from happening. @adamgibs this might have conflicts with the Reskin PR but we can deal with that at that time.

adamgibs (approved):

PR #314, opened 2022-03-10
SCRUM-1319 Add AuditedObject fields to disease annotations

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/entities/DiseaseAnnotation.java
```diff
@@ -135,6 +136,23 @@
...
+    @JsonView(View.FieldsOnly.class)
+    private String modifiedBy;
+    
+    @Basic
+    @JsonView(View.FieldsOnly.class)
+    private OffsetDateTime creationDate;
+    
+    @Basic
+    @JsonView(View.FieldsOnly.class)
+    private OffsetDateTime dateLastModified;
```
oblodgett, line 155: @markquintontulloch it might be better to use either a ZonedDateTime or a LocalDateTime for these?

markquintontulloch: I went with Offset as that's the format the DQMs have traditionally submitted in.  We can always change that, or convert on our side, but what's the reason for preferring Zoned over Offset?

oblodgett: So just did some reading...

`First, it doesn't make sense (without conversions) to directly compare two dates with full timezone information. Therefore, we should always prefer storing OffsetDateTime in the database over the ZonedDateTime, as dates with a local time offset always represent the same instants in time.`

I didn't know this but looks like we are good to go.

oblodgett (approved):

PR #319, opened 2022-03-11
Scrum 1020

kthorat-prog:


src/main/cliapp/src/containers/dataLoadsPage/DataLoadsComponent.js
```diff
@@ -151,7 +151,7 @@ export const DataLoadsComponent = () => {
   const loadActionBodyTemplate = (rowData) => {
     let ret = [];
 
-    ret.push(<Button key="edit" icon="pi pi-pencil" className="p-button-rounded p-button-warning mr-2" onClick={() => editLoad(rowData)} />);
+    ret.push(<Button key="edit" con="pi pi-pencil" className="p-button-rounded p-button-warning mr-2" onClick={() => editLoad(rowData)} />);
```
oblodgett: @kthorat-prog something is not right with this PR as its reverting code I put in earlier this week.

oblodgett (changes requested):
@kthorat-prog something is not right with this PR as its reverting code I put in earlier this week.

src/main/cliapp/src/containers/dataLoadsPage/DataLoadsComponent.js
kthorat-prog: @oblodgett yes. I see that now. Need to revery commits on this file. There should be total only 6 changed files.

oblodgett (approved):

PR #322, opened 2022-03-11
UI fix for change from enum to VocabularyTerm

markquintontulloch:


src/main/cliapp/src/containers/diseaseAnnotationsPage/DiseaseAnnotationsTable.js
```diff
@@ -292,7 +292,7 @@ export const DiseaseAnnotationsTable = () => {
     let updatedAnnotations = [...props.props.value];
     console.log(updatedAnnotations);
     if (event.value || event.value === '') {
-      updatedAnnotations[props.rowIndex].diseaseRelation = event.value.name;//this needs to be fixed. Otherwise, we won't have access to the other subject fields
+      updatedAnnotations[props.rowIndex].diseaseRelation.name = event.value.name;//this needs to be fixed. Otherwise, we won't have access to the other subject fields
```
oblodgett, line 295: I think here you might want:
`updatedAnnotations[props.rowIndex].diseaseRelation = event.value`
so that the whole object is used not just the name.

PR #323, opened 2022-03-11
Vocab UI fix

markquintontulloch:
Closed other PR and opened this one as it wasn't synching with new commits for some reason.

@oblodgett - your comments on the previous PR have been addressed.

I've disable CRUD validation on the required fields data_provider, created_by, and modified_by as these fields are not in the annotations loaded from the FMS files and cannot yet be added in the UI, so it was throwing errors when trying to save edits to the disease annotation table.

### DIFF
--- a/src/main/cliapp/src/components/ControlledVocabularyMultiSelector.js
+++ b/src/main/cliapp/src/components/ControlledVocabularyMultiSelector.js
@@ -1,22 +1,22 @@
 import React, {useState} from 'react';
-import { Dropdown } from "primereact/dropdown"
+import { MultiSelect } from "primereact/multiselect"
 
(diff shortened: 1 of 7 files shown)


markquintontulloch:
Bunch of UI display and editing thrown in

adamgibs (approved):
The UI changes look good to me. Just a note, these editor components that use the Autocomplete component may be able to use the AutocompleteEditor component that is in the components folder. I still need to work out some kinks with that, but I think going forward it may simply the code somewhat to use a single editor component and it's tooltip rather than new ones for each new field type since these editors do all share a lot of code.  In terms of the Java stuff, @oblodgett may need to look at that.

oblodgett:
Java stuff looked good.

oblodgett (approved):

PR #328, opened 2022-03-15
Switch vocabularyTerm lookup from search to find

markquintontulloch:
Search for 'Annotation notes' was also returning 'Disease note
types' terms, presumably due to the 'note' part.

### DIFF
--- a/src/main/cliapp/src/service/useControlledVocabularyService.js
+++ b/src/main/cliapp/src/service/useControlledVocabularyService.js
@@ -28,7 +28,7 @@ export function useControlledVocabularyService(termType) {
 
   useQuery(['terms', termType],
     () => {
-      return searchService.search("vocabularyterm", 15, 0, null, { "vocabFilter": { "vocabulary.name": { "queryString": termType } } })
+      return searchService.find("vocabularyterm", 15, 0, {"vocabulary.name" : termType } )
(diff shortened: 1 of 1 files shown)


oblodgett (commented):
Does the search endpoint not work?

markquintontulloch:
@oblodgett - It retrieves the 'Disease annotation notes' terms along with the 'Annotation types' terms - presumably due to the shared 'note'.

oblodgett:
Then something is wrong that shouldn't happen... this is used in many places so we would need to look into this issue more.

src/main/cliapp/src/service/useControlledVocabularyService.js
```diff
@@ -28,7 +28,7 @@ export function useControlledVocabularyService(termType) {
 
   useQuery(['terms', termType],
     () => {
-      return searchService.search("vocabularyterm", 15, 0, null, { "vocabFilter": { "vocabulary.name": { "queryString": termType } } })
+      return searchService.find("vocabularyterm", 15, 0, {"vocabulary.name" : termType } )
```
oblodgett, line 31: This is probably better in the long run to pull the list directly from the the database vs ES.

oblodgett (approved):

PR #329, opened 2022-03-15
SCRUM-734 fix

markquintontulloch:
Created GenomicEntity classes and use these for DiseaseAnnotation diseaseGeneticModifier (was previously using BiologicalEntity which only have curie).

### DIFF
--- a/src/main/cliapp/src/components/AutocompleteEditor.js
+++ b/src/main/cliapp/src/components/AutocompleteEditor.js
@@ -36,8 +36,6 @@ export const AutocompleteEditor = (
         queryString: event.query,
         ...((isSubject || isWith) && { tokenOperator: "AND" })
       }
-      console.log(field);
-      console.log(filter);
     });
 
     searchService.search(endpoint, 15, 0, [], { [filterName]: filter, ...otherFilters })
--- a/src/main/cliapp/src/components/AutocompleteEditor.js
+++ b/src/main/cliapp/src/components/AutocompleteEditor.js
@@ -36,6 +36,8 @@ export const AutocompleteEditor = (
         queryString: event.query,
         ...((isSubject || isWith) && { tokenOperator: "AND" })
       }
+      console.log(field);
+      console.log(filter);
     });
 
     searchService.search(endpoint, 15, 0, [], { [filterName]: filter, ...otherFilters })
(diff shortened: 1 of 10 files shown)


src/main/java/org/alliancegenome/curation_api/model/entities/DiseaseAnnotation.java
```diff
@@ -127,7 +127,7 @@
     @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
     @ManyToOne
     @JsonView({View.FieldsOnly.class})
-    private BiologicalEntity diseaseGeneticModifier;
+    private GenomicEntity diseaseGeneticModifier;
```
oblodgett, line 130: This will need a migration?

src/main/cliapp/src/components/AutocompleteEditor.js
```diff
@@ -36,6 +36,8 @@ export const AutocompleteEditor = (
         queryString: event.query,
         ...((isSubject || isWith) && { tokenOperator: "AND" })
       }
+      console.log(field);
+      console.log(filter);
```
oblodgett: Might want to turn these off?

oblodgett (commented):
Not sure if this needs changes?

src/main/cliapp/src/components/AutocompleteEditor.js
markquintontulloch: Done

src/main/java/org/alliancegenome/curation_api/model/entities/DiseaseAnnotation.java
markquintontulloch: @oblodgett - I'm not sure it needs one.  The column name in the diseaseannotation table remains diseasegeneticmodifier_curie and presumably Hibernate will take care of pointing the foreign key to genomicentity rather than biologicalentity?

oblodgett:
@markquintontulloch can you double check that hibernate updated the FK in the database to point to the other table?

markquintontulloch:
> @markquintontulloch can you double check that hibernate updated the FK in the database to point to the other table?

yup
Screenshot 2022-03-15 at 14 17 10

oblodgett (approved):

oblodgett:
I think the true fix for this is outlined in this issue: https://hibernate.atlassian.net/browse/HSEARCH-438 I would suggest not going down this road... because the requirement is going to be to search via symbol.

PR #330, opened 2022-03-15
The changes to reflect in table as new term/vocab added is added.

kthorat-prog:


adamgibs (approved):

oblodgett:
Do we need the empty file: src/appCss.css?

oblodgett (changes requested):
Do we need the empty file: src/appCss.css?

kthorat-prog:
> Do we need the empty file: src/appCss.css?

I accidently committed it..Will delete this one.

oblodgett (approved):

PR #331, opened 2022-03-15
The ID column made non-sortable

kthorat-prog:


src/main/cliapp/src/containers/controlledVocabularyPage/ControlledVocabularyComponent.js
```diff
@@ -392,7 +393,7 @@ export const ControlledVocabularyComponent = () => {
           key={col.field}
           field={col.field}
           header={col.header}
-          sortable={isEnabled}
+          sortable={(col.field == "id")? false : isEnabled}
```
oblodgett: Ahh this is a bug in this file... the code should be:
`sortable={{col.sortable}}`

oblodgett (changes requested):

src/main/cliapp/src/containers/controlledVocabularyPage/ControlledVocabularyComponent.js
kthorat-prog: Yes. Will check and update it. Thanks.

adamgibs: Maybe sortable on the <Column> component should just be set to col.sortable and set the sortable field on the ID column object to false?

oblodgett: Not sure I understand this comment.

kthorat-prog: Please review the File changed now.

adamgibs: @oblodgett I think we just said the same thing. I just didn't notice your comment.

oblodgett (approved):

PR #338, opened 2022-03-16
SCRUM-731 disease qualifiers

markquintontulloch:
Populate multiselect dropdown with current values.
Also renabled secondaryDataProvider column now that RGD is loaded.

### DIFF
--- a/src/main/cliapp/src/components/ControlledVocabularyMultiSelector.js
+++ b/src/main/cliapp/src/components/ControlledVocabularyMultiSelector.js
@@ -13,8 +13,8 @@ export function ControlledVocabularyMultiSelectDropdown({ options, editorChange,
             <MultiSelect
                 value={selectedValues}
                 options={options}
-                onShow={selectedValues}
                 onChange={(e) => onChange(e)}
+                display="chip"
                 optionLabel="name"
                 placeholder={placeholderText}
                 style={{ width: '100%' }}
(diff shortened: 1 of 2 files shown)


oblodgett (approved):

src/main/cliapp/src/components/ControlledVocabularyMultiSelector.js
```diff
@@ -2,19 +2,19 @@ import React, {useState} from 'react';
 import { MultiSelect } from "primereact/multiselect"
 
 export function ControlledVocabularyMultiSelectDropdown({ options, editorChange, props, placeholderText }) {
-    const [selectedValues, setSelectedValues] = useState();
+    const [selectedValues, setSelectedValues] = useState(props.rowData.diseaseQualifiers);
```
oblodgett, line 5: Actually not sure about passing props in.... typically its either all the variables or just props.... but defer to @adamgibs on this one.

adamgibs: Technically, I don't think it matters. There's just a prop named props. But it could be confusing since there is a convention of passing in a props object that references all of the props passed to the component and then referencing them as `props.propName`. A possible solution could be to rename the props prop to editorProps or something. But in terms of functionality, it shouldn't matter.

PR #345, opened 2022-03-17
SCRUM-1360 fixed dropdown refocus bug

adamgibs:


oblodgett (approved):

markquintontulloch:
Will this also fix the issue for diseaseQualifiers (a MultiSelect dropdown)?

adamgibs:
> Will this also fix the issue for diseaseQualifiers (a MultiSelect dropdown)?

Yeah, it seems to on my local machine. Essentially, setting the diseaseAnnotations state variable was causing the component to re-render. And I guess this updated version to the DataTable focuses on the save icon when it first renders. So the same solution worked for for everything.

markquintontulloch:
nice :)

PR #348, opened 2022-03-18
SCRUM-1377 add paper handle attributes to ConditionRelationDTO

cmpich:


src/main/java/org/alliancegenome/curation_api/model/ingest/dto/ConditionRelationDTO.java
```diff
@@ -13,5 +14,10 @@
     private String conditionRelationType ;
     
     @JsonInclude(JsonInclude.Include.NON_NULL)
-    private List<ExperimentalConditionDTO> conditions;  
+    private List<ExperimentalConditionDTO> conditions;
+
+    private String handle;
+    @JsonProperty("single_reference")
```
oblodgett, line 19: Is this coming from the FMS? I really don't like remapping columns I would rather we just use the name single_reference in java and not use the @JsonPropertiy

oblodgett (approved):

PR #351, opened 2022-03-18
SCRUM-1366 merge from beta to alpha

markquintontulloch:


oblodgett (approved):

mluypaert:
@oblodgett This merge should not have been approved, as Github's way of merging branches in a PR merged alpha into beta first, causing all alpha changes now to be on the beta branch as well. I will revert beta to its previous state, and update the `check-branching-rules` check to block these merges for potential future PRs.

oblodgett:
Sorry I did not catch the `merged 4 commits into alpha from beta` this typically doesn't happen. I think we need to have a way of naming the PR so that we can see its a PR into beta and then also on the merge back to alpha.... typically we would need to take a branch of beta and merge that into alpha.

PR #353, opened 2022-03-19
SCRUM-1369 Hide Parental Populations column in AGM table by default

markquintontulloch:


adamgibs:
@markquintontulloch The comment on this ticket somewhat conflicts with the description and title. If the goal is to remove the column for now, then this code is good and solves that problem. If the goal is to have Parental Population available as a selectable option but unchecked by default, then the solution would be slightly different. We would need a second array of `columnNames `that are selected by default and set `intialTableState.selectedColumnNames` to that array, which would be missing "Parental Population" in this case. `defaultColumnNames `would stay the same. Maybe change its name to be `columnOptions `or something for clarity.

markquintontulloch:
@adamgibs  - isn't that what I've done?   There's now defaultColumnNames and defaultVisibleColumns, the latter of which is what initialTableState.selectedColumnNames gets set to.  Or am I missing your point?

adamgibs:
@markquintontulloch You're right, I just missed that somehow. My brains not working today

adamgibs (approved):

PR #355, opened 2022-03-21
SCRUM-1155 Limit ZECO terms for conditionClass to AGR SLIM

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationFmsService.java
```diff
@@ -295,7 +295,9 @@ private ExperimentalCondition validateExperimentalConditionFmsDTO(ExperimentalCo
         }
         if (dto.getConditionClassId() != null) {
             ZecoTerm term = zecoTermDAO.find(dto.getConditionClassId());
-            if (term == null) return null;
+            if (term == null || term.getSubsets().isEmpty() || !term.getSubsets().contains("ZECO_0000267")) {
```
oblodgett: @markquintontulloch you may want to pull this out into the constants so it can be used in all three classes?

markquintontulloch: @oblodgett agreed, and done

oblodgett (approved):

PR #357, opened 2022-03-21
Template made for all datatables across website

kthorat-prog:


src/main/cliapp/src/components/DataTableHeaderFooterTemplate.js
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+import {Splitter, SplitterPanel} from "primereact/splitter";
+import { Button } from 'primereact/button';
+
+export function DataTableHeaderFooterTemplate({title, tableState, multiselectComponent, buttons, onclickEvent, isEnabled}){
+
+    return (
+        <div className="card">
```
oblodgett: There is a primereact component that generates this:
```code
<Card>
</Card>
```
I think it would be better to use those when we can.

kthorat-prog: This one --> import { Card } from 'primereact/card';  as in <Card> tag?

oblodgett: Yes

kthorat-prog: Sure, will change it.

kthorat-prog: Just committed above change.

oblodgett (approved):

src/main/cliapp/src/components/DataTableHeaderFooterTemplate.js
oblodgett: I know there are other places in the code where we have `<div class="card">` but we can just change them out as we find them. @adamgibs

PR #361, opened 2022-03-22
Scrum 1341

adamgibs:


src/main/cliapp/src/containers/diseaseAnnotationsPage/DiseaseAnnotationsTable.js
```diff
@@ -170,7 +170,7 @@ export const DiseaseAnnotationsTable = () => {
     if (rowData && rowData.with) {
       const sortedWithGenes = rowData.with.sort((a, b) => (a.symbol > b.symbol) ? 1 : (a.curie === b.curie) ? 1 : -1);
       return <>
-        <ul style={{ listStyleType: 'none' }}>
+        <ul style={{ listStyleType: 'none', paddingLeft: 0 }}>
```
oblodgett, line 173: @adamgibs it might be good to look into the Primereact OrderedList and see if those would better styled than trying to do manual css.

oblodgett (approved):

src/main/cliapp/src/containers/diseaseAnnotationsPage/DiseaseAnnotationsTable.js
adamgibs: Yeah, it looks like that component would replace the whole `<ul>` and mapping logic. I should have some extra time today, so I can look into it and see if it make sense to refactor our list cells to use that component. For now I'm going to merge this in as is, since it's working, but I may push out an update to this at some point this sprint.

PR #362, opened 2022-03-22
SCRUM-1318

adamgibs:
This isn't complete yet as it's waiting on the handle field to be available from the backend. However, I'm pushing it out so Chris can take a look at it now.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/model/entities/DiseaseAnnotation.java
+++ b/src/main/java/org/alliancegenome/curation_api/model/entities/DiseaseAnnotation.java
@@ -70,7 +70,7 @@ public class DiseaseAnnotation extends Association {
     @JsonView({View.FieldsAndLists.class})
     private List<EcoTerm> evidenceCodes;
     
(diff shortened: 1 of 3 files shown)


src/main/java/org/alliancegenome/curation_api/model/entities/DiseaseAnnotation.java
```diff
@@ -70,7 +70,7 @@
     @JsonView({View.FieldsAndLists.class})
     private List<EcoTerm> evidenceCodes;
     
-    @IndexedEmbedded(includeDepth = 1)
+    @IndexedEmbedded(includeDepth = 2)
```
oblodgett, line 73: @adamgibs did you test this change locally?

adamgibs: @oblodgett Yep, it seemed to work fine.

oblodgett (approved):

PR #363, opened 2022-03-23
SCRUM-734 property bridge fix

markquintontulloch:
Renabled diseaseGeneticModifier and diseaseGeneticModifierRelation fields.
Changed analyzer on VocabularyTerm name to prevent issues with dropdown filters.


src/main/java/org/alliancegenome/curation_api/model/entities/VocabularyTerm.java
```diff
@@ -26,7 +26,7 @@
 @Schema(name="VocabularyTerm", description="POJO that represents the Vocabulary Term")
 public class VocabularyTerm extends BaseGeneratedEntity {
 
-    @FullTextField(analyzer = "autocompleteAnalyzer", searchAnalyzer = "autocompleteSearchAnalyzer")
+    @FullTextField(analyzer = "whitespace", searchAnalyzer = "whitespace")
```
oblodgett, line 29: Curious the reason for this change? What are autocomplete Analyzers not doing that they should?

markquintontulloch: We don't really want the VocabularyTerm names to be split on underscores when tokenised.  If they are then the filters cannot separate out 'not_ameliorated_by' from 'ameliorated_by' if you only select 'ameliorated_by' in the dropdown menu, for example.  There are other examples in the controlled vocabularies.

oblodgett: Ahh ok... I didn't think about the _'s in the terms

oblodgett (approved):

PR #370, opened 2022-03-24
fix AGM creation logic

cmpich:


src/main/java/org/alliancegenome/curation_api/controllers/crud/AffectedGenomicModelCrudController.java
```diff
@@ -31,4 +37,10 @@ public APIResponse updateAGMs(AffectedGenomicModelMetaDataFmsDTO agmData) {
         return agmFmsExecutor.runLoad(agmData);
     }
 
+    public ObjectResponse<AffectedGenomicModel> create(AffectedGenomicModel entity) {
+
+        return affectedGenomicModelService.createModel(entity);
+    }
+
+
```
oblodgett: Instead of overriding this method in the controller.... can you just override the `create` method in the service?

oblodgett (approved):

PR #381, opened 2022-03-28
SCRUM-1403 Autopopulate 'Modified By' field on editing

markquintontulloch:
Field populated with first and last name

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/helpers/validators/DiseaseAnnotationValidator.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/helpers/validators/DiseaseAnnotationValidator.java
@@ -5,6 +5,7 @@ import java.util.*;
 
 import javax.inject.Inject;
 
+import org.alliancegenome.curation_api.auth.AuthenticatedUser;
 import org.alliancegenome.curation_api.dao.*;
 import org.alliancegenome.curation_api.dao.ontology.*;
 import org.alliancegenome.curation_api.model.entities.*;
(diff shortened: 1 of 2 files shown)


src/main/java/org/alliancegenome/curation_api/services/helpers/validators/DiseaseAnnotationValidator.java
```diff
@@ -198,13 +192,11 @@ public DiseaseAnnotation validateCommonDiseaseAnnotationFields(DiseaseAnnotation
...
         if (createdBy != null) dbEntity.setCreatedBy(createdBy);
 
-        String modifiedBy = validateModifiedBy(uiEntity);
-        if (modifiedBy != null) dbEntity.setModifiedBy(modifiedBy);
-    
         if (uiEntity.getCreationDate() != null)
             dbEntity.setCreationDate(uiEntity.getCreationDate());
         
         dbEntity.setDateLastModified(OffsetDateTime.now());
+        dbEntity.setModifiedBy(authenticatedPerson.getFirstName() + " " + authenticatedPerson.getLastName());
```
oblodgett, line 199: @markquintontulloch im not sure this is the best place to set this... I would suggest not setting, these in the validator but rather in the service.

markquintontulloch: @oblodgett I'd agree, except that unless we do some refactoring we would then need this code in the GeneDiseaseAnnotationService, AlleleDiseaseAnnotationService, and AGMDiseaseAnnotationService as that's where the update method gets called from

oblodgett: Yes I agree... we need this at a much upper level. Also basically advocating for these four fields should be at the base of every class. I would suggest have a base service method that does this and that would work for ALL objects.

markquintontulloch: Feels like that would be part of a larger ticket to refactor code if/when we change the AuditedObject model

oblodgett (approved):

PR #389, opened 2022-03-31
SCRUM-1424 Renamed migration files to clear patch release version number

mluypaert:
This PR moves the migration file numbering into a separate, additional, digit in the filename, to clear the patch release number for actual patch releases (and avoid potential confusion with them).

**Note:** Before merging this PR, manual edits will need to be made in the `flyway_schema_history` table in the alpha postgres database, in order to ensure the records in that table match the new file names, as merging this PR will trigger a new deployment to alpha, which would otherwise try to reapply all existing migrations (if the records don't match the file names in the new deployment)


oblodgett:
Looks like this should work

image

oblodgett (approved):

PR #399, opened 2022-04-05
Date modified and Modified fields are updated immediately on success

kthorat-prog:


markquintontulloch (approved):

oblodgett:
@kthorat-prog @markquintontulloch typically these should be set on the backend

kthorat-prog:
@oblodgett The response of the service is the updated row data. So, I think it is already set on the backend.

src/main/cliapp/src/containers/diseaseAnnotationsPage/DiseaseAnnotationsTable.js
```diff
@@ -315,6 +315,8 @@ export const DiseaseAnnotationsTable = () => {
         let annotations = [...diseaseAnnotations];
         annotations[event.index].subject = data.data.entity.subject;
         annotations[event.index].object = data.data.entity.object;
+        annotations[event.index].dateLastModified = data.data.entity.dateLastModified;
+        annotations[event.index].modifiedBy = data.data.entity.modifiedBy;
```
oblodgett, line 319: @kthorat-prog then with this code shouldn't we just be doing a:
`annotations[event.index] = data.data.entity;`

kthorat-prog: @oblodgett Exactly what I thought but I can't recall why only specifically we did it for subject & object only. Or is it that after adding new fields, they are left out?

I would want to do it in just one line though.

PR #418, opened 2022-04-13
Alpha deployment trigger update proposal

mluypaert:
This codechange should change the deployment to alpha from triggering on every push to alpha (including direct pushing) to only trigger on PR merges. Furthermore, PRs with the `no-deploy` label attached on merge should not trigger a deployment (the GH actions workflow will trigger, but all jobs are skipped).

This PR should serve as an example of its function, as it has the `no-deploy` label attached and thus should not trigger any deployments.

@oblodgett I believe this is what you were after?

### DIFF
--- a/.github/workflows/alpha-build-and-deploy.yml
+++ b/.github/workflows/alpha-build-and-deploy.yml
@@ -1,15 +1,12 @@
 name: Alpha branch Build and Deployment
 on:
-  push:
+  pull_request:
+    types: [closed]
     branches:
       - alpha
-    paths-ignore:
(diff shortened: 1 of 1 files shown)


oblodgett:
So "no-deploy" is the tag to skip deployment?

oblodgett (approved):

mluypaert:
> So "no-deploy" is the tag to skip deployment?

Yes exactly. Note that this change would mean that direct commits to the `alpha` branch also no longer trigger any deployments, and thus every deployment to alpha requires a PR

PR #426, opened 2022-04-22
Removed duplicate code PR into PR

oblodgett:
@markquintontulloch duplicate code has to be maintained twice so worked on getting that removed.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadJobExecutor.java
+++ b/src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadJobExecutor.java
@@ -162,8 +162,9 @@ public class BulkLoadJobExecutor {
                 processTerms(bulkLoadFile, zfsTermService, config);
             } else if(ontologyType == OntologyBulkLoadType.XBA_XBS) {
                 processTerms(bulkLoadFile, OntologyBulkLoadType.XBA, xbaTermService, config);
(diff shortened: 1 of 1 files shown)


markquintontulloch (commented):
I don't think this sets the record counts for the file correctly.  Won't it just give the count for the last load rather than the combined count?

markquintontulloch:
Also, getting a duplicate primaryKey error when loading the 2nd ontology (XBS) using the refactored code
`ERROR: duplicate key value violates unique constraint "ontologyterm_pkey"
  Detail: Key (curie)=(XAO:0004284) already exists.`
  XAO:0004284 is a XBA term

oblodgett:
Yes you are right on the term counts... here is a push to fix that

oblodgett:
Does the first ontology need a specific config for its namespace? I would think XBA is loading all the terms?

markquintontulloch:
XBA should just be loading those terms that match the default namespace

oblodgett:
Ahh ok... might it be better to specify it?

config.getAltNameSpaces().add("xenopus_developmental_annatomy");
				processTerms(bulkLoadFile, OntologyBulkLoadType.XBA, xbaTermService, config);
				config.getAltNameSpaces().clear();
				config.getAltNameSpaces().add("xenopus_developmental_stage");
				processTerms(bulkLoadFile, OntologyBulkLoadType.XBS, xbsTermService, config);

markquintontulloch (approved):

PR #428, opened 2022-04-22
SCRUM-1232 Add additional namespace for XBA load

markquintontulloch:


oblodgett (approved):

markquintontulloch:
need to add it to the controller too - will push that now

PR #430, opened 2022-04-22
Changed exception so they will print in the logs

oblodgett:


mluypaert:
@oblodgett How come [src/main/java/org/alliancegenome/curation_api/exceptions/RuntimeExceptionMapper.java](https://github.com/alliance-genome/agr_curation/pull/430/files#diff-871ee86f60ef95d679ecc3214a834ad5a75459384d57253ef8461a2af813cfa9) got removed, yet no usage of it was removed? Or was this unused and thus simply cleanup?

mluypaert (approved):

PR #436, opened 2022-04-27
SCRUM-1371 + SCRUM-1396: non-loaded gene and agm removal

mluypaert:
See title.


src/main/java/org/alliancegenome/curation_api/jobs/executors/GeneFmsExecutor.java
```diff
@@ -1,66 +1,103 @@
...
 import org.alliancegenome.curation_api.util.ProcessDisplayHelper;
+import org.apache.commons.collections4.ListUtils;
 
 import lombok.extern.jbosslog.JBossLog;
 
 @JBossLog
 @ApplicationScoped
 public class GeneFmsExecutor extends LoadFileExecutor {
 
+    @Inject GeneDAO geneDAO;
```
oblodgett, line 28: So typically its better for the service (geneService) to know how to call the DAO vs using the DAO directly. The service should have a method like `getGeneByCurie(String curie)` which then calls the DAO findByField that way the logic is in the service and can be reused.

src/main/java/org/alliancegenome/curation_api/model/entities/AGMDiseaseAnnotation.java
```diff
@@ -23,7 +23,9 @@ public class AGMDiseaseAnnotation extends DiseaseAnnotation {
 
     @IndexedEmbedded(includeDepth = 1)
     @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
-    @ManyToOne
+    @ManyToOne(cascade = {})
```
oblodgett, line 26: Does the cascade in the @ManyToOne not work? I would think that should work or you might need to cascade = ALL?

src/main/java/org/alliancegenome/curation_api/model/entities/GeneDiseaseAnnotation.java
```diff
@@ -23,7 +23,9 @@ public class GeneDiseaseAnnotation extends DiseaseAnnotation {
 
     @IndexedEmbedded(includeDepth = 1)
     @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
-    @ManyToOne
+    @ManyToOne(cascade = {})
+    @org.hibernate.annotations.OnDelete(action = org.hibernate.annotations.OnDeleteAction.CASCADE)
+    @JoinColumn(foreignKey = @ForeignKey(name="fk_genedasubject"))
```
oblodgett, line 28: Do we need to explicitly define the ForeignKeys?

mluypaert: > Do we need to explicitly define the ForeignKeys?

Excplicitly naming the ForeingKeys with a human readable name makes it much easier and less error prone to do manual or flyway migrations on it afterwards should that need to happen, like in the [flyway migration for this PR](https://github.com/alliance-genome/agr_curation/pull/436/files#diff-248cb3e18ab5ac25b9a33cd8fa7e167f7c3087bb8f6116e66c3abe0b13ce8635), where the only way to figure out the actual name of the ForeignKey to manipulate was looking at my (local) postgres database, and I can only hope that seemingly randomly generate name stays the same accross all environments (which currently seems to be the case).

src/main/java/org/alliancegenome/curation_api/jobs/executors/GeneFmsExecutor.java
mluypaert: Seems to me like you wrote [the removal code for FMS Disease annotation data](https://github.com/alliance-genome/agr_curation/commit/2b8f805471e6c5231958691fc1a605b3dceb3a9d) the same way as I did here, which I essentially just replicated to Allele- (previously), Gene- and AGM removal (this PR)?

src/main/java/org/alliancegenome/curation_api/model/entities/AGMDiseaseAnnotation.java
mluypaert: No, the `@ManyToOne` cannot have the cascade set to all, as that results in the deletion of the agm when a related agmDiseaseAnnotation gets deleted, which is not what we want (we only want it the other way around, deleting an agm should also delete the related agmDiseaseAnnotations). I've tested the code in this PR and this works as intended.

src/main/java/org/alliancegenome/curation_api/jobs/executors/GeneFmsExecutor.java
oblodgett: This code was not "created" in that PR it was moved. Do a search for: `runLoad(String taxonId, List<GeneDiseaseAnnotationDTO> annotations)` and you will see it was moved from the Service to the Executer. Which at that time more refactor should have happened... My point with this, is that for new code we need to keep these kind of principles in mind.

src/main/java/org/alliancegenome/curation_api/model/entities/AGMDiseaseAnnotation.java
oblodgett: I guess I would be in favor then of removing the `@ManyToOne(cascade = {})` -> `@ManyToOne` also you can define the relationship on the other side and put the Cascade.DELETE there. 

More generally, mixing `javax.persistence` annotations with `org.hibernate.annotations` is not a good idea.

oblodgett:
@mluypaert are you trying to get this in for the demo?

mluypaert:
> @mluypaert are you trying to get this in for the demo?

Yes, this is essentially a replication of the code that was introduced in #411, so I was not expecting much need for changes at this point, as all the concepts were already approved then.

oblodgett (approved):

mluypaert:
@oblodgett I think you'll still need to mark the two remaining conversations as resolved if you want to allow me to merge this PR

PR #450, opened 2022-05-04
SCRUM-1508 add condition relation handles table

cmpich:


src/main/java/org/alliancegenome/curation_api/model/entities/ConditionRelation.java
```diff
@@ -41,7 +42,7 @@ public class ConditionRelation extends BaseGeneratedAndUniqueIdEntity {
     @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
     @ManyToOne
     @JsonView({View.FieldsOnly.class})
-    @JsonProperty("single_reference")
+    @JsonProperty("singleReference")
```
oblodgett: @cmpich we don't need the @jsonproperty if its going to be the same name as the field in java, plus its better not to remap names anyway.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/model/entities/ConditionRelation.java
cmpich: Agree.

oblodgett (approved):

PR #452, opened 2022-05-04
Reverted tabs to spaces

oblodgett:
@cmpich your IDE is committing with tabs. I would be more than happy to switch this repo over to using tabs vs spaces, everyone else is using tabs.

### DIFF
--- a/src/main/cliapp/src/components/ListTableCell.js
+++ b/src/main/cliapp/src/components/ListTableCell.js
@@ -1,9 +1,9 @@
 import React from 'react';
 
 export const ListTableCell = ({ template, listData, showBullets }) => {
-	let suppress = 'list-none'
-	if(showBullets)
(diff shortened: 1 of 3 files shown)


cmpich (approved):
That is good. For Java code my IDE is using spaces instead of tabs. Looks like this needs to be configured for javascript files too.

oblodgett:
Can we just switch your ENV to using spaces and let everyone else use tabs? as in that is what everyone is already doing?

cmpich:
I am used to always use 4 spaces per tab to make it always look the same.

oblodgett:
Right... and when using tabs everyone can choose whatever tabstop they want (personally I like 3) for themselves and in your case we create a gitattrbutes that changes all tabs to 4 spaces on checkout and all 4 spaces back to tabs on checkin.

PR #456, opened 2022-05-05
SCRUM-1519 SCRUM-1121 Make security global, remove VPN, and up swap space

oblodgett:


src/main/java/org/alliancegenome/curation_api/config/RestApplication.java
```diff
@@ -3,11 +3,13 @@
 import javax.ws.rs.ApplicationPath;
 import javax.ws.rs.core.Application;
 
+import org.alliancegenome.curation_api.auth.Secured;
 import org.eclipse.microprofile.openapi.annotations.*;
 import org.eclipse.microprofile.openapi.annotations.enums.SecuritySchemeType;
 import org.eclipse.microprofile.openapi.annotations.info.Info;
 import org.eclipse.microprofile.openapi.annotations.security.*;
 
+@Secured
```
oblodgett, line 12: @mluypaert putting @Secured here "turns it on" for the whole API vs endpoint by endpoint.

oblodgett:
@mluypaert there was more UI work then I thought and got @adamgibs to help with it and it should all be taken care of that this point.

mluypaert (approved):

PR #457, opened 2022-05-05
SCRUM-1505 AuditedObject and Person refactor

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/helpers/persons/LoggedInPersonUniqueId.java
```diff
@@ -0,0 +1,18 @@
+package org.alliancegenome.curation_api.services.helpers.persons;
+
+import javax.enterprise.context.RequestScoped;
+
+import org.alliancegenome.curation_api.model.entities.LoggedInPerson;
+import org.alliancegenome.curation_api.services.helpers.CurieGeneratorHelper;
+
+@RequestScoped
+public class LoggedInPersonUniqueId {
+    public String createLoggedInPersonUniqueId (LoggedInPerson loggedInPerson) {
```
oblodgett: @markquintontulloch when I first saw this class I was a little confused... just wondering if you can put `Helper` on the end of its name thanks.

src/main/java/org/alliancegenome/curation_api/services/helpers/validators/ExperimentalConditionValidator.java
```diff
@@ -14,7 +14,7 @@
 import org.apache.commons.lang3.*;
 
 @RequestScoped
-public class ExperimentalConditionValidator {
+public class ExperimentalConditionValidator extends AuditedObjectValidator<ExperimentalCondition> {
```
oblodgett, line 17: I like this approach it puts all these redundant fields up in the super class and has a single spot for all the error messages.

src/main/resources/db/migration/v0.4.0.2__agr_curation_api.sql
```diff
@@ -0,0 +1,230 @@
...
+	DROP COLUMN IF EXISTS created,
+	DROP COLUMN IF EXISTS lastupdated;
+
+ALTER TABLE biologicalentity
+	DROP COLUMN IF EXISTS created,
+	DROP COLUMN IF EXISTS lastupdated;
+	
+ALTER TABLE bulkload
+	ADD COLUMN IF NOT EXISTS created timestamp without time zone,
+	ADD COLUMN IF NOT EXISTS lastupdated timestamp without time zone,
```
oblodgett, line 47: Do we need these columns? Maybe I just need to keep reading.

oblodgett (changes requested):
I guess I have some questions about the migration file, but other than that. Great work on this PR there is a lot going on. Also I guess I would have left the DTO stuff for another PR but sense its part of this one its fine.

src/main/resources/db/migration/v0.4.0.2__agr_curation_api.sql
markquintontulloch: @oblodgett We are getting rid of them, but need to copy the values across to datecreated and dateupdated.  Because the integration tests create the DB from scratch, they don't have those columns so we need to create them (in order to copy nothing and then drop them!).  At least that was my thinking, I may be wrong.

oblodgett: Ahh ok, but also what about all the other tables, and the AUD tables?

src/main/java/org/alliancegenome/curation_api/services/helpers/persons/LoggedInPersonUniqueId.java
markquintontulloch: @oblodgett  - Done

src/main/resources/db/migration/v0.4.0.2__agr_curation_api.sql
markquintontulloch: I don't see these columns in any of the AUD tables.  As for other tables, I don't think it makes sense to copy over the timestamps for all of them - the @UpdateTimestamp and @CreationTimestamp appear to overwrite non-null values so you end up with the time they were loaded into the database rather than any submitted values.

oblodgett:
@markquintontulloch I am going to approve this but I think we need to hold on bring in any PR's given the state of the deployments.

oblodgett (approved):

PR #463, opened 2022-05-09
SCRUM-1509 Java model for BIRT reports

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/entities/curationreports/CurationReportGroup.java
```diff
@@ -0,0 +1,24 @@
...
+@Audited
+@Entity
+@Data @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
+@ToString
+//@ToString(exclude = {"loads"})
+public class CurationReportGroup extends GeneratedAuditedObject {
+    
+    @JsonView({View.FieldsOnly.class})
+    private String name;
+}
```
oblodgett, line 30: @markquintontulloch we will need the @OneToMany relationship here in order to get the reports page to have all the data

src/main/java/org/alliancegenome/curation_api/model/entities/curationreports/CurationReport.java
```diff
@@ -0,0 +1,43 @@
...
+    @JsonView({View.FieldsOnly.class})
+    private Boolean scheduleActive;
+    
+    @ManyToOne
+    private CurationReportGroup curationReportGroup;
+
+    @JsonView({View.FieldsOnly.class})
+    private String birtReportFilePath;
+
+}
```
oblodgett, line 53: We will also need the @OneToMany here for the history's we might break that in the future in that the histories are TOO many but for now it would be good to have it.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/model/entities/curationreports/CurationReportGroup.java
markquintontulloch: @oblodgett - Done (I think!)

src/main/java/org/alliancegenome/curation_api/model/entities/curationreports/CurationReport.java
markquintontulloch: Also done

src/main/java/org/alliancegenome/curation_api/model/entities/curationreports/CurationReportGroup.java
oblodgett: @markquintontulloch the JsonView should probally be `FieldsAndLists` and the `@ToString` should exclude them. so that if you Log.info(curationReport) it doesn't print all the history. Sorry to keep adding.

oblodgett: Will probably need create a custom view for this so we can control it better. (But not as part of this PR)

markquintontulloch: Makes sense - done

oblodgett (approved):

PR #470, opened 2022-05-13
SCRUM-1522 Enable loading of genes, alleles, and AGMs from direct submissions

markquintontulloch:
Alpha will need to be restarted after merge to rebuild database constraints dropped in migration.


src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmExecutor.java
```diff
@@ -0,0 +1,105 @@
...
+        return runLoad(taxonIds, agms);
+    }
+        
+    public APIResponse runLoad(String taxonId, List<AffectedGenomicModelDTO> agms) {
+        List<String> taxonIds = new ArrayList<String>();
+        taxonIds.add(taxonId);
+        return runLoad(taxonIds, agms);
+    }
+            
+    public APIResponse runLoad(List<String> taxonIds, List<AffectedGenomicModelDTO> agms) {
```
oblodgett, line 71: What case does this cover? Just wondering why we would have a list of taxons?

src/main/java/org/alliancegenome/curation_api/dao/GeneDAO.java
```diff
@@ -19,10 +20,46 @@ public Gene getByIdOrCurie(String id) {
...
-        Query jpqlQuery = entityManager.createQuery("SELECT annotation.curie FROM Gene annotation WHERE annotation.taxon.curie=:taxonId");
-        jpqlQuery.setParameter("taxonId", taxonID);
+    public List<String> findAllCuriesByTaxon(String taxonId) {
+        Query jpqlQuery = entityManager.createQuery("SELECT gene.curie FROM Gene gene WHERE gene.taxon.curie=:taxonId");
+        jpqlQuery.setParameter("taxonId", taxonId);
         return (List<String>) jpqlQuery.getResultList();
     }
+    
+    @Transactional
+    public void deleteReferencingDiseaseAnnotations(List<String> geneCuries) {
```
oblodgett, line 30: There may be a better method than this...

src/main/java/org/alliancegenome/curation_api/model/entities/AlleleDiseaseAnnotation.java
```diff
@@ -19,11 +28,12 @@
...
 @Schema(name = "Allele_Disease_Annotation", description = "Annotation class representing a allele disease annotation")
 @JsonTypeName("AlleleDiseaseAnnotation")
+@OnDelete(action = OnDeleteAction.CASCADE)
 public class AlleleDiseaseAnnotation extends DiseaseAnnotation {
 
     @IndexedEmbedded(includeDepth = 1)
     @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
-    @ManyToOne(cascade = {})
+    @ManyToOne
     @org.hibernate.annotations.OnDelete(action = org.hibernate.annotations.OnDeleteAction.CASCADE)
```
oblodgett, line 37: Ahh ok with the class level @OnDelete do we need both of these?

src/main/java/org/alliancegenome/curation_api/model/entities/BiologicalEntity.java
```diff
@@ -26,5 +37,8 @@ public class BiologicalEntity extends CurieAuditedObject {
     @JsonView({View.FieldsOnly.class})
     private NCBITaxonTerm taxon;
     
+    //@OneToMany(mappedBy = "diseaseGeneticModifier", cascade = CascadeType.ALL, orphanRemoval = true, fetch = FetchType.LAZY)
+    //private    List<DiseaseAnnotation> diseaseGeneticModifiers;
+
```
oblodgett: Are these coming back? There is the exclude in the @ToString

src/main/java/org/alliancegenome/curation_api/model/entities/DiseaseAnnotation.java
```diff
@@ -36,7 +59,6 @@
 @Entity
 @Data
 @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
-//@ToString(exclude = {"genomicLocations"})
```
oblodgett, line 39: Was this intentional?

oblodgett (commented):
Whats a fairly large change and it looks good for the most part.

src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmExecutor.java
markquintontulloch: SGD has multiple taxon IDs in their AGM submission

src/main/java/org/alliancegenome/curation_api/dao/GeneDAO.java
markquintontulloch: Probably, but I couldn't find a better way of cascading deletes where @ManyToMany relationships are involved

src/main/java/org/alliancegenome/curation_api/model/entities/BiologicalEntity.java
markquintontulloch: No, I will remove those and the exclude

src/main/java/org/alliancegenome/curation_api/model/entities/DiseaseAnnotation.java
markquintontulloch: It was commented out so I just removed it.

oblodgett: Ahh ok... I think we should bring it back the idea is that every class should have it. Might need to adjust the exclude.

src/main/java/org/alliancegenome/curation_api/dao/GeneDAO.java
oblodgett: Deleting the object should remove all the entries in each of those tables. If not then we need to look at the correct combination of CASCADE DELETE etc. And another option is to Pull the object from the database clear all the lists save it then delete it.

markquintontulloch: I tried various combinations of cascade options but was getting nowhere.  I was trying to follow the logic described here: [https://thorben-janssen.com/avoid-cascadetype-delete-many-assocations/](https://thorben-janssen.com/avoid-cascadetype-delete-many-assocations/)

src/main/java/org/alliancegenome/curation_api/model/entities/DiseaseAnnotation.java
markquintontulloch: It's back now

src/main/java/org/alliancegenome/curation_api/model/entities/BiologicalEntity.java
markquintontulloch: Done

src/main/java/org/alliancegenome/curation_api/model/entities/AlleleDiseaseAnnotation.java
markquintontulloch: My understanding was that the class-level annotation was for cascading the delete of diseaseannotations with gene/allele/agmdiseaseannotations, whereas this is to cascade the deletion of referenced genes.

src/main/java/org/alliancegenome/curation_api/dao/GeneDAO.java
oblodgett: Ahh ok... that is a good tutorial.

src/main/java/org/alliancegenome/curation_api/model/entities/AlleleDiseaseAnnotation.java
oblodgett: ok

oblodgett (approved):

PR #472, opened 2022-05-16
SCRUM-1522 Cascade delete of bulkloadfile to bulkloadhistory

markquintontulloch:
Deletion of bulkloadfile entries fails due to FK constraint to bulkloadhistory.  Need to cascade delete to be able to clean up old FMS loads.

Restart required after initial merge to rebuild constraints following deletion by migration.


mluypaert (approved):

mluypaert:
@markquintontulloch If you would have some time, maybe you can implement a flyway callback to automatically restart the curation application after every successfull migration (as describe [here](https://flywaydb.org/documentation/usage/api/hooks#callsbacks))? That would give hibernate the chance to rebuild anything needed after all migrations, rather than requiring us to remember to restart the application manually after every deployment (including rolling this change forward to beta & prod).

PR #476, opened 2022-05-18
SCRUM-1522 Cascade deletion of BulkLoadHistory to exceptions

markquintontulloch:


mluypaert:
@markquintontulloch You may need a cascade delete here as well?
https://github.com/alliance-genome/agr_curation/blob/d78d7ba0b7ed263728bc58e69b8871af7372ae14/src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkLoadFile.java#L46-L47

markquintontulloch:
@mluypaert  - We don't need that as you can delete the bulkLoadFiles from the UI and once they are all deleted then you can delete the bulkLoad.  I'm not sure we would want to be able to delete the bulkLoad without having checked and manually deleted the bulkLoadFiles first.

mluypaert:
> @mluypaert - We don't need that as you can delete the bulkLoadFiles from the UI and once they are all deleted then you can delete the bulkLoad. I'm not sure we would want to be able to delete the bulkLoad without having checked and manually deleted the bulkLoadFiles first.

Ok, fair point.

mluypaert (approved):

PR #478, opened 2022-05-19
Added code for automatically running reports

oblodgett:


adamgibs (approved):

src/main/java/org/alliancegenome/curation_api/base/entity/AuditedObject.java
```diff
@@ -47,11 +48,13 @@ public class AuditedObject extends BaseEntity {
     @FullTextField(analyzer = "autocompleteAnalyzer", searchAnalyzer = "autocompleteSearchAnalyzer", valueBridge = @ValueBridgeRef(type = OffsetDateTimeValueBridge.class))
     @KeywordField(name = "dateCreated_keyword", sortable = Sortable.YES, searchable = Searchable.YES, aggregable = Aggregable.YES, valueBridge = @ValueBridgeRef(type = OffsetDateTimeValueBridge.class))
     @JsonView({View.FieldsOnly.class})
+    @CreationTimestamp
```
markquintontulloch: I'm not sure we want to use this.  It overwrites any submitted dateCreated value with the time of the load.

src/main/java/org/alliancegenome/curation_api/base/entity/AuditedObject.java
```diff
@@ -47,11 +48,13 @@ public class AuditedObject extends BaseEntity {
...
     private OffsetDateTime dateCreated;
     
     @FullTextField(analyzer = "autocompleteAnalyzer", searchAnalyzer = "autocompleteSearchAnalyzer", valueBridge = @ValueBridgeRef(type = OffsetDateTimeValueBridge.class))
     @KeywordField(name = "dateUpdated_keyword", sortable = Sortable.YES, searchable = Searchable.YES, aggregable = Aggregable.YES, valueBridge = @ValueBridgeRef(type = OffsetDateTimeValueBridge.class))
     @JsonView(View.FieldsOnly.class)
+    @UpdateTimestamp
```
markquintontulloch: Again, not sure we want this due to overwriting of submitted dateUpdated.

markquintontulloch (changes requested):

src/main/java/org/alliancegenome/curation_api/base/entity/AuditedObject.java
oblodgett: if we call an update on a object doesn't matter source I would think we would want this one to fire?

src/main/java/org/alliancegenome/curation_api/base/entity/AuditedObject.java
oblodgett: So in setting the dateCreated manually this annotation overrides it?

oblodgett: Because the alternative is to have to have code everywhere that sets this, or just nulls all over the database.

markquintontulloch: Yeah, it overwrites submitted values.  I've not found a way to set it to only change null values.  I agree that having to set it all over is not ideal, but maybe nulls are acceptable for a lot of classes?

oblodgett: So I think I found a compromise not sure if it works: `@ColumnDefault("current_timestamp")` its suppose to only set it if the value coming in is null.

src/main/java/org/alliancegenome/curation_api/base/entity/AuditedObject.java
markquintontulloch: I think it would also overwrite dateUpdated values coming in from the DQM submissions - I would think that we would want to keep any submitted values rather than have the date of loading

src/main/java/org/alliancegenome/curation_api/base/entity/AuditedObject.java
markquintontulloch: Yeah, that should do what we want

src/main/java/org/alliancegenome/curation_api/base/entity/AuditedObject.java
oblodgett: I would think we would want to indicate that we updated the record when it was loaded into the db? Also the @ColumnDefault wont work on this because saving will just take the incoming value and keep it. So to me anytime we touch a record we should update its timestamp and its modified by.

oblodgett: For instance with MGI we set the updated_by column to the "load" process so that we know it was "last touched" by the load vs a curation. Plus in our case here we have the AUD tables so we can see the changes?

markquintontulloch: Fair enough, if we want the date of loading then this saves a lot of hassle.  We should remove the code where this is explicitly set (I think possibly only the AuditedObjectValidator).

markquintontulloch: Although I do still wonder whether the date of loading is that useful for a curator vs. when the annotation was actually last changed

src/main/java/org/alliancegenome/curation_api/base/entity/AuditedObject.java
```diff
@@ -66,4 +57,61 @@ public class AuditedObject extends BaseEntity {
...
+    // Handle updating the dates
+    // The idea here is if this is a new object and the dates are null these methods will
+    // set the date to "now" before saving on which ever one or both if they are null
+    
+    // If this is an existing object then it HAS to be loaded in order to make an update on it.
+    // In loading the object we save the dateUpdated to previousDateUpdated
+    // 
+    // if the caller does not set a New date then dateUpdated == previousDateUpdated and we
+    // will force an updated date otherwise the caller did set the date and we will use that instead
+
```
oblodgett, line 81: @markquintontulloch so does this logic make sense? And it will happen automatically on every save

markquintontulloch: yeah, I think so

markquintontulloch (approved):

PR #480, opened 2022-05-20
SCRUM-1080 Add userSettings to LoggedInPerson entity

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/entities/LoggedInPerson.java
```diff
@@ -38,4 +44,9 @@ public class LoggedInPerson extends Person {
 
     @JsonView({View.FieldsOnly.class})
     private String apiToken;
+    
+    @Type(type = JsonTypes.JSON_BIN)
+    @Column(columnDefinition = JsonTypes.JSON_BIN)
+    private JsonNode userSettings;
```
oblodgett: @markquintontulloch I think we need this type to be a `Map<String, Object>` instead of `JsonNode`

oblodgett: Maybe it doesn't matter... but I would like to see how they both work. (The API will not be using this field... just passing it through)

markquintontulloch: I've changed it to Map<String,Object>

oblodgett (approved):

PR #481, opened 2022-05-20
SCRUM-1559 Add endpoint for saving user settings

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/interfaces/crud/LoggedInPersonCrudInterface.java
```diff
@@ -0,0 +1,28 @@
...
+
+
+@Path("/loggedinperson")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public interface LoggedInPersonCrudInterface {
+    @POST
+    @Path("/savesettings")
+    @Operation(hidden=true)
+    public void saveSettings(@RequestBody HashMap<String, Object> settings);
```
oblodgett: @markquintontulloch savesettings doesn't need to be hidden because you have the check on the back end to make sure its the authenticated user.

src/main/java/org/alliancegenome/curation_api/services/LoggedInPersonService.java
```diff
@@ -21,6 +23,11 @@ protected void init() {
         setSQLDao(loggedInPersonDAO);
     }
     
+    public void saveSettings(HashMap<String, Object> settings) {
```
oblodgett, line 28: @markquintontulloch this method will need a `@Transactional` and you can persist vs merge for the save.

oblodgett: If fact you probably don't even need the persist or merge considering its inside the transactional any changes will just happen and get persisted back to the db.

markquintontulloch: Done

src/main/java/org/alliancegenome/curation_api/interfaces/crud/LoggedInPersonCrudInterface.java
markquintontulloch: Unhidden

src/main/java/org/alliancegenome/curation_api/interfaces/person/PersonInterface.java
```diff
@@ -0,0 +1,27 @@
...
+
+@Path("/person")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public interface PersonInterface {
+    @POST
+    @Path("/")
+    @Operation(hidden=true)
+    @JsonView(View.FieldsOnly.class)
+    public ObjectResponse<Person> create(Person entity);
```
oblodgett: @markquintontulloch can this create happen in the LoggedIn Person interface and controller? The okta fields are not required.

oblodgett: Then we can get rid of the Person controller and interface all together

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/interfaces/person/PersonInterface.java
markquintontulloch: Done

PR #486, opened 2022-05-23
Added new view for DA update

oblodgett:


oblodgett:
@markquintontulloch effectively this only goes one level deep and includes FieldsOnly... if we were needing it to go deeper we could assign this single view to whatever level we needed to go. It just help fine tune the result object.

markquintontulloch (approved):

PR #491, opened 2022-05-24
SCRUM-1537 make condition relation handle table editable

cmpich:


src/main/java/org/alliancegenome/curation_api/model/entities/ConditionRelation.java
```diff
@@ -47,7 +47,7 @@ public class ConditionRelation extends UniqueIdAuditedObject {
     @IndexedEmbedded(includeDepth = 1)
     @IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
     @ManyToMany
-    @JsonView({View.FieldsAndLists.class})
+    @JsonView({View.FieldsOnly.class})
```
oblodgett: I would guess this is causing the IT's to fail? Maybe we create a separate view for the specific operation.

cmpich: Ok, I created a new update method.

oblodgett: Just override the current one we dont want two

src/main/java/org/alliancegenome/curation_api/services/helpers/validators/ConditionRelationValidator.java
```diff
@@ -1,116 +1,148 @@
...
-        }
-        
-        return conditions;
-    }
+         }
+         conditions.add(condition);
+      }
+
+      return conditions;
+   }
```
oblodgett: Once again the tab vs spaces issue. We need to convert this repo to tabs... then in your editor you can have the tab stop set to whatever you like and on commit it will convert back to tabs.

src/main/java/org/alliancegenome/curation_api/controllers/crud/ConditionRelationCrudController.java
```diff
@@ -31,4 +31,8 @@ public ObjectResponse<ConditionRelation> validate(ConditionRelation entity) {
         return conditionRelationService.validate(entity);
     }
 
+	@Override
+	public ObjectResponse<ConditionRelation> update(ConditionRelation entity) {
+		return super.update(entity);
+	}
```
oblodgett: This doesn't need to be overridden if it does the same as the base class.

src/main/cliapp/src/containers/conditionRelationPage/ConditionRelationTable.js
```diff
@@ -17,326 +17,500 @@ import {ErrorMessageComponent} from "../../components/ErrorMessageComponent";
...
+						  paginator totalRecords={totalRecords} onPage={onLazyLoad} lazy
+						  paginatorTemplate="CurrentPageReport FirstPageLink PrevPageLink PageLinks NextPageLink LastPageLink RowsPerPageDropdown"
+						  currentPageReportTemplate="Showing {first} to {last} of {totalRecords}" rows={tableState.rows} rowsPerPageOptions={[10, 20, 50, 100, 250, 1000]}
+			>
+				<Column field='rowEditor' rowEditor style={{maxWidth: '7rem', minWidth: '7rem'}}
+						  headerStyle={{width: '7rem', position: 'sticky'}} bodyStyle={{textAlign: 'center'}} frozen headerClassName='surface-0'/>
+				{columnList}
+			</DataTable>
+		</div>
+	)
```
oblodgett: @adamgibs @cmpich tab vs spaces again. We need to sort this out or we are going to be flight over commits going forward.

oblodgett (changes requested):
@adamgibs @cmpich we need to sort out the tabs vs spaces

src/main/java/org/alliancegenome/curation_api/controllers/crud/ConditionRelationCrudController.java
cmpich: It needs to be as it overides ConditionRelationCrudInterface class with a different JsonView....
Without I got an abstract methods not implemented error.

oblodgett: But you don't need to over ride it in the controller... just in the interface.

cmpich: I'll try again...

oblodgett: PS so long as the method definition doesn't change.

cmpich:
I have new checkins...

src/main/cliapp/src/containers/conditionRelationPage/ConditionRelationTable.js
```diff
@@ -17,326 +17,500 @@ import {ErrorMessageComponent} from "../../components/ErrorMessageComponent";
...
+                    paginator totalRecords={totalRecords} onPage={onLazyLoad} lazy
+                    paginatorTemplate="CurrentPageReport FirstPageLink PrevPageLink PageLinks NextPageLink LastPageLink RowsPerPageDropdown"
+                    currentPageReportTemplate="Showing {first} to {last} of {totalRecords}" rows={tableState.rows} rowsPerPageOptions={[10, 20, 50, 100, 250, 1000]}
+         >
+            <Column field='rowEditor' rowEditor style={{maxWidth: '7rem', minWidth: '7rem'}}
+                    headerStyle={{width: '7rem', position: 'sticky'}} bodyStyle={{textAlign: 'center'}} frozen headerClassName='surface-0'/>
+            {columnList}
+         </DataTable>
+      </div>
+   )
```
oblodgett: This whole file has changed due to tab / space issue. The repo is set for 2 spaces for indent for JS and 4 spaces for Java. We need to switch this over to tabs and then every developer can choose whatever tabstop they would like.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/interfaces/crud/ConditionRelationCrudInterface.java
```diff
@@ -28,6 +28,11 @@ public interface ConditionRelationCrudInterface extends BaseIdCrudInterface<Cond
     @JsonView(View.FieldsAndLists.class)
     public ObjectResponse<ConditionRelation> create(ConditionRelation entity);
 
+    @PUT
+    @Path("/")
+    @JsonView(View.FieldsAndLists.class)
+    public ObjectResponse<ConditionRelation> update(ConditionRelation entity);
+
```
oblodgett, line 30: @cmpich I think you might need to create a custom view for this... might be why the IT's are failing.

oblodgett (approved):

oblodgett:
@cmpich there is no `service/ConditionRelationService.js` file? Did you not include it?

cmpich:
Thanks. Added the missing file.

oblodgett:
@cmpich The validation for condition relations changed. But the file for testing these were not updated. This is why the IT's are failing.

cmpich:
I noticed that too. Will tend to it tomorrow.

oblodgett:
Probably more specifically the call to verify the reference, as Github does not have access to the literature ES service.

PR #496, opened 2022-06-07
SCRUM-1569 Save null values on editing VocabularyTerm abbreviation/definition

markquintontulloch:
Fixes bug where values for (non-required) abbreviation and definition fields were only saved if non-empty string


oblodgett:
@markquintontulloch is what is coming in from the UI an empty string?

markquintontulloch:
@oblodgett - not sure, but it's theoretically possible at least via the API.  This should cover all bases.

oblodgett (approved):

PR #497, opened 2022-06-07
SCRUM-1534

adamgibs:
-   Puts common JSX and templates in a GenericDataTable Component
-   Puts common logic in a useGenericDataTable hook
-   Adds a FilterComponent that switches between input, dropdown, and multiselct types based on a type prop
--  This includes a change to the filterElement field in the columns array. The field is now an object with three fields: type, fitlerName, fields
-   This component has not yet been applied to the following:
--     ControlledVocabularyComponent:
---       There's an api bug that will cause a ui exception if this change is impmented. Once SCRUM-1603 is complete, this component could be changed over
--    ConditionRelationTable:
---       There is still a an unmerged PR for this table
--     Ontology Tables:
---       The ontology tables already have a generic OntologyTable component. It could be replaced with this, but I don't know if it would be useful


oblodgett:
@adamgibs you are injecting 2 spaces and 4 spaces which is a mix in the files. You git filters should be fixing this you need to decide either on tabs or spaces.

oblodgett (approved):

PR #499, opened 2022-06-07
Convert Spaces to Tabs for repo

oblodgett:


markquintontulloch (approved):

mluypaert:
@oblodgett You may want to ensure to use tabs for indentation, but space for (positional) alignments, otherwise your alignments get skewed when changing the tab size (different developers will use a different default tab size in their IDEs, terminals, ...)

image

mluypaert (approved):

adamgibs (approved):

oblodgett:
With this change we will need to get rid of the tabspace and tabspace2 in our `~/.gitconfig` files otherwise git will revert this change.

PR #511, opened 2022-06-14
SCRUM-1537 validate null handle, disallow same handle/pub combination

cmpich:


src/main/java/org/alliancegenome/curation_api/exceptions/ApiErrorExceptionMapper.java
```diff
@@ -13,7 +13,7 @@ public class ApiErrorExceptionMapper implements ExceptionMapper<ApiErrorExceptio
 	@Override
 	@Produces(MediaType.APPLICATION_JSON)
 	public Response toResponse(ApiErrorException e) {
-		log.error(e);
+		log.info(e);
```
oblodgett, line 16: @cmpich is this not an error?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/exceptions/ApiErrorExceptionMapper.java
cmpich: I want to suppress those outputs as it is used for validation errors and they should not show up in the normal log.

PR #515, opened 2022-06-15
SCRUM-1641 convert list of records to delete into batches

cmpich:


oblodgett:
@cmpich do we need this? Or can we just go with the fix in the submission file?

cmpich:
I think we need to fix this.

markquintontulloch:
If you are going to do this, should probably do it for AGMs and alleles at the same time?

PR #518, opened 2022-06-17
SCRUM-1537 add auto-complete on reference column on condition-relation-handle-table

cmpich:


src/main/cliapp/src/components/AutocompleteEditor.js
```diff
@@ -58,14 +59,14 @@ export const AutocompleteEditor = (
 	};
 
 	const onValueChange = (event) => {
+		console.log(event)
```
oblodgett: Might want to pull out these logs

oblodgett (changes requested):

src/main/cliapp/src/components/AutocompleteEditor.js
cmpich: Yes.

oblodgett (approved):

PR #519, opened 2022-06-20
SCRUM-516 SCRUM 1537 create getEntityType() method to select and display the auto-complete list, auto-complete for DA table

cmpich:


oblodgett:
@cmpich instead of having specific code in the auto complete editor can we pass in a function that does the display for that field. `valueDisplay` function? @adamgibs maybe you have some ideas so that if else doesn't get any larger.

cmpich:
That was exactly what I was aiming for but only went partially into this direction. I'd like to have a switch-statement instead of the if-else statements. I am just not fully sure who is using the different cases..

cmpich:
Could we accept this PR and continue working on it? I'd like to get feedback from Chris early on.

oblodgett:
@chris-grove can we get a story to refactor the autocomplete component to use a valueDisplay function vs if else statements.

oblodgett (approved):

chris-grove:
@oblodgett 
https://agr-jira.atlassian.net/browse/SCRUM-1676
☝️

PR #526, opened 2022-06-27
Upgrade quarkus to fix ontology Endpoints

oblodgett:


markquintontulloch (approved):

mluypaert:
@oblodgett Why did you change so much existing flyway migration files? Changing migration files that have already been applied doesn't retrigger them, and potentially breaks deployments to new environments. Shouldn't those migration files be immutable, and only new migration files added to undo previous changes?

PR #540, opened 2022-07-06
SCRUM-1689 Handling of references

markquintontulloch:
Change the way reference objects are stored in the curation DB.  The following attributes are now stored:
curie - now the AGR curie, was previously the PubMed / MOD curie
submittedCrossReference - primary key of the table, value is the same as what was previously stored as curie
primaryCrossReference - curie to display in curation DB tables (preferentially PubMed curie, otherwise first alphabetically)
secondaryCrossReferences - cross-references other than the primaryCrossReference

Endpoint for synchronisation of references with literature system implemented.  If a reference is no longer found in the literature system then the object on the curation DB side is marked as obsolete.


markquintontulloch:
Wrong branch pushed

PR #541, opened 2022-07-06
SCRUM-1712 Handling of references in curation DB

markquintontulloch:
Changes the way reference objects are stored in the curation DB. The following attributes are now stored:
curie - now the AGR curie, was previously the PubMed / MOD curie
submittedCrossReference - primary key of the table, value is the same as what was previously stored as curie
primaryCrossReference - curie to display in curation DB tables (preferentially PubMed curie, otherwise first alphabetically)
secondaryCrossReferences - cross-references other than the primaryCrossReference

Endpoint for synchronisation of references with literature system implemented. If a reference is no longer found in the literature system then the object on the curation DB side is marked as obsolete.


cmpich (approved):
Looks good.

abecerra (approved):

src/main/resources/db/migration/v0.6.0.1__agr_curation_api.sql
```diff
@@ -0,0 +1,135 @@
...
+	handle varchar (255),
+	reference_curie varchar (255),
+	revtype smallint
+	);
+	
+ALTER TABLE paperhandle
+	DROP CONSTRAINT IF EXISTS "fkb11h1yvb7lchgw07wxspntpsc";
+	
+ALTER TABLE paperhandle
+	RENAME reference_curie TO reference_submittedcrossreference;
```
oblodgett, line 119: @markquintontulloch why do we have a create table then a drop constraint and then a alter? Just create the table correctly? Or if the table already exists just modify it?

src/main/resources/db/migration/v0.6.0.1__agr_curation_api.sql
```diff
@@ -0,0 +1,135 @@
...
+ALTER TABLE reference_secondarycrossreferences
+	ADD CONSTRAINT reference_secondarycrossreferences_submittedcrossreference_fk
+	FOREIGN KEY (reference_submittedcrossreference) REFERENCES reference(submittedcrossreference);
+	
+CREATE TABLE IF NOT EXISTS note_reference (
+	note_id bigint,
+	references_curie varchar (255)
+	);
+
+CREATE TABLE IF NOT EXISTS note_reference_aud (
```
oblodgett, line 37: Also can we get away from using "IF NOT EXISTS" either the table exists or it doesn't?

markquintontulloch: We have to do this because of the integration tests.  I think it's something to do with tables that didn't exist in the base schema.  They weren't explicitly created in the initial migrations (because we were letting Hibernate do that), but now we need to create it.  Outside of the integration tests, the table already exists.

markquintontulloch: Or in most of these cases I think it's because we're still allowing Hibernate to create tables in dev mode.  If you do development and then write the migration file, those tables have already been created once the migration runs.  I don't know, this migration stuff makes my head hurt!  I think in some cases I've added 'IF NOT EXISTS' due to failing migration scripts, but I've possibly been overly liberal with them.  Maybe we should always revert our local database to the same state as alpha before running the migration script locally to avoid this, but does it really matter if we leave some unnecessary 'IF NOT EXISTS's in?

src/main/resources/db/migration/v0.6.0.1__agr_curation_api.sql
markquintontulloch: The paperhandle table was an example of where it exists in the alpha database, but not in the base schema and was not created by the previous migrations.  In this migration we wanted to rename a column and drop the previous FK constrain on that column.  For that to work with the integration tests, we needed to create the table in its previous state first.
I wonder if, now that we've switched Flyway/Hibernate strategies, we need to create new base schema and start the migrations from scratch?

src/main/resources/db/migration/v0.6.0.1__agr_curation_api.sql
oblodgett: In my previous commit with the migration change I went through and made sure that ALL tables were created in the base schema, so that either via hibernate or via migration the DB was the same.

Well the issue I see is that you never know if the statement actual ran or not. Integration tests should not fail due to missing tables as I made sure in my commit a few days ago that everything is in the base schema. And yes I agree that we should be taking the schema as in on Alpha, and then running the migrations against it even if the current code has us creating tables via hibernate.

src/main/resources/db/migration/v0.6.0.1__agr_curation_api.sql
oblodgett: As in the other comment I altered the base schema to create all the table... maybe I missed one? but I did test letting flyway create all the tables, and the validate did work.

oblodgett: #526 this was the one where I made the chnages.

markquintontulloch: I think you missed paperhandle - maybe this is only an issue as I was trying to modify that table in the migration?

mluypaert: @oblodgett Changing migration files that have already been run doesn't rerun them, those files are supposed to be immutable once applied, see [my earlier comment](https://github.com/alliance-genome/agr_curation/pull/526#issuecomment-1173895713) on that ticket.

markquintontulloch: @mluypaert Doesn't it change things for the integration tests though, which is the issue?  In that case the database is build from scratch, so presumably all migrations are rerun?

oblodgett: Yes it does change the integration tests. So the idea is to make sure that the base mirrors what is currently in production. And any further migrations should build upon the current DB vs fixing issues that might have been missing in the base.

PR #548, opened 2022-07-11
SCRUM-1733 Fix bug with manual entry for Reference fields

markquintontulloch:
Previously, manually entering a value in the Reference field (instead of picking from the Autocomplete) caused a blank screen error.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/ReferenceService.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/ReferenceService.java
@@ -90,7 +90,6 @@ public class ReferenceService extends BaseCrudService<Reference, ReferenceDAO> {
 	}
 		
 	protected Reference copyLiteratureReferenceFields(LiteratureReference litRef, Reference ref, String searchCurie) {
-		log.info(litRef);
 		ref.setCurie(litRef.getCurie());
 		
 		ArrayList<String> otherXrefs = new ArrayList<String>();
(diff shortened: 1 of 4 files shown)


markquintontulloch:
Had to limit the validation of the submitted reference ID to the first 50 results returned by the literature system search.  This is a problem for IDs such as 'PMID:2' - this result would return nearly 0.5 million search result and so the validation reports it as an invalid reference as it doesn't find the result quickly enough.

We could reduce the chances of this false invalidation by increasing the no. of results returned but this increases the delay between hitting save and the reporting of success/failure.  Too big a number of results causes a socket timeout error.

My proposal to avoid this is to import all references from the Literature Service, and resynchronise on a fairly regular basis.  That way, most IDs can be validated on our side without needing to call the LiteratureService.  If an invalid ID were entered, or a valid ID that had not yet been synched with the curation DB, then the LiteratureService would not be called - but the search ID should be specific enough (e.g. not low PMID numbers causing all curies containing 'PMID' and the number 2 to be returned) to avoid large numbers of results and the potential for an ID to be incorrectly called invalid.

src/main/java/org/alliancegenome/curation_api/services/ReferenceService.java
```diff
@@ -107,8 +111,10 @@ protected Reference copyLiteratureReferenceFields(LiteratureReference litRef, Re
 			} else {
 				ref.setPrimaryCrossReference(otherXrefs.get(0));
 			}
+		} else {
+			ref.setSecondaryCrossReferences(otherXrefs);
 		}
-		
+		log.info(ref);
```
oblodgett: Its also best these log lines go in as debugs so the logs don't fill up.

oblodgett (commented):
Are all these extra fields defined in LinkML?

oblodgett:
Also I agree that this should be turned into a load that runs each night.

markquintontulloch:
> Are all these extra fields defined in LinkML?

Not as yet.  Wasn't sure whether we wanted to bother adding fields that are only used internally to the curation DB.

src/main/java/org/alliancegenome/curation_api/services/ReferenceService.java
markquintontulloch: Done

markquintontulloch:
Closing PR due to change in approach, fix will be encompassed by work on SCRUM-1712

PR #549, opened 2022-07-12
SCRUM-1724 create endpoint to generate AGMDiseaseAnnotationDTO

cmpich:


oblodgett:
Do we not have endpoints for these?

cmpich:
No. we don't. We have an endpoint to load bulk data but I cannot use it as it basically drops and re-loads according to the submitted file.

cmpich:
I need an endpoint to only load one AgmDiseaseAnnotationDTO object.

oblodgett:
And you don't want to use the endpoint for the AgmDiseaseAnnotation object directly?

cmpich:
No, because I don't have all the internal / curie ids.

oblodgett:
I would prefer to not go down the road of using the DTO's. Those are only for ingest data and will go away once a mod moves over. So it would mean that we have 2 create methods for each data type. It would be better to use only one create API call.

src/main/java/org/alliancegenome/curation_api/base/services/BaseDTOCrudService.java
```diff
@@ -0,0 +1,15 @@
...
+import org.alliancegenome.curation_api.base.BaseDTO;
+import org.alliancegenome.curation_api.base.dao.BaseEntityDAO;
+import org.alliancegenome.curation_api.base.entity.BaseEntity;
+import org.alliancegenome.curation_api.exceptions.ObjectUpdateException;
+
+
+public abstract class BaseDTOCrudService<E extends BaseEntity, T extends BaseDTO, D extends BaseEntityDAO<E>> extends BaseEntityCrudService<E, D> {
+
+	protected abstract void init();
+	public abstract E upsert(T dto) throws ObjectUpdateException;
```
oblodgett, line 12: @cmpich if you need to have an ObjectResponse we would need to change this right now they all return just the Entity from the DB.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/base/services/BaseDTOCrudService.java
oblodgett: `E` -> `ObjectResponse<E>`

PR #553, opened 2022-07-14
SCRUM-1756 Fix saving of notes attached to DAs

markquintontulloch:
Saving of new notes attached to DAs is broken on alpha.  It's not clear to me how this is working on beta, or how that has been lost for alpha, but this fixes it!

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/helpers/validators/DiseaseAnnotationValidator.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/helpers/validators/DiseaseAnnotationValidator.java
@@ -188,7 +188,6 @@ public class DiseaseAnnotationValidator extends AuditedObjectValidator<DiseaseAn
 		
 		List<Long> previousNoteIds = dbEntity.getRelatedNotes().stream().map(Note::getId).collect(Collectors.toList());
 		List<Long> validatedNoteIds = validatedNotes.stream().map(Note::getId).collect(Collectors.toList());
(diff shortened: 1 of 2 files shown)


src/main/java/org/alliancegenome/curation_api/services/helpers/validators/DiseaseAnnotationValidator.java
```diff
@@ -186,6 +188,12 @@ public List<Note> validateRelatedNotes(DiseaseAnnotation uiEntity, DiseaseAnnota
 		
 		List<Long> previousNoteIds = dbEntity.getRelatedNotes().stream().map(Note::getId).collect(Collectors.toList());
 		List<Long> validatedNoteIds = validatedNotes.stream().map(Note::getId).collect(Collectors.toList());
+		List<Long> idsToPersist = ListUtils.subtract(validatedNoteIds, previousNoteIds);
```
cmpich: What is the variable idsToPersist used for? Seems to not be used anywhere...

markquintontulloch: Yep, forgot to remove it - will do so now

markquintontulloch: @cmpich - now removed.  Also, integration test added for good measure.

oblodgett (approved):

PR #554, opened 2022-07-15
SCRUM-1555 Edit existing disease annotation 'condition_relations' (no handle in use)

markquintontulloch:


adamgibs (approved):
Looks good. The validation and error display on rowEditSave is definitely better.

PR #557, opened 2022-07-18
SCRUM-1712 Fix synchronisation of references with literature service

markquintontulloch:
Implemented findAllCuries method for Reference
Moved synchronisation methods to helper


src/main/java/org/alliancegenome/curation_api/dao/ReferenceDAO.java
```diff
@@ -33,5 +42,30 @@ protected void deleteReferenceForeignKey(String table, String column, String ori
...
+			allQuery.setFirstResult(first);
+			allQuery.setMaxResults(pagination.getLimit());
+		}
+		SearchResponse<String> results = new SearchResponse<String>();
+		
+		List<String> resultCuries = allQuery.getResultStream().map(Reference::getCurie).collect(Collectors.toList());
+		results.setResults(resultCuries);
+		results.setTotalResults(totalResults);
+		return results;
+	}
```
oblodgett: @markquintontulloch I was thinking of turning this into a base function so all dao's could use it. Just put it in the BaseSQLDao and change `<Reference>` to `<E>`

oblodgett (approved):

PR #559, opened 2022-07-18
SCRUM-1765 Bring editing of DA relatedNotes in line with conditionRelations

markquintontulloch:
Validation occurs on row edit save rather than on clicking 'Keep Edits'.
'Keep Edits' button not enabled unless valid edits in place.
Fixes bugs in display of error messages.
Also fixes bug on conditionRelations editing where the 'Keep Edits' button was enabled when invalid edits were in place, thus bypassing validation.


markquintontulloch:
Commits on correct branch this time!

adamgibs (approved):

PR #586, opened 2022-07-29
SCRUM-1799 Deletion of experimental conditions

markquintontulloch:


adamgibs (approved):

src/main/java/org/alliancegenome/curation_api/dao/base/BaseSQLDAO.java
```diff
@@ -147,18 +163,41 @@ public E merge(E entity) {
...
+		} catch (PersistenceException e) {
+			handlePersistenceException(entity, e);
+		}
 		return entity;
 	}
 	
+	private void handlePersistenceException(E entity, Exception e) {
+		ObjectResponse<E> response = new ObjectResponse<E>(entity);
+		if (e.getCause() instanceof ConstraintViolationException) {
+			ConstraintViolationException cve = (ConstraintViolationException) e.getCause();
```
oblodgett: Might want to just double check that you are getting the "rootCause" vs any cause it would then list the table and constraint that is violated.

oblodgett: I also agree with your slack message this would be better handled in the controller where it could craft a better error message. At this point in the code there is no indication of who the caller is or what happened.

src/main/java/org/alliancegenome/curation_api/dao/base/BaseSQLDAO.java
```diff
@@ -147,18 +163,44 @@ public E merge(E entity) {
...
+			entityManager.flush();
+		} catch (PersistenceException e) {
+			handlePersistenceException(entity, e);
+		}
 		return entity;
 	}
 	
+	private void handlePersistenceException(E entity, Exception e) {
+		ObjectResponse<E> response = new ObjectResponse<E>(entity);
+		Throwable rootCause = e.getCause();
```
oblodgett, line 192: @markquintontulloch this should get changed to getRootCause() plus this code shouldn't be here in the DAO.

markquintontulloch: @oblodgett - getRootCause() doesn't appear to exist.  I agree about the code not being in the DAO, but moving it to the controller seemed like a big job outside the scope of this ticket as it affects everything that calls the DAO remove method.

PR #599, opened 2022-08-08
SCRUM-1850 integration tests

markquintontulloch:


mluypaert (changes requested):
This PR seems to contain much more than only the integration tests for new Experimental condition creation, it also contains changes in the Disease annotation validators (alllele, gene agm and general), the notevalidator, the ConditionRelationValidator and even some in the ExperimentalConditionValidator? Seems to me the experimental condition integration tests can all be cherry-picked from d23cdeb663?

markquintontulloch:
The changes to all the validators are the same - a fault in the logic for validation of obsolete terms/entities where null objects were not accounted for.  I guess I could pull all of that out into its own PR but it might as well go in with this stuff as it was doing the ExperimentalCondition integration tests that identified the bug.  Do we really need to create extra work?

mluypaert:
> The changes to all the validators are the same - a fault in the logic for validation of obsolete terms/entities where null objects were not accounted for. I guess I could pull all of that out into its own PR but it might as well go in with this stuff as it was doing the ExperimentalCondition integration tests that identified the bug. Do we really need to create extra work?

Separation of concerns makes for easier review and testing. If you want to be able to trace these additional changes and test them on alpha independently, then yes, I'd recommend to make a separate PR for merging those changes to alpha instead, and only merge the relevant new experimental condition integration tests in this branch (I can cherry-pick them from the commit mentioned earlier).
If these changes don't require any further testing or approval, then fair enough, you can leave them in this PR.

PR #605, opened 2022-08-09
SCRUM-1711 API work for Disease Annotation creation

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/DiseaseAnnotationCurie.java
```diff
@@ -27,6 +27,10 @@ public static String getConditionRelationUnique(ConditionRelation relation) {
 	}
 
 	public abstract String getCurieID(DiseaseAnnotationDTO annotationDTO);
+	
+	public abstract String getCurieID(GeneDiseaseAnnotation annotation);
+	public abstract String getCurieID(AlleleDiseaseAnnotation annotation);
+	public abstract String getCurieID(AGMDiseaseAnnotation annotation);
```
oblodgett: @markquintontulloch curious why you need these three? Does the other function work its doing the same thing... there is a lot of duplicate code here?

src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/WormDiseaseAnnotationCurie.java
```diff
@@ -28,4 +32,27 @@ public String getCurieID(String subject, String object, String reference, List<S
...
+	public String getCurieID(AlleleDiseaseAnnotation annotation) {
+		return getDiseaseAnnotationCurieID(annotation.getSubject().getCurie(), annotation);
+	}
+
+	@Override
+	public String getCurieID(AGMDiseaseAnnotation annotation) {
+		return getDiseaseAnnotationCurieID(annotation.getSubject().getCurie(), annotation);
+	}
+	
+	private String getDiseaseAnnotationCurieID(String subjectCurie, DiseaseAnnotation annotationDTO) {
```
oblodgett: @markquintontulloch should probably pull DTO off the name here as its an entity and not a DTO.

markquintontulloch: done

src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/DiseaseAnnotationCurie.java
markquintontulloch: We need the subject field to generate the curie, and that's not available from the core DiseaseAnnotation class - or am I missing a trick here?

oblodgett: Yes thats true but every DiseaseAnnotation has a curie... putting in a PR to this so you can see what I mean.

oblodgett:
@markquintontulloch did this PR mess up the IT's?

markquintontulloch:
yep - `Unrecognized field "subjectCurie" (class org.alliancegenome.curation_api.model.entities.GeneDiseaseAnnotation), not marked as ignorable`

oblodgett (approved):

PR #616, opened 2022-08-16
SCRUM-1759 - linkML model compatibility reporting

mluypaert:
This PR introduces java annotations on all java classes representing linkML curation schema classes, indicating the schema version of each to which it corresponds. This enables an updated version reporting, reporting the compatible schema version per class rather than claiming compatibility of the whole application to a single linkML curation schema version.


oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/model/entities/ontology/OntologyTerm.java
```diff
@@ -22,6 +23,7 @@
 @Entity
 @Inheritance(strategy = InheritanceType.JOINED)
 @ToString(exclude = { "isaParents", "isaChildren", "isaAncestors", "isaDescendants", "crossReferences", "synonyms", "secondaryIdentifiers", "subsets"}, callSuper = true)
+@AGRCurationSchemaVersion("1.2.1")
```
oblodgett: @mluypaert does this include 1.2.1 and all current pull requests? Because the "isaParents", "isaChildren", "isaAncestors", "isaDescendants" has not been released yet.

oblodgett: Even though we have implemented it.

mluypaert: Right, I missed that, these ancestry changes came in after merging in alpha into this branch so weren't represented in this schema version. I've now updated it to represent these new changes (now reported as `1.2.1+` like I did for other classes that require a new linkML model version which doesn't exist yet).

PR #617, opened 2022-08-16
SCRUM-1816 implement nonNullFields and nullFields on Column-level fil…

cmpich:
…tering

### DIFF
--- a/src/main/cliapp/src/components/GenericDataTable/FilterComponent.js
+++ b/src/main/cliapp/src/components/GenericDataTable/FilterComponent.js
@@ -43,8 +43,6 @@ export const FilterComponent = ({
 				<FilterComponentDropDown
 					isEnabled={isEnabled}
 					field={fields[0]}
-					nonNullFields={nonNullFields}
-					nullFields={nullFields}
 					filterName={filterName}
 					currentFilters={tableState.filters}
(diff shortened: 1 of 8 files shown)


src/main/cliapp/src/components/GenericDataTable/FilterComponent.js
```diff
@@ -39,6 +43,8 @@ export const FilterComponent = ({
 				<FilterComponentDropDown
 					isEnabled={isEnabled}
 					field={fields[0]}
+					nonNullFields={nonNullFields}
+					nullFields={nullFields}
```
oblodgett: @cmpich lets take these out as they are not implemented in the FilterComponentDropDown or FilterMultiSelectComponent

oblodgett (changes requested):

oblodgett (approved):

src/main/cliapp/src/components/GenericDataTable/FilterComponent.js
cmpich: Ok. Also removing from FilterMultiSelectComponent as well.

oblodgett: I pushed in the code already.

PR #618, opened 2022-08-16
SCRUM-1867 Show all condition relations

markquintontulloch:


oblodgett:
@markquintontulloch @cmpich @chris-grove could you guys fight this out somewhere?

chris-grove:
@oblodgett I think there was just a mixup/miscommunication. Please approve and merge this PR

oblodgett (approved):

cmpich:
The reason to have only non-null handle records was because ZFIN has only those records and needs to be able to create new handles. I am ok to include all records if there is a configuration / button that allows me to suppress the records to only display non-null handles and have this a customization in the profile.

PR #625, opened 2022-08-25
Release/v0.9.0 rc1

adamgibs:


oblodgett (approved):

mluypaert:
Hang on, the most recent commit in this PR is from 3 days ago, do we want to include commits that far into the current sprint in the beta release? @adamgibs @oblodgett

oblodgett:
@mluypaert good catch sorry I should not have approved this.

adamgibs:
Ah, okay. Should it go back to last Wednesday?

oblodgett:
@mluypaert upon second look at this. There is only 4 commits. 
1. Would have been in the sprint review.
2. Was only changes to the AGRSchema link annotation on java classes.
3. Was a README change
4. Was a single line fix to the UI to support disease annotations

So I don't think it would be bad to go from this point.

mluypaert:
> @mluypaert upon second look at this. There is only 4 commits.
> 
>     1. Would have been in the sprint review.
> 
>     2. Was only changes to the AGRSchema link annotation on java classes.
> 
>     3. Was a README change
> 
>     4. Was a single line fix to the UI to support disease annotations
> 
> 
> So I don't think it would be bad to go from this point.

Ok, if that's the only change made since sprint review then I agree, we can use this as the prerelease.

PR #629, opened 2022-08-29
SCRUM-1892 API support for DAA deletion

cmpich:


src/main/java/org/alliancegenome/curation_api/controllers/crud/DiseaseAnnotationCrudController.java
```diff
@@ -1,35 +1,43 @@
...
+		if (ret != null && ret.getTotalResults() == 1) {
+			return new ObjectResponse<>(ret.getResults().get(0));
 		} else {
-			return new ObjectResponse<DiseaseAnnotation>();
+			return new ObjectResponse<>();
 		}
 	}
+
+	@Override
+	public ObjectResponse<DiseaseAnnotation> delete(Long id) {
```
oblodgett: This method does not need to be overridden. This is what the base controller already does.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/controllers/crud/DiseaseAnnotationCrudController.java
cmpich: Done.

oblodgett (approved):

PR #635, opened 2022-09-06
SCRUM-1942 Validation and testing of empty strings in bulk submissions

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/NoteService.java
```diff
@@ -47,6 +48,17 @@ protected void init() {
 	@Transactional
 	public ObjectResponse<Note> update(Note uiEntity) {
 		Note dbEntity = noteValidator.validateNote(uiEntity, null, true);
+		if (dbEntity == null)
+			return null;
+		return new ObjectResponse<Note>(noteDAO.persist(dbEntity));
+	}
+	
+	@Override
```
oblodgett: it might be better to create a `upsert` method and have both create and update from the controller to call the upsert method, considering that the code is the same for both.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/services/NoteService.java
markquintontulloch: @oblodgett - done

abecerra (approved):

oblodgett (approved):

PR #636, opened 2022-09-07
SCRUM-1873, SCRUM-1878, SCRUM-1926, SCRUM-1965 - Update Java code to LinkML schema v1.2.4

markquintontulloch:


oblodgett:
So far the only issue I have with this... is it would have been better to split it up, and do this kind of a change in smaller chunks.

oblodgett:
PS. Generally it's not good to mix a refactor with functionality change.

oblodgett (approved):

markquintontulloch:
Fair point - noted

src/main/java/org/alliancegenome/curation_api/services/base/BaseOntologyTermService.java
```diff
@@ -201,30 +211,39 @@ private void handleCrossReferences(OntologyTerm dbTerm, OntologyTerm incomingTer
...
+		List<String> currentSynonymNames = currentSynonyms.stream().map(Synonym::getName).collect(Collectors.toList());
 		
-		Set<String> newSynonyms;
+		Set<Synonym> newSynonyms;
 		if(incomingTerm.getSynonyms() == null) {
 			newSynonyms = new HashSet<>();
 		} else {
 			newSynonyms = incomingTerm.getSynonyms().stream().collect(Collectors.toSet());
 		}
+		List<String> newSynonymNames = currentSynonyms.stream().map(Synonym::getName).collect(Collectors.toList());
```
oblodgett, line 229: I should have caught this:

`List<String> newSynonymNames = currentSynonyms.stream().map(Synonym::getName).collect(Collectors.toList());`

should be:

`List<String> newSynonymNames = newSynonyms.stream().map(Synonym::getName).collect(Collectors.toList());`

And it goes back to this comment:
"So far the only issue I have with this... is it would have been better to split it up, and do this kind of a change in smaller chunks."
Smaller changes are easier to catch things.

markquintontulloch: Indeed!

PR #640, opened 2022-09-09
SCRUM-2000 Fix display of Last Loaded date for bulk load files

markquintontulloch:


oblodgett:
I don't think this solves the true fix we need... we need to have a separate date for the "last run time" and leave these other dates alone... lets look at this later.

markquintontulloch:
@oblodgett - How about this?  Uses the last `loadFinished` date from the BulkLoadFileHistory for the "Last Loaded" date.  Avoids the need for creating and populating a new date field.

oblodgett:
Ontology loads don't get the history object (maybe they should for this reason)

markquintontulloch:
@oblodgett - 3rd time lucky?

src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkLoadFile.java
```diff
@@ -21,6 +30,11 @@
 @AGRCurationSchemaVersion("1.2.4")
 public class BulkLoadFile extends GeneratedAuditedObject {
 	
+	@FullTextField(analyzer = "autocompleteAnalyzer", searchAnalyzer = "autocompleteSearchAnalyzer", valueBridge = @ValueBridgeRef(type = OffsetDateTimeValueBridge.class))
+	@KeywordField(name = "dateLastLoaded_keyword", sortable = Sortable.YES, searchable = Searchable.YES, aggregable = Aggregable.YES, valueBridge = @ValueBridgeRef(type = OffsetDateTimeValueBridge.class))
+	@JsonView(View.FieldsOnly.class)
+	private OffsetDateTime dateLastLoaded;
```
oblodgett, line 36: @markquintontulloch we need to add this to LinkML?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkLoadFile.java
markquintontulloch: yeah

PR #642, opened 2022-09-12
SCRUM-1709

kthorat-prog:
Pushing branch now to check save api (as Literature references need a vpn connection on local system)


src/main/java/org/alliancegenome/curation_api/model/entities/GenomicEntity.java
```diff
@@ -15,6 +17,15 @@
 
 import lombok.*;
 
+@JsonTypeInfo(
```
oblodgett, line 20: @kthorat-prog I don't think we want this at both levels?

kthorat-prog: So shall I remove it from genomic Entity?

oblodgett: Was it solving an issue? I think it depends on what the subject field in the disease annotation is pointing to.

oblodgett: So Gene, Allele, and AGM are all sub classes of GenomicEntity so I would guess we would want it at that level.

kthorat-prog: Yes. I think so too.

kthorat-prog: I have removed type from BiologicalEntity class .Please check.

oblodgett (approved):

PR #651, opened 2022-09-14
Negated default value set & validation around subject field done

kthorat-prog:


src/main/cliapp/src/containers/diseaseAnnotationsPage/useNewAnnotaionReducer.js
```diff
@@ -1,7 +1,15 @@
 import { useImmerReducer } from "use-immer";
 
 const initialNewAnnotationState = {
-	newAnnotation: {},
+	newAnnotation: {
+		subject: {},
+		diseaseRelation: {},
+		singleReference : {},
+		object : {},
+		negated: null,
```
oblodgett: @kthorat-prog why not just default the negated to false here and then you don't have to do the check earlier?

kthorat-prog: ohh yes...correct...will make the changes.

oblodgett: Plus it would be better to not define the fields in the blank object. You can use the `?` operator to not have to worry about things being undefined:
`if(newAnnotation?.negated === false)` @adamgibs I can't remember how this works.

kthorat-prog: changes done as per comment. Please check.

src/main/cliapp/src/containers/diseaseAnnotationsPage/useNewAnnotaionReducer.js
```diff
@@ -1,7 +1,15 @@
...
-	newAnnotation: {},
+	newAnnotation: {
+		subject: {},
+		diseaseRelation: {},
+		singleReference : {},
+		object : {},
+		negated: false,
+		evidence : [],
+		with: []
+	},
```
oblodgett: @kthorat-prog do we need to have this object defined? vs just `newAnnotation: {},`

src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js
```diff
@@ -176,8 +178,13 @@ export const NewAnnotationForm = ({
 							<FormErrorMessageComponent errorMessages={errorMessages} errorField={"negated"}/>
 						</SplitterPanel>
 						<SplitterPanel style={{paddingRight: '10px'}}>
+<<<<<<< Updated upstream
 							<label htmlFor="object"><font color={'red'}>*</font>Disease</label>
 							<AutocompleteEditor
+=======
```
oblodgett: @kthorat-prog not sure how these got in there?

oblodgett (changes requested):

src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js
kthorat-prog: Changes done as per comments above.

oblodgett (approved):

PR #653, opened 2022-09-16
Changes around valuedisplay & other validations done

kthorat-prog:


markquintontulloch (changes requested):
I've created a PR into this branch that fixes the evidencesCodes/evidence inconsistency issue, plus addresses a few backend issues.

Once that is merged, the only outstanding issue that I can see is that when you click 'Save and Add Another' the populated fields disappear if there are any validation errors, forcing the user to re-complete the submission.  This doesn't appear to be a problem if you click 'Save and close'.

Update: There was also a bug where (on successful submission) the 'Save & Add Another' button just closed the form rather than clearing it.  I've implemented fixes for both these issues and have added them to my PR, so once that is approved and merged I think we're all good.

oblodgett (approved):

markquintontulloch (approved):

oblodgett:
Actually this need to wait until the delete bug is fixed... otherwise its getting deployed to production.

oblodgett (commented):
Actually this need to wait until the delete bug is fixed... otherwise its getting deployed to production.

markquintontulloch:
@oblodgett - that may be preferable to the currently broken functionality on alpha?

PR #654, opened 2022-09-16
Error code consistency and API/interface fixes

markquintontulloch:
This sorts out the inconsistency between 'evidence' and 'evidenceCodes'.  It also addresses the following points:

- We were getting a 500 error if required fields were null due to null pointer errors in the curie generation code
- The evidenceCode and with fields were not displaying on saving a new DA (but were being saved to the DB and appearing on refresh).  This was due to the POST interfaces for disease annotations not returning list fields.
- DateCreated was not being automatically populated when creating DAs (nor when creating anything else)

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/interfaces/crud/GeneDiseaseAnnotationCrudInterface.java
+++ b/src/main/java/org/alliancegenome/curation_api/interfaces/crud/GeneDiseaseAnnotationCrudInterface.java
@@ -27,12 +27,12 @@ public interface GeneDiseaseAnnotationCrudInterface extends BaseIdCrudInterface<
 	
 	@PUT
 	@Path("/")
-	@JsonView(View.DiseaseAnnotationUpdate.class)
+	@JsonView(View.DiseaseAnnotation.class)
 	public ObjectResponse<GeneDiseaseAnnotation> update(GeneDiseaseAnnotation entity);
 
 	@POST
 	@Path("/")
-	@JsonView(View.DiseaseAnnotationCreate.class)
+	@JsonView(View.DiseaseAnnotation.class)
 	public ObjectResponse<GeneDiseaseAnnotation> create(GeneDiseaseAnnotation entity);
 	
 	@POST
--- a/src/main/java/org/alliancegenome/curation_api/interfaces/crud/GeneDiseaseAnnotationCrudInterface.java
+++ b/src/main/java/org/alliancegenome/curation_api/interfaces/crud/GeneDiseaseAnnotationCrudInterface.java
@@ -32,7 +32,7 @@ public interface GeneDiseaseAnnotationCrudInterface extends BaseIdCrudInterface<
 
 	@POST
 	@Path("/")
-	@JsonView(View.DiseaseAnnotationUpdate.class)
+	@JsonView(View.DiseaseAnnotationCreate.class)
 	public ObjectResponse<GeneDiseaseAnnotation> create(GeneDiseaseAnnotation entity);
 	
 	@POST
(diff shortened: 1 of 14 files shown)


src/main/java/org/alliancegenome/curation_api/interfaces/crud/GeneDiseaseAnnotationCrudInterface.java
```diff
@@ -29,6 +29,11 @@ public interface GeneDiseaseAnnotationCrudInterface extends BaseIdCrudInterface<
 	@Path("/")
 	@JsonView(View.DiseaseAnnotationUpdate.class)
 	public ObjectResponse<GeneDiseaseAnnotation> update(GeneDiseaseAnnotation entity);
+
+	@POST
+	@Path("/")
+	@JsonView(View.DiseaseAnnotationUpdate.class)
```
oblodgett: If we could call this DiseaseAnnotationCreate then if we need an update view these would not conflict

markquintontulloch: @oblodgett @kthorat-prog Separate DiseaseAnnotationCreate view now created.  I've only added this to the `evidenceCodes` and `with` fields in the DA entity class for now, so will have to add to others once we implement more list fields in the UI creation

src/main/java/org/alliancegenome/curation_api/view/View.java
```diff
@@ -17,5 +17,6 @@ public static class BulkLoadFileHistory extends FieldsOnly { }
 	public static class ReportHistory extends FieldsOnly { }
 	
 	public static class DiseaseAnnotationUpdate extends FieldsOnly { }
+	public static class DiseaseAnnotationCreate extends FieldsOnly { }
```
oblodgett: @markquintontulloch you might not know this so I am not trying to be picky but one of the things you can do is have a interum class like the following:

```
public static class DiseaseAnnotation extends FieldsOnly { }
public static class DiseaseAnnotationUpdate extends DiseaseAnnotation { }
public static class DiseaseAnnotationCreate extends DiseaseAnnotation { }
```
Then in your interfaces you can do:
`@JsonView({View.FieldsAndLists.class, View.DiseaseAnnotation.class})`

and it will include both views... or if you need the specific view you could choose it by itself, this just helps with finer grain control of the resulting object.

markquintontulloch: @oblodgett - good to know.  What happens in the entity class?  Presumably that stays as `@JsonView({View.FieldsAndLists.class, View.DiseaseAnnotationUpdate.class, View.DiseaseAnnotationCreate.class})`?

markquintontulloch: or in cases where you want it to apply to both update and create views I guess it can be `DiseaseAnnotation.class` in the entity too?

oblodgett: Ya... whenever you want to use both you would use the: `View.DiseaseAnnotation.class`

oblodgett: I mean the classes are basically the same but to take this idea further (i'm not suggesting that) you could have separate views for each object View.GeneAnnotationCreate.class, Update, etc, etc. And those could all end up extending `View.DiseaseAnnotation.class` then on the endpoint you would just need the DA view and all those classes would kick in when needed or if you have a very specific endpoint you could just use the Specific view. We do a lot move with view in the other API.

https://github.com/alliance-genome/agr_java_software/blob/master/agr_java_core/src/main/java/org/alliancegenome/neo4j/view/View.java

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/interfaces/crud/GeneDiseaseAnnotationCrudInterface.java
oblodgett: @markquintontulloch does this still need to be changed to "Create"?

oblodgett (approved):

PR #658, opened 2022-09-21
SCRUM-1709 Don't clear form on clicking 'Save & Add Another'

markquintontulloch:


src/main/cliapp/src/containers/diseaseAnnotationsPage/useNewAnnotaionReducer.js
```diff
@@ -1,7 +1,7 @@
 import { useImmerReducer } from "use-immer";
 
 const initialNewAnnotationState = {
-	newAnnotation: {},
+	newAnnotation: {"negated": false},
```
oblodgett, line 4: Does this field need to be specified here?

markquintontulloch: I think so, otherwise it gets cleared when you do CLEAR

oblodgett: We were trying to not set any fields in the object, and just handle any issues elsewhere.

oblodgett: You could just default it in the clear call.

markquintontulloch: I tried just setting it in the 'CLEAR' case (as it was previously done in the 'OPEN_DIALOG' case) but it was throwing errors the second time you hit clear: 
`Uncaught TypeError: Cannot assign to read only property 'negated' of object '#<Object>'`

oblodgett: So doing a `draft.newAnnotation["negated"] = false;` in the CLEAR was throwing that error?

markquintontulloch: yep, but only on the 2nd time you hit Clear

markquintontulloch: and only if you'd entered something in one of the non-list fields

oblodgett (approved):

markquintontulloch:
Redundant due to code incorporated in SCRUM-1903

PR #659, opened 2022-09-21
Merge pull request #657 from alliance-genome/release/v0.10.0-rc1

abecerra:
Deploying beta v0.10.0-rc1 to beta


mluypaert:
There's essentially no need for this PR, as this is a deployment from alpha to beta, hence there are no new change to merge back into alpha (as can be seen in the files changed section). PRmerge branches are only relevant for deployments from beta to production.

mluypaert:
My bad, forgot the fact that the mergeback of the PR merge commit is required for correct tag reporting on alpha 🤦‍♂️ 
That being said, I think the `v0.10.0-rc1` tag should be on commit fb771c34 (the last commit on beta, which is the merge commit after merging in alpha) rather than on 5ac2fb1a9 (the commit before that, currently the last commit on alpha).

mluypaert (approved):

PR #661, opened 2022-09-23
SCRUM-2037 Update experimental condition Unique ID generation

markquintontulloch:
Removes the use of conditionStatement from the ExperimentalCondition and ConditionRelation uniqueIds and adds the conditionFreeText field.  Database migration deletes all ExperimentalCondition and ConditionRelation entries in order that they are reloaded with new uniqueIds when DiseaseAnnotation loads are rerun.

### DIFF
--- a/src/main/resources/db/migration/v0.10.0.1__agr_curation_api.sql
+++ b/src/main/resources/db/migration/v0.10.0.1__agr_curation_api.sql
@@ -1,0 +1,9 @@
+DELETE FROM diseaseannotation_conditionrelation;
+
+DELETE FROM conditionrelation_experimentalcondition;
+
+DELETE FROM experimentalcondition;
+
+DELETE FROM conditionrelation;
+		
+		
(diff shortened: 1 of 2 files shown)


src/main/resources/db/migration/v0.10.0.1__agr_curation_api.sql
```diff
@@ -0,0 +1,9 @@
+DELETE FROM diseaseannotation_conditionrelation;
+
+DELETE FROM conditionrelation_experimentalcondition;
+
+DELETE FROM experimentalcondition;
+
+DELETE FROM conditionrelation;
+		
+
```
mluypaert, line 9: I believe this migration would effectively strip disease annotations that have conditionrelations from their conditionrelations, and leave the disease annotation? Would a disease annotation reload afterwards then repopulate those diseaseannotation-conditionrelation links, or would this migration need to be edited to remove the disease annotations with conditionrelations? Also, would there be a less brute-force migration possible (like setting the values of only the experimentalcondition uniqueId to null rather than removing all experimental condition records)?

markquintontulloch: Yes, that's exactly what this migration is intended to do.  DA reload will reattach all condition relations and experimental conditions.   A less brute-force method would not achieve the cleanup of duplicate experimental conditions required.

mluypaert (approved):

mluypaert (commented):
@markquintontulloch I assume you still want to wait with merging this PR until after the data rolldown to alpha? (Mainly as a note to others so that this PR does not accidentally get merged anyway)

markquintontulloch:
Yep, might as well leave it until ready to do the reload to avoid there being a prolonged period of no experimental conditions or condition relations on alpha.

PR #671, opened 2022-09-28
Library setup

oblodgett:


pom.xml
```diff
@@ -33,11 +33,30 @@
...
+		</license>
+	</licenses>
+
+	<developers>
+		<developer>
+			<name>Olin Blodgett</name>
+			<organization>Jackson Laboratory</organization>
+			<organizationUrl>http://www.jax.org</organizationUrl>
+		</developer>
+	</developers>
```
mluypaert: You may wanna use `Alliance of Genome Resources` here instead of `Jackson Laboratory` as the organization (and use the appropriate URL as well). But do we need to mention individual developers here? That could be a long list if so...

oblodgett: Well... I was thinking the list of current devs... and this could get modified. I would agree that would make sense but each dev is from a different organization so... I don't know.

mluypaert: I'd argue that in the context of this code, the fact that we are all employed by different organisations doesn't matter, we're all working for the Alliance and that is what should be reflected here.

oblodgett: I disagree... if AGR was a normal organization I totally agree, but it doesn't exist as an entity. And I think of this section as more of a contact section... Also we might be able to just leave it out all together, which would be my vote. I need to see if maven requires the developer section.

mluypaert: Ya, if maven doesn't require it then I agree, better to just drop this section alltogether.

oblodgett: `Invalid POM: /org/alliancegenome/agr_chipmunk/0.5.0-rc5/agr_chipmunk-0.5.0-rc5.pom: Developer information missing`
Guess that answers that.

mluypaert (approved):

PR #676, opened 2022-09-30
SCRUM-2011: adds service to mint identifiers calling MaTI

abecerra:
1. This would still need work on github actions ENV vars
2. I think the usage (duplication) of 2 classes here is easier than including MaTI as a library in agr_curation
3. It can be tested calling integration tests

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/model/entities/Identifier.java
+++ b/src/main/java/org/alliancegenome/curation_api/model/entities/Identifier.java
@@ -1,0 +1,13 @@
+package org.alliancegenome.curation_api.model.entities;
+
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class Identifier {
+	private Long counter;
+	private String subdomain_code;
+	private String subdomain_name;
+	private String curie;
+}
--- a/src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
@@ -1,0 +1,82 @@
+package org.alliancegenome.curation_api.services.mati;
+
+import io.restassured.RestAssured;
+import io.restassured.http.ContentType;
+import lombok.NoArgsConstructor;
+import org.alliancegenome.curation_api.model.entities.IdentifiersRange;
+import org.eclipse.microprofile.config.inject.ConfigProperty;
+
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+import java.io.IOException;
+import java.util.Base64;
+
+@NoArgsConstructor
+@ApplicationScoped
+public class MaTIService {
+
+	@Inject
+	@ConfigProperty(name = "okta.client.id")
+	String client_id;
+	@Inject
+	@ConfigProperty(name = "okta.client.secret")
+	String client_secret;
+	@Inject
+	@ConfigProperty(name = "okta.url")
+	String okta_url;
+	@Inject
+	@ConfigProperty(name = "okta.scopes")
+	String okta_scopes;
+	@Inject
+	@ConfigProperty(name = "mati.url")
+	String mati_url;
+
+
+	private String fetchOktaToken() throws IOException {
+		String authorization = "Basic " +
+			Base64.getEncoder().encodeToString((client_id+":"+client_secret).getBytes());
+		String token = RestAssured.given().
+			contentType(ContentType.URLENC).
+			header("Accept", "application/json").
+			header("Cache-Control", "no-cache").
+			header("Authorization", authorization).
+			formParam("grant_type", "client_credentials").
+			formParam("scope", okta_scopes).
+			when().
+			post(okta_url + "/oauth2/default/v1/token").
+			then().
+			extract().path("access_token").toString();
+		return token;
+	}
+
+	public String mintIdentifier(String subdomain)  throws IOException {
+		String token = fetchOktaToken();
+		String authorization = "Bearer: " + token;
+		String identifier = RestAssured.given().
+			contentType(ContentType.JSON).
+			header("Accept", "application/json").
+			header("Authorization", authorization).
+			header("subdomain", subdomain).
+			when().
+			put(mati_url).
+			then().
+			extract().path("value").toString();
+		return identifier;
+	}
+
+	public IdentifiersRange mintIdentifierRange(String subdomain, String howMany)  throws IOException {
+		String token = fetchOktaToken();
+		String authorization = "Bearer: " + token;
+		IdentifiersRange range = RestAssured.given().
+			contentType(ContentType.JSON).
+			header("Accept", "application/json").
+			header("Authorization", authorization).
+			header("subdomain", subdomain).
+			header("value", howMany).
+			when().
+			post(mati_url).
+			then().
+			extract().body().as(IdentifiersRange.class);
+		return range;
+	}
+}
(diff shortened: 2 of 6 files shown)


src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
```diff
@@ -0,0 +1,79 @@
...
+	private String okta_scopes;
+//	@ConfigProperty(name = "mati.url")
+	private String mati_url;
+
+	public MaTIService() {
+		client_id = System.getenv("OKTA_CLIENT_ID");
+		client_secret = System.getenv("OKTA_CLIENT_SECRET");
+		okta_url = System.getenv("OKTA_URL");
+		okta_scopes = System.getenv("OKTA_SCOPES");
+		mati_url = System.getenv("MATI_URL");
```
oblodgett: These should be pulled via the @ConfigProperty and some default values can be added to the application.properties.defaults file. See the Bulk services for examples using a `Instance<String>`  doesn't throw the error when the variables are not defined. You can see an example in the file: https://github.com/alliance-genome/agr_curation/blob/alpha/src/main/java/org/alliancegenome/curation_api/auth/AuthenticationFilter.java

src/main/java/org/alliancegenome/curation_api/model/entities/Identifier.java
```diff
@@ -0,0 +1,13 @@
...
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@Data
+@AllArgsConstructor
+public class Identifier {
+	private Long counter;
+	private String subdomain_code;
+	private String subdomain_name;
+	private String curie;
```
oblodgett, line 12: Using the library will prevent us from having duplicate code between the two repos.

src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
```diff
@@ -0,0 +1,79 @@
...
+			header("Accept", "application/json").
+			header("Cache-Control", "no-cache").
+			header("Authorization", authorization).
+			formParam("grant_type", "client_credentials").
+			formParam("scope", okta_scopes).
+			when().
+			post(okta_url + "/oauth2/default/v1/token").
+			then().
+			extract().path("access_token").toString();
+		return token;
```
oblodgett, line 49: And using the RestProxy via the library will prevent all the hardcoding of these headers and credentials... etc.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
abecerra: From a class with  @ConfigProperty, the creation of an object in an integration test yields NULL. It does not matter if the attributes are typed directly or with Instance. 

The use of ENV vars is the alternative I found, is there another?

src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
abecerra: if by RestProxy you are referring to  rest-client-reactive. The problem could be the same: in an integration test an object created from an interface annotated with @RegisterRestClient would have to be Injected, producing NULL, I guess.

src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
oblodgett: with integration tests, those values just need to be set otherwise yes they will be NULL... and we can make sure and have them set when running.

src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
oblodgett: sorry what I mean by rest proxy is:
```java
IdentifierResourceRESTInterface mati_api = RestProxyFactory.createProxy(IdentifierResourceRESTInterface.class, mati_base_url);
Response resp = mati_api.increment(authorization, subdomain, value);
```
Plus the integration tests are not going to work via github actions because MATI is behind a firewall. So I would think so long as they might work locally, then we can disable them when pushing the code.

This is also much less code than using RestAssured.

abecerra: The tests run locally with ENV vars

oblodgett: They will also run via @ConfigProperty we do that for all our other tests.

src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
abecerra: I created MATI_URL and OKTA_SCOPES in aws eb environment for alpha, beta and prod

src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
abecerra: The argument authorization would need a token. So, the use of RestProxyFactory would still need to add more code.

src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
oblodgett: Here is the documentation on how @ConfigProperty works.

https://quarkus.io/guides/config-reference#configuration-sources

As you can see there is a chain of configuration options which System.getenv is included. So it's better to use the convention. If the values are not coming through its not the @ConfigProperty that is the issue, it's that the configuration variables are not being set correctly.

oblodgett: On another note, MATI_URL is not going to be available to github action. Mati is behind the AWS firewall. @mluypaert this is another reason I wanted to make the Mati API public, it is protected via okta auth, and should be available for this kind of thing.

abecerra: Making it public also eases development and testing for Blue Team

abecerra: Last commit cleans up the comments and the integration test runs with all the others.

src/main/java/org/alliancegenome/curation_api/services/mati/MaTIService.java
```diff
@@ -10,24 +10,11 @@
...
-		client_secret = System.getenv("OKTA_CLIENT_SECRET");
-		okta_url = System.getenv("OKTA_URL");
-		okta_scopes = System.getenv("OKTA_SCOPES");
-		mati_url = System.getenv("MATI_URL") + "api/identifier";
-	}
+	private String client_id = System.getenv("OKTA_CLIENT_ID");
+	private  String client_secret= System.getenv("OKTA_CLIENT_SECRET");
+	private String okta_url = System.getenv("OKTA_URL");
+	private String okta_scopes = System.getenv("OKTA_SCOPES");
+	private String mati_url = System.getenv("MATI_URL") + "api/identifier";
```
oblodgett: These variables are already defined in the curation software as the following:
```java
	@ConfigProperty(name = "okta.authentication")
	Instance<Boolean> okta_auth;

	@ConfigProperty(name = "okta.url")
	Instance<String> okta_url;

	@ConfigProperty(name = "okta.client.id")
	Instance<String> client_id;

	@ConfigProperty(name = "okta.client.secret")
	Instance<String> client_secret;

	@ConfigProperty(name = "okta.api.token")
	Instance<String> api_token;
	```

abecerra: Added a temporary controller to be deleted after testing and after using the service to mint DA IDs

oblodgett (approved):

PR #677, opened 2022-10-03
SCRUM-2119 LinkML schema version reporting

markquintontulloch:
- Report range of versions each class is compatible with
- Only display submitted classes in Data Loads page
- Add schema version to bulkLoadFile
- Disable reload of files outside of supported schema versions


src/main/java/org/alliancegenome/curation_api/services/APIVersionInfoService.java
```diff
@@ -0,0 +1,122 @@
...
+@RequestScoped
+public class APIVersionInfoService {
+	
+	public Boolean isPartiallyImplemented(Class<?> clazz, AGRCurationSchemaVersion version) {
+		if (version.partial()) return true;
+		
+		Set<String> dependencyClasses = getDependencyClasses(version);
+		for (String className : dependencyClasses) {
+			try {
+				String packageName = "org.alliancegenome.curation_api.model.entities.";
```
oblodgett: Instead of hardcoding these to a package name, which makes this a hidden dependency to refactoring package names, which will break this code unknowingly. Can we change the dependencies as a list of classes vs list of strings in the java annotation? As in:
```
@AGRCurationSchemaVersion("1.2.4", dependencies =[Synonym.class, AGM.class, CrossReference.class])
```

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/services/APIVersionInfoService.java
markquintontulloch: Done - didn't realise you could have an array of Classes in these annotations, but it seems you can and this makes much more sense

src/main/java/org/alliancegenome/curation_api/interfaces/AGRCurationSchemaVersion.java
```diff
@@ -1,10 +1,14 @@
 package org.alliancegenome.curation_api.interfaces;
 
-import java.lang.annotation.*;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
 
 @Retention(RetentionPolicy.RUNTIME)
 public @interface AGRCurationSchemaVersion {
 
-	String value();
```
oblodgett, line 8: The only draw back to not having value is you can't specify the `AGRCurationSchemaVersion("default value");` it might be nice to have "default value" set both min and max.

src/main/java/org/alliancegenome/curation_api/model/ingest/dto/IngestDTO.java
```diff
@@ -12,6 +12,10 @@
 @Getter
 public class IngestDTO {
 	
+	@JsonView({View.FieldsOnly.class})
+	@JsonProperty("linkml_version")
```
oblodgett, line 16: I think this is going to cause push back... but I guess we will see.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/interfaces/AGRCurationSchemaVersion.java
markquintontulloch: True, but we could just specify default min and default max.

oblodgett: But it might be nice to do it on a class by class basis.

src/main/java/org/alliancegenome/curation_api/model/ingest/dto/IngestDTO.java
markquintontulloch: I'm hoping not.  The DQMs are used to submitting metadata, and personally I find it useful to be easily identify the schema version submitted to by looking at the file.

PR #683, opened 2022-10-06
SCRUM-2130 Load and display of additional Allele attributes (iteration 1)

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/dao/GenomicEntityDAO.java
```diff
@@ -0,0 +1,33 @@
...
+	
+	protected GenomicEntityDAO() {
+		super(GenomicEntity.class);
+	}
+
+	public void deleteAttachedSynonym(Long id) {
+		Query jpqlQuery = entityManager.createNativeQuery("DELETE FROM genomicentity_synonym WHERE synonyms_id = '" + id + "'");
+		jpqlQuery.executeUpdate();
+		
+		synonymDAO.remove(id);
```
oblodgett: Its better if we don't use hardcoded queries see my later comment.

src/main/java/org/alliancegenome/curation_api/services/helpers/validators/GenomicEntityValidator.java
```diff
@@ -37,4 +54,85 @@ public String validateName(GenomicEntity uiEntity) {
...
+		if (CollectionUtils.isNotEmpty(validatedSynonyms))
+			validatedIds = validatedSynonyms.stream().map(Synonym::getId).collect(Collectors.toList());
+		for (Synonym validatedSynonym : validatedSynonyms) {
+			if (!previousIds.contains(validatedSynonym.getId()))
+				synonymDAO.persist(validatedSynonym);
+		}
+		
+		List<Long> idsToRemove = ListUtils.subtract(previousIds, validatedIds);
+		for (Long id : idsToRemove) {
+			genomicEntityDAO.deleteAttachedSynonym(id);
```
oblodgett: Rather than running the hardcoded query you could do something like this:
```java
Predicate<Synonym> removeCondition = synonym -> idsToRemove.contains(synonym.getId());
dbEntity.getSynonyms().removeIf(removeCondition);
```

oblodgett: And you may need a DELETE CASCADE on the GE.synonyms field so that you don't have to delete synonyms directly.

markquintontulloch: @oblodgett - good tip, done

src/main/java/org/alliancegenome/curation_api/dao/GenomicEntityDAO.java
markquintontulloch: now removed

oblodgett (approved):

PR #691, opened 2022-10-07
SCRUM-2130 Add value bridge that doesn't index null as false

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/bridges/BooleanAndNullValueBridge.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.hibernate.search.mapper.pojo.bridge.ValueBridge;
+import org.hibernate.search.mapper.pojo.bridge.runtime.ValueBridgeToIndexedValueContext;
+
+public class BooleanAndNullValueBridge implements ValueBridge<Boolean, String> {
+	@Override
+	public String toIndexedValue(Boolean value, ValueBridgeToIndexedValueContext context) {
+		if(value == null)
+			return null;
+		return value ? "true" : "false";
```
oblodgett, line 11: Good ole three values for a boolean.

oblodgett (approved):

PR #697, opened 2022-10-11
SCRUM-2098

adamgibs:


src/main/cliapp/src/service/useSessionStorage.js
```diff
@@ -1,9 +1,10 @@
 import { useState, useEffect } from "react";
+import { LoggedInPersonService } from './LoggedInPersonService';
```
oblodgett: @adamgibs can we get rid of useSessionStorage? Or do you think we will use it in the future... I would rather that we incorporate it into the useGetUserSettings... that way it would be much faster to pull the value from localStorage vs the DB.

src/main/cliapp/src/components/GenericDataTable/useGenericDataTable.js
```diff
@@ -299,15 +300,15 @@ export const useGenericDataTable = ({
...
+
 	const resetTableState = () => {
 		let _tableState = {
 			...initialTableState,
 			isFirst: false,
 		};
 
 		setTableState(_tableState);
-		setDefaultColumnOrder(columns, dataTable, defaultColumnNames, deletionEnabled);
+		setDefaultColumnOrder(columns, dataTable, defaultColumnNames, deletionEnabled, isLoading);
```
oblodgett: @adamgibs do we want to remove this isLoading variable?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/model/entities/PersonSetting.java
```diff
@@ -0,0 +1,50 @@
...
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@Audited
+@Entity
+@Data
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
+@ToString(callSuper = true, exclude = "person")
+@JsonIgnoreProperties(ignoreUnknown = true)
+@AGRCurationSchemaVersion(min="1.3.2", max=LinkMLSchemaConstants.LATEST_RELEASE, dependencies={Agent.class})
```
markquintontulloch: Should be `dependencies={AuditedObject.class}`

oblodgett: Ok will change.

PR #709, opened 2022-10-17
SCRUM-2139 Return exact matches only from dropdown filters

markquintontulloch:
The useKeywordFields option was not being set as true for  a number of the dropdown columns.  This fixed all columns except for 'In Collection' in the Allele table - the vocabulary terms for this field had spaces which prevent this from working, so this PR replaces all those terms with versions with underscores instead of spaces.


oblodgett:
@markquintontulloch do the key word fields not match exact with spaces? This probably is something we need to fix on the backend vs making the names have underscores? Granted this may be a "quick" fix but this is going to be an issue going forward.

oblodgett (approved):

markquintontulloch:
@oblodgett  - no doesn't work with spaces.  I think the logic splits the querystring on spaces so that you can filter for exact matches on multiple terms.  There might be a better way around this, but we could just enforce that vocabulary term names must not have spaces (doesn't seem too unreasonable).

oblodgett:
@markquintontulloch we are using a simplequery on the ES side... this is causing this issue. If we were using a "match" query on the ES side, we wouldn't have this issue, we would have other issues where we would not partial match, just saying we need to be able to switch between the two based on the UI needs.... so maybe we need a queryType: simple or match then we don't have to change our data in order to get correct results.

PR #713, opened 2022-10-17
Added indexes to allele touching tables

oblodgett:


markquintontulloch (approved):

markquintontulloch (changes requested):
@oblodgett - shouldn't the migration version be 0.12.0.1 (or even 0.11.0.8 given we haven't deployed alpha to beta yet)

markquintontulloch (approved):

PR #715, opened 2022-10-18
SCRUM-2011: temporal change to debug curation - mati interaction

abecerra:


oblodgett:
This should be testable locally?

abecerra:
> This should be testable locally?

 I need to read the response in alpha. Locally everything works fine.
After testing, this commit can be rolled back.

mluypaert (approved):

PR #720, opened 2022-10-18
SCRUM-2161 Make allele attributes editable

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/interfaces/crud/AlleleCrudInterface.java
```diff
@@ -46,4 +46,16 @@ public void reindex(
...
+	@Override
+	@PUT
+	@Path("/")
+	@JsonView(View.FieldsAndLists.class)
+	public ObjectResponse<Allele> update(Allele entity);
+	
+	@Override
+	@POST
+	@Path("/")
+	@JsonView(View.FieldsAndLists.class)
```
oblodgett: Are these going to cause issues with circular references?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/interfaces/crud/AlleleCrudInterface.java
markquintontulloch: @oblodgett Shouldn't do at this point, but have added allele views for future-proofing

PR #727, opened 2022-10-20
Add LinkML version submission documentation

chris-grove:
Added documentation for including the LinkML version in the JSON file header to the Submitting Data section

### DIFF
--- a/README.md
+++ b/README.md
@@ -585,6 +585,16 @@ This is a grouping mechanism to group files together
 | WB    | Worm Base |
 | ZFIN  | Zebrafish Information Network |
 
+### Including corresponding LinkML version in the JSON file submission header
+
+The LinkML version for which the file is being submitted now needs to be added to the JSON file header, for example:
+
+```json
+{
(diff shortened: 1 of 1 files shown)


markquintontulloch (approved):

README.md
```diff
@@ -585,6 +585,16 @@ This is a grouping mechanism to group files together
 | WB    | Worm Base |
 | ZFIN  | Zebrafish Information Network |
 
+### Including corresponding LinkML version in the JSON file submission header
+
+The LinkML version for which the file is being submitted now needs to be added to the JSON file header, for example:
+
+```json
+{
+  "linkml_version" : "v1.3.2",
```
oblodgett, line 594: @markquintontulloch just wondering that if we could maybe look for this but if it's null then just assume the current version? Or at least handle the exception better when it is missing.

markquintontulloch: yeah, Christian mentioned that - I'm in two minds about it.  It think just sticking the current version on there may cause more issues than it solves in the long term.  Better error handling is definitely a good idea though.

PR #730, opened 2022-10-21
SCRUM-2161 fix reference autosuggest

markquintontulloch:
Displays AGR curies for populated references in edit mode rather than PMID/other curies, but @chris-grove has approved that as a short-term solution


adamgibs (approved):
@markquintontulloch This will be different once I push my code out, but I'll make sure to preserve the logic in some way.

PR #736, opened 2022-10-25
Scrum 1917 experiments changes

adamgibs:


oblodgett (approved):

src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js
```diff
@@ -376,18 +368,30 @@ export const NewAnnotationForm = ({
...
 								name="experiments"
+								customRef={experimentsRef}
 								editorChange={onDropdownExperimentsFieldChange}
 								referenceCurie={newAnnotation.singleReference.curie}
 								value={newAnnotation.conditionRelations[0]?.handle}
 								showClear={false}
 								placeholderText={newAnnotation.conditionRelations[0]?.handle}
-								isEnabled={isExperimentsEnabled}
+								isEnabled={
+									(
```
oblodgett: @adamgibs wouldn't this be better to call a method vs inline code?

adamgibs: @oblodgett Sure, I just updated it with the logic in functions

PR #744, opened 2022-10-27
SCRUM-2223 Fix discrepancy in publication curies used for DA and CR uniqueId fields

markquintontulloch:
Uses AGR reference curie for generating uniqueIds via bulk uploads.

Reload of disease annotations required as DB migration removes all condition relations.

### DIFF
--- a/src/main/resources/db/migration/v0.12.0.4__agr_curation_api.sql
+++ b/src/main/resources/db/migration/v0.12.0.4__agr_curation_api.sql
@@ -1,0 +1,5 @@
+DELETE FROM diseaseannotation_conditionrelation;
+
+DELETE FROM conditionrelation_experimentalcondition;
+		
+DELETE FROM conditionrelation;
--- a/src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/DiseaseAnnotationCurie.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/DiseaseAnnotationCurie.java
@@ -26,19 +26,19 @@ public abstract class DiseaseAnnotationCurie {
 		return curie.getCurie();
 	}
 	
-	public static String getConditionRelationUnique(ConditionRelationDTO dto) {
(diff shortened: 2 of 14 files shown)


src/main/resources/db/migration/v0.12.0.4__agr_curation_api.sql
```diff
@@ -0,0 +1,5 @@
+DELETE FROM diseaseannotation_conditionrelation;
+
+DELETE FROM conditionrelation_experimentalcondition;
+		
+DELETE FROM conditionrelation;
```
oblodgett, line 5: @markquintontulloch does this make sense in light of previous migrations?

markquintontulloch: @oblodgett Not sure which previous migrations you mean, but essentially we need to clean up condition relations that have incorrect paper curies in them.  We could just delete any that have 'PMID' or 'AGR-Reference' in the uniqueId but that wouldn't clean up any that were generated in the period between when the LiteratureService started using AGRKB curies, and when we switched to their production site.  That's possibly only an issue on alpha and I think we could clean those up by checking the conditionRelation.singleReference to see if it's obsolete (from previous migration), but given we have to reload all disease annotations anyway (to fix the DA uniqueId) I figured we might as well delete them all.

oblodgett: Right, but taking a bundle of migrations and putting them on beta will everything be ok with the deployment?

oblodgett: Maybe everything will be fine.

oblodgett (approved):

src/main/resources/db/migration/v0.12.0.4__agr_curation_api.sql
markquintontulloch: I don't see why not.  For v0.12 we've got 4 migrations - one to create database indexes, the next to obsolete references, one to cleanup detached notes, and then this one.  Shouldn't be any clashes.

PR #746, opened 2022-10-28
SCRUM-2184 Backend implementation of Vocabulary Term Sets

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/interfaces/crud/VocabularyTermSetCrudInterface.java
```diff
@@ -0,0 +1,41 @@
...
+	
+	@GET
+	@Path("/{id}/terms")
+	@JsonView(View.VocabularyTermSetView.class)
+	ObjectListResponse<VocabularyTerm> getTerms(@PathParam("id") Long id);
+	
+	@GET
+	@Path("/findBy/{name}")
+	@JsonView(View.VocabularyTermSetView.class)
+	public ObjectResponse<VocabularyTermSet> findByName(@PathParam("name") String name);
```
oblodgett: Shouldn't need this endpoint? Can use the `/find` endpoint with a { name: "name to lookup" }

markquintontulloch: I don't see any /find endpoint in swagger

oblodgett: Its under the database search endpoints

markquintontulloch: Ah, of course - we do the same thing for other things like Vocabulary though, and personally I like the convenience.  Can remove if you feel strongly otherwise though.

oblodgett: Ya it would be good to use the generic one if possible. I think we need to more the endpoints under the name section for each type.

src/main/java/org/alliancegenome/curation_api/view/View.java
```diff
@@ -7,9 +7,10 @@ public static class FieldsOnly { }
 	public static class FieldsAndLists extends FieldsOnly { }
 	public static class ConditionRelationUpdateView extends FieldsOnly { }
 
-	public static class VocabularyTermView extends FieldsAndLists { }
+	public static class VocabularyTermView extends FieldsOnly { }
```
oblodgett: Would be curious all the places that use the VocabularyTermView and make sure this wouldn't cause any issues.

markquintontulloch: Don't think I intended to switch that - will switch it back

src/main/java/org/alliancegenome/curation_api/interfaces/crud/VocabularyTermSetCrudInterface.java
markquintontulloch: @oblodgett - has now been removed

oblodgett (approved):

PR #747, opened 2022-10-28
Add vocabulary name index

markquintontulloch:


src/main/resources/db/migration/v0.12.0.6__agr_curation_api.sql
```diff
@@ -0,0 +1 @@
+CREATE INDEX vocabularyterm_name_index ON public.vocabularyterm USING btree (name);
```
oblodgett: @markquintontulloch need the code on the class also.

markquintontulloch: Done, I think

oblodgett: Do we need this on the vocabulary.name also?

oblodgett: I would guess we are going to need this on the Vocab Set name also?

markquintontulloch: There's only 13 vocabularies at this point, but there will be more soon.  What do you think?

markquintontulloch: What the hell - I've stuck it on for good measure!

oblodgett: If we are doing the join from the terms to the vocabularies... with no indexes then its taking (terms * vocabularies) and having to do a full scan. So I would suggest we put it on both so that it can use both indexes on the look up.

oblodgett (approved):

src/main/resources/db/migration/v0.12.0.6__agr_curation_api.sql
markquintontulloch: I haven't added an index for the VocabularyTermSet as currently there's only one of those - but something to bear in mind for the future

PR #751, opened 2022-11-01
Added missing schemas and extra indexes

oblodgett:
On my machine this took the Allele load from 2r/s to 20r/s not great but better.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/model/entities/Allele.java
+++ b/src/main/java/org/alliancegenome/curation_api/model/entities/Allele.java
@@ -54,7 +54,10 @@ public class Allele extends GenomicEntity {
 	@IndexedEmbedded(includeDepth = 2)
 	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
 	@ManyToMany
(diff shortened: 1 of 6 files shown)


src/main/java/org/alliancegenome/curation_api/model/entities/Allele.java
```diff
@@ -54,7 +54,10 @@ public class Allele extends GenomicEntity {
 	@IndexedEmbedded(includeDepth = 2)
 	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
 	@ManyToMany
-	@JoinTable(indexes = @Index( columnList = "Allele_curie, references_curie"))
+	@JoinTable(indexes = {
+		@Index( columnList = "Allele_curie"),
+		@Index( columnList = "references_curie")
+	})
```
oblodgett, line 60: @markquintontulloch composite index vs single indexes...
When a query does a `index_field1=value1 and index_field2=value2` then the composite will be used, otherwise it won't, and these queries were not generating this.

markquintontulloch (approved):

PR #753, opened 2022-11-01
SCRUM-1967 VocabularyTermSet fixes

markquintontulloch:


src/main/cliapp/src/service/VocabularyTermSetService.js
```diff
@@ -8,7 +8,7 @@ export class VocabularyTermSetService extends BaseAuthService {
 
 	async deleteVocabularyTermSet(updatedVocabularyTermSet) {
 		const deletionService = new DeletionService();
-		return await deletionService.delete('/vocabularytermset', updatedVocabularyTermSet.id);
+		return await deletionService.delete('vocabularytermset', updatedVocabularyTermSet.id);
```
oblodgett, line 11: @adamgibs the way the "DeletionService" is setup causes this confusion / error. I think it would be better to pull out the `/` so that this is uniform in the service files.

adamgibs: @oblodgett Yeah, sure. That makes sense to me

oblodgett (approved):

PR #757, opened 2022-11-02
SCRUM-2240 Change handling of deleting notes attached to DAs

markquintontulloch:
Not sure if this will fix the deletion FK bug as cannot recreate locally.  However, the use of `searchByField` on `id` rather than a straightforward `find` seems odd and there was no check to ensure that only a single result was being returned.


oblodgett:
@markquintontulloch yes you are right, surprised this wasn't caught in a PR review.

oblodgett (approved):

PR #761, opened 2022-11-03
GH actions maintenance

mluypaert:
GH actions was throwing several warnings on the use of deprecated `actions/checkout@v2` and `set-output`. I put in place replacements for these in this PR.


oblodgett:
Is this a kind of PR that might have multiple iterations?

oblodgett (approved):

mluypaert:
> Is this a kind of PR that might have multiple iterations?

That will depend on whether is deploys successfully to alpha once merge. Potentially yes, but hopefully not.

PR #762, opened 2022-11-04
SCRUM-2252 Added check for string on bioEntity fields

adamgibs:


src/main/cliapp/src/utils/utils.js
```diff
@@ -226,3 +226,32 @@ export function generateInitialTableState(columns, defaultVisibleColumns){
 	}
 	return { defaultColumnNames, initialTableState };
 }
+
+export function validateBioEntityFields(updatedRow, setUiErrorMessages, event, setIsEnabled, closeRowRef, areUiErrors) {
+	const bioEntityFieldNames = ["subject", "diseaseGeneticModifier", "sgdStrainBackground" ];
```
markquintontulloch: We should also add "assertedAllele" to this.  Entering text into the "with" and "assertedGenes" fields doesn't throw an error but just removes the text on saving - I don't know if we can deal with those fields in the same way due to the multi-select nature of those columns, but the current behaviour may be confusing (probably something for a separate ticket though)

markquintontulloch (changes requested):
Apart from the addition of "assertedAllele" to the list of fields to check, this looks good to me

src/main/cliapp/src/utils/utils.js
adamgibs: @markquintontulloch Ah, okay. I'll added assertedAllele. The multi-value fields of all kinds can't submit string types since the only value that they track is the array of values. My think was that adding a string check for the multi-value fields was out of the scope of this ticket, but maybe Chris will think differently.

markquintontulloch: I agree about it being outside the scope of this ticket, as the ticket was to address the fields that throw errors

adamgibs: @markquintontulloch "assertedAllele" has been added to the validation list

markquintontulloch (approved):

markquintontulloch:
@adamgibs might be worth holding off on merging this until Ketaki's code has been merged - the uncaught errors that are fixed by this are a good test case for her stuff

adamgibs:
@markquintontulloch  Got it. Will do.

PR #766, opened 2022-11-08
More node.js 14 to 16 workflow upgrades, SCRUM-2269

mluypaert:


.github/workflows/github-release-build-and-deploy.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       uses: actions/checkout@v3
     - run: cp src/main/resources/application.properties.defaults src/main/resources/application.properties
     - name: AWS credentials configuration
-      uses: aws-actions/configure-aws-credentials@v1
+      uses: aws-actions/configure-aws-credentials@v1-node16
```
oblodgett, line 13: @adamgibs have we tested the UI with node16?

oblodgett: maybe thats not a thing in this context

adamgibs: @oblodgett I think it should be okay. I've tried it locally and it seemed to work. I think create react app only upgraded to Node 16 at the end of last year, so prior to that, it would have caused problems. They appear to have upgraded now. https://github.com/facebook/create-react-app/issues/11037

mluypaert: @oblodgett @adamgibs The changes in this PR really only affect the workflows used in github actions for AWS authentication and EB deployment, not the actual curation application runtimes, so nothing changes regarding the curation UI here.

oblodgett: I realized that after my first message.

oblodgett (approved):

PR #769, opened 2022-11-11
Split repo into submodules in order to only include the needed software  in public API

oblodgett:
Top this @markquintontulloch 1,602 files changed.

The reason for this PR is that Thorntail packages are conflicting with the Quarkus library import. On the public API we only need the Interfaces and the Models. Which "should" eliminate all the conflicting files, if this doesn't work. We may have to upgrade the public API to quarkus.


markquintontulloch:
Meh, a mere 1,425 after merging!  Looks good to me, just need to update the location of application.properties.default in the GitHub workflow files

oblodgett:
Ya there are still a lot of issues with it I figured i would push the PR as a heads up

mluypaert:
@oblodgett As we're needing to update the public API to use quarkus in the long run anyway, is this refactor of the curation repo to work with thorntail really worth it? Can we not prioritise using quarkus in the public API instead?

oblodgett:
So this refactor is a proof of concept, its very broken and does not work on the curation side. However it does split out the code into sub modules, I don't like this, but this is an attempt to see if Thorntail will play nice with Quarkus at least in the interim. And after spending most of yesterday and last night I see that this is not going to work either. So I will be closing this PR as it will rapidly get out of sync with new code and cause a merge nightmare, we are going to have to travel a different road with the public API. (Indexer works great with the library however the API does not).

PR #770, opened 2022-11-11
SCRUM-2267 Backend work for addition of allele mutation types

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/interfaces/crud/AlleleCrudInterface.java
```diff
@@ -38,7 +41,7 @@ public interface AlleleCrudInterface extends BaseCurieCrudInterface<Allele>, Bas
 
 	@Override
 	@GET
-	@JsonView(View.FieldsAndLists.class)
+	@JsonView(View.Allele.class)
```
oblodgett: I think it would be better to rename this to View.AllleleView I know its redundant but at least we won't have two classes in the project named "Allele".

src/main/resources/db/migration/v0.13.0.1__agr_curation_api.sql
```diff
@@ -19,6 +19,10 @@ ALTER TABLE diseaseannotation
 	ALTER COLUMN uniqueid
  		SET NOT NULL;
 	
+ALTER TABLE diseaseannotation
+	ALTER COLUMN uniqueid
+		SET NOT NULL;
+
```
oblodgett: If this is a change to an existing file. It will not get rerun against alpha... so it would need to go in a new file.

oblodgett (changes requested):

src/main/resources/db/migration/v0.13.0.1__agr_curation_api.sql
markquintontulloch: Duplication caused by merging of local branches - now removed

src/main/java/org/alliancegenome/curation_api/interfaces/crud/AlleleCrudInterface.java
markquintontulloch: Done

oblodgett (approved):

oblodgett (approved):

PR #777, opened 2022-11-14
Add indexes for AlleleMutationType

markquintontulloch:


src/main/resources/db/migration/v0.13.0.3__agr_curation_api.sql
```diff
@@ -0,0 +1,4 @@
+CREATE INDEX allelemutationtypeslotannotation_soterm_index ON public.allelemutationtypeslotannotation_soterm USING btree (allelemutationtypeslotannotation_id, mutationtypes_curie);
+CREATE INDEX slotannotation_informationcontententity_index ON public.slotannotation_informationcontententity USING btree (slotannotation_id, evidence_curie);
```
oblodgett: I don't see these indexes defined in the code? Plus it's better to do the indexes separately vs together.

oblodgett (changes requested):

src/main/resources/db/migration/v0.13.0.3__agr_curation_api.sql
markquintontulloch: Where do indexes for join tables get defined in the code?  Do you mean separately as in separate PRs?

markquintontulloch: ah, ok - I see an example in the code - will update

oblodgett: They can get defined on either side of the relationship... just not on both. 

`CREATE INDEX index1 ON public.slotannotation_informationcontententity USING btree (slotannotation_id);`
`CREATE INDEX index2 ON public.slotannotation_informationcontententity USING btree (evidence_curie);`

vs

`CREATE INDEX slotannotation_informationcontententity_index ON public.slotannotation_informationcontententity USING btree (slotannotation_id, evidence_curie);`

markquintontulloch: I was using the following example from a previous migration as a guideline:
`CREATE INDEX idx48t7pceipvyg9kqy9eue51mps ON public.genomicentity_synonym USING btree (genomicentities_curie, synonyms_id);
`
Why is this different?

oblodgett: that is a composite index which means it only gets used in a query if the query contains (genomicentities_curie = 1234 AND synonyms_id = 4567) if only one of the fields is used then the index does not get used at all. It is better if you are always going to have the AND but if not then its better to have the separate indexes so that they will be used instead... ideally you would have 3 indexes the individual ones and the composite.

oblodgett: ya... I didn't catch this:
`@JoinTable(indexes = @Index( columnList = "genomicentities_curie, synonyms_id"))`
Should be:
```java
@JoinTable(indexes = {
	@Index( columnList = "genomicentities_curie"),
	@Index( columnList = "synonyms_id")
})
```

oblodgett: And if you want to go crazy with it:
```java
@JoinTable(indexes = {
        @Index( columnList = "genomicentity_curie, crossreferences_curie")
	@Index( columnList = "genomicentities_curie"),
	@Index( columnList = "synonyms_id")
})
```

oblodgett: It all depends on the queries that hibernate generates... and my suggestion is based on the premise: "we don't know what those queries are", given that, to be safe, single field indexes are the way to go and they will always be used, when those fields are used in a query.

markquintontulloch: ah ok, good to know - updates now pushed

oblodgett (approved):

PR #778, opened 2022-11-14
Remove unnecessary indexes

markquintontulloch:
On 2nd thoughts, probably no need for these two indexes given that AlleleMutationTypes are not reused and only created/edited in the context of Alleles

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/model/entities/slotAnnotations/SlotAnnotation.java
+++ b/src/main/java/org/alliancegenome/curation_api/model/entities/slotAnnotations/SlotAnnotation.java
@@ -33,10 +33,6 @@ import lombok.ToString;
 @Data @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
 @Inheritance(strategy = InheritanceType.JOINED)
 @Schema(name = "SlotAnnotation", description = "POJO that represents a SlotAnnotation")
-@Table(indexes = {
-	@Index(name = "slotannotation_createdby_index", columnList = "createdBy_id"),
-	@Index(name = "slotannotation_updatedby_index", columnList = "updatedBy_id"),
-})
 @ToString(callSuper = true)
 @AGRCurationSchemaVersion(min="1.4.0", max=LinkMLSchemaConstants.LATEST_RELEASE, dependencies={AuditedObject.class})
 public class SlotAnnotation extends GeneratedAuditedObject {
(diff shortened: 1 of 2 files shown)


markquintontulloch:
And the same will apply for other SlotAnnotations

oblodgett:
Well... if Alleles have a list of AlleleMutationTypes even if its a back reference, then hibernate will join to those objects... so I would leave these indexes.

markquintontulloch:
how are updatedBy and createdBy involved in that?

oblodgett:
AlleleMutationTypeSlotAnnotation -> is_a -> SlotAnnotation -> is_a -> GeneratedAuditedObject -> is_a -> AuditedObject

To hibernate AlleleMutationTypeSlotAnnotation will have an updated_by and created_by which will need to be indexed.
Both AuditedObject, and GeneratedAuditedObject are a @MappedSuperclass which means the fields will be on the SlotAnnotation table which is where the index needs to be.

markquintontulloch:
I guess what I'm not understanding is why (for SlotAnnotations) updated_by and created_by need to be indexed.  I can't think when we will be querying for SlotAnnotations using those fields.  I'm probably missing something obvious though.

oblodgett:
Given this on the allele:
```java
	@OneToMany(mappedBy = "singleAllele", cascade = CascadeType.ALL)
	@JsonManagedReference
	@JsonView({View.FieldsAndLists.class, View.AlleleView.class})
	private List<AlleleMutationTypeSlotAnnotation> alleleMutationTypes;
```
`@OneToMany` by its nature is LAZY so pulling Alleles will not add the alleleMutationTypes. However for indexing `@IndexedEmbedded(includeDepth = 2)` means that the mass indexer is going to pull the alleleMutationTypes. Which will result in a join to the person table. Looking at all the fields on the AlleleMutationTypeSlotAnnotation

```java
@IndexedEmbedded(includeDepth = 1) @ManyToOne (EAGER) private Allele singleAllele;
@IndexedEmbedded(includeDepth = 1) @ManyToMany (LAZY) private List<SOTerm> mutationTypes;
@IndexedEmbedded(includeDepth = 2) @ManyToMany (LAZY) private List<InformationContentEntity> evidence;
@IndexedEmbedded(includeDepth = 1) @ManyToOne (EAGER) private Person createdBy;
@IndexedEmbedded(includeDepth = 1) @ManyToOne (EAGER) private Person updatedBy;
```

So in short every object, that is a `AuditedObject` will "eagerly" join to the persons table in order to get those fields.

markquintontulloch:
Ok, makes sense - thanks for explaining.  Closing this PR.

PR #779, opened 2022-11-15
Index changes mk2

markquintontulloch:
@oblodgett - What about this approach?  Would this work?

If we specify the specific fields we want to index then the mass indexer won't try to index the AuditedObject fields, avoiding the need for the indexes on those fields for SlotAnnotations.

Seems to speed the allele load a bit on my local system (although not drastically).


oblodgett:
@markquintontulloch having the indexes in place slows down the import?

markquintontulloch:
Yep, after adding the indexes yesterday the allele load went from ~28r/s to ~7

PR #782, opened 2022-11-15
Initial Code for RO loading

oblodgett:


adamgibs:
@oblodgett I think a GeneralOntologyComponent needs to be added to the routes.js for the table to be accessible

oblodgett:
@adamgibs ahh ok that is what was missing, thanks.

markquintontulloch (approved):

oblodgett:
Forgot the migration file.

PR #795, opened 2022-11-23
SCRUM-1986: AGR Curation Pre-Release v0.13.0 rc1

abecerra:


mluypaert:
@abecerra You'll need to remove the current `v0.13.0-rc1` tag you just created (currently pointing to commit e893d68) and recreate it once this PR gets merge into beta, as that tag needs to point to the merge commit of this PR (which currently does not exist yet, as it only gets created after approval and merge).

mluypaert:
Also, I wonder if there is a point in going through this beta env deployment still, as we're not going to be able to complete this before sprint review this afternoon. Perhaps it would be better for us to bundle it in a single deployment together with the changes about to be demoed (and hence not do a production deployment at the start of the next sprint either).

PR #797, opened 2022-11-24
SCRUM-2327 Exclude obsolete results from AutoComplete suggestions

markquintontulloch:
Exclude obsolete terms from AutoComplete suggestions by default.  Can be overridden for cases where obsolete fields are required or potential future cases where obsolete is not a field of the searched class.


src/main/cliapp/src/utils/utils.js
```diff
@@ -211,8 +211,15 @@ export function onSelectionOver(event, item, query, op, setAutocompleteHoverItem
 }
 
 
-export function autocompleteSearch(searchService, endpoint, filterName, filter, setSuggestions, otherFilters={}) {
-	searchService.search(endpoint, 15, 0, [], {[filterName]: filter, ...otherFilters})
+export function autocompleteSearch(searchService, endpoint, filterName, filter, setSuggestions, otherFilters={}, applyObsoleteFilter=true) {
+	const obsoleteFilter = applyObsoleteFilter ? {
```
adamgibs, line 215: I think this makes sense here for now. Perhaps in the future, if we have a lot of commonly used filters or default filters, they could be pulled out into their own file and just imported in when needed. Perhaps by the component calling the autocompleteSearch function.

adamgibs (approved):

PR #802, opened 2022-11-26
SCRUM-1484 Increase max_gram size to cater for longest curies

markquintontulloch:


kthorat-prog (approved):

src/main/resources/hibernate/custom-index-settings.json
```diff
@@ -1,7 +1,7 @@
 {
   "number_of_shards": "4",
   "number_of_replicas": "0",
-  "max_ngram_diff": "20",
+  "max_ngram_diff": "29",
```
oblodgett, line 4: @markquintontulloch there is also a local-index-setting file. Those will need to be changed also otherwise it wont work when deving locally.

oblodgett: @mluypaert this also has an impact on index size, might want to keep an eye on it.

PR #808, opened 2022-11-30
added for for alliance member lookup

oblodgett:


abecerra (approved):

markquintontulloch (approved):

oblodgett:
I am going to need to run this migration by hand considering it contains email addresses that I don't want to put into a commit. Or I mean the part that I didn't include in this PR.

PR #809, opened 2022-11-30
clear button working

kthorat-prog:


src/main/cliapp/src/components/Autocomplete/AutocompleteFormEditor.js
```diff
@@ -0,0 +1,54 @@
...
+				itemTemplate={itemTemplate}
+				completeMethod={(event) => search(event, setSuggestions, setQuery, rowProps)}
+				onHide={(e) => op.current.hide(e)}
+				onChange={(e) => onValueChangeHandler(e)}
+				className={classNames}
+			/>
+			<EditorTooltip op={op} autocompleteHoverItem={autocompleteHoverItem} dataType={fieldName}/>
+		</div>
+	)
+}
```
oblodgett, line 54: Are these new files?

oblodgett (commented):
And are there a delete of the old ones?

src/main/cliapp/src/components/Autocomplete/AutocompleteFormEditor.js
kthorat-prog: yes. These are new files created to solve this issue.

oblodgett: but are they copies of the other ones?

kthorat-prog: yes. Except for the fieldValue & setFieldValue things which needs to be avoided in order to clear the fields. Because they make it controlled behaviour and doesn't clear the values.

adamgibs (approved):

PR #815, opened 2022-12-06
SCRUM-2345 Update to LinkML v1.5.0 (NameSlotAnnotations)

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/interfaces/crud/GeneCrudInterface.java
```diff
@@ -21,12 +34,47 @@
...
+	public SearchResponse<Gene> find(
+			@DefaultValue("0") @QueryParam("page") Integer page,
+			@DefaultValue("10") @QueryParam("limit") Integer limit,
+			@RequestBody HashMap<String, Object> params);
+	
+	@Override
+	@POST
+	@Path("/search")
+	@Tag(name = "Elastic Search Endpoints")
+	@JsonView({View.GeneView.class})
```
oblodgett, line 73: Could probably promote `@JsonView({View.GeneView.class})` so the class level.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/interfaces/crud/GeneCrudInterface.java
markquintontulloch: I think we tried that with the AlleleView.class and couldn't get it to work.  There must be a way of doing that to avoid all these overrides though - maybe something I can look at in the off release sprint.

oblodgett:
`Failed to start application (with profile prod): org.postgresql.util.PSQLException: ERROR: more than one row returned by a subquery used as an expression`

markquintontulloch:
Yup.  Tested the migration on a recent dump of production, which was dumb because most of the affected data hasn't made it to production yet - doh!  Dumping alpha now to work out what's going on.

oblodgett:
I restarted the server and things a looking a little different but not sure that its up

markquintontulloch:
yeah, the migration takes a while before it fails - I expect it to imminently

oblodgett:
yep still failed

oblodgett:
@markquintontulloch I think at this point if you made a change to the migration file itself would be the best bet considering flyway is not able to move forward with 0.13.0.8.

PR #816, opened 2022-12-06
Added subject readable data underneath it

kthorat-prog:


src/main/cliapp/src/components/FormAdditionalFieldData.js
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+
+
+export function FormAdditionalFieldData({ field, fieldData }){
```
adamgibs, line 4: @kthorat-prog What is this field prop used for? I may be mistaken, but it looks like it's not being referenced within the component.

kthorat-prog: That is for other fields, whenever introduced for this.

adamgibs: Got it. Makes sense.

adamgibs (approved):

PR #818, opened 2022-12-07
SCRUM-2345 Add indexes for NameSlotAnnotation tables

markquintontulloch:


oblodgett (approved):

oblodgett:
@markquintontulloch `Caused by: org.hibernate.search.util.common.SearchException: HSEARCH000610: Unknown field 'subject.symbol'.` not sure if that is due to this PR or not?

oblodgett:
Guessing it should now be subject.geneSymbol.displayText?

markquintontulloch:
Yeah.  It hadn't occurred to me that all the changes for NameSlotAnnotations are going to screw up the disease annotations.   The problem we have is that it could be subject.geneSymbol.displayText or subject.alleleSymbol.displayText.  And for names it could be subject.geneFullName.displayText, subject.alleleFullName.displayText, or subject.name (for AGMs which haven't changed).

markquintontulloch:
I guess we need a PropertyBridge, like we use for diseaseGeneticModifier

markquintontulloch:
In fact, maybe we can use the same one

oblodgett:
Ya at least all the search columns for subject are going to change. There is another way where you can put certain "paths" on the document as fields instead of the whole object.

oblodgett:
Why do we need a property bridge?

markquintontulloch:
So that we can search for symbol/name despite the different field names.  Not sure how we deal with it in the display of subject though.

oblodgett:
I mean... we just need to switch over the UI to use all the new fields.

markquintontulloch:
So how would we deal with the following?
`{
		field: "subject.symbol",
		header: "Subject",
		sortable: isEnabled,
		filter: true,
		filterElement: {type: "input", filterName: "subjectFilter", fields: ["subject.symbol", "subject.name", "subject.curie"]},
		editor: (props) => subjectEditorTemplate(props),
		body: subjectBodyTemplate,
	},`

I guess we can put all possible options in the filterElement fields, but what about `field`?

oblodgett:
yes all the possible fields as it is a OR

oblodgett:
so even when it references fields that don't exist on some documents, those don't cause an error in ES

PR #819, opened 2022-12-07
Fixed profile page

oblodgett:


oblodgett:
This is ridiculous that a UI code change fails IT.

oblodgett:
@mluypaert the more this happens I think we might need to consider running our own github runner.

mluypaert:
> @mluypaert the more this happens I think we might need to consider running our own github runner.

I don't think that would make our lives any easier, or fix the issue. Assuming the integration tests also randomly fail when you run them on your local system, running our own github runner would not prevent these random itermittent failures. We should try to figure out why these test faile intermittendly. If these failure only ever occur on GH actions (I don't believe so) then we'd need to figure out why that is instead of trying to fix it by maintaining our own runners.

markquintontulloch (approved):

PR #821, opened 2022-12-07
add convenience method to data provider (Organization) class

cmpich:


src/main/java/org/alliancegenome/curation_api/model/entities/DiseaseAnnotation.java
```diff
@@ -178,4 +178,12 @@ public abstract class DiseaseAnnotation extends Association {
 	@Transient
 	public abstract String getSubjectTaxonCurie();
 
+	public String getDataProviderString(){
```
oblodgett, line 184: @cmpich this needs to be @Transient also.

oblodgett: Plus can't this get done on the UI side? @adamgibs?

cmpich: I need this in the download logic of the API.

src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
```diff
@@ -16,27 +13,56 @@
...
-	@JoinTable(indexes = { @Index(columnList = "Reference_curie"), @Index(columnList = "crossReferences_curie") })
+	@JsonView({View.FieldsOnly.class})
+	@JoinTable(indexes = {@Index(columnList = "Reference_curie"), @Index(columnList = "crossReferences_curie")})
 	@EqualsAndHashCode.Include
 	private List<CrossReference> crossReferences;
 
+
+	public static List<String> speciesPrefix;
+
+	static {
```
oblodgett: @cmpich can this be moved into a Enum... we need to have a enum package that can be used for this kind of thing.

oblodgett (approved):

PR #832, opened 2022-12-16
add getReferenceID() method to Reference class

cmpich:


oblodgett:
@cmpich it looks like you didn't `git pull` before creating this branch. So the branch is taken from an OLD version of production.

oblodgett (changes requested):
Update your production branch first before taking the branch.

oblodgett:
@mluypaert how do we create a patch for production?

oblodgett:
@cmpich sorry the branch for this needs to be called `release/v0.14.1`

PR #833, opened 2022-12-16
Release/v0.14.1

oblodgett:


src/main/java/org/alliancegenome/curation_api/enums/CrossReferencePrefix.java
```diff
@@ -0,0 +1,5 @@
+package org.alliancegenome.curation_api.enums;
+
+public enum CrossReferencePrefix {
+	   MGI, ZFIN, SGD, FB, WB, XB, OMIM, GO
```
cmpich: Need to add 
RGD

src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
```diff
@@ -39,4 +43,17 @@ public class Reference extends InformationContentEntity {
...
+	 */
+	@Transient
+	@JsonIgnore
+	public String getReferenceID() {
+		Optional<CrossReference> opt = getCrossReferences().stream().filter(reference -> reference.getCurie().startsWith("PMID:")).findFirst();
+		// if no PUBMED ID try MOD ID
+		if (opt.isEmpty()) {
+			opt = getCrossReferences().stream().filter(reference -> CrossReferencePrefix.valueOf(reference.getPrefix()) != null).findFirst();
+		}
+		return opt.map(CrossReference::getPrefix).orElse(null);
```
cmpich: This needs to be
		return opt.map(CrossReference::getCurie).orElse(null);

oblodgett:
@cmpich do you want to make these changes?

cmpich (approved):
Looks good.

PR #836, opened 2022-12-19
Production v0.14.1 into Alpha

oblodgett:


markquintontulloch (approved):

src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
```diff
@@ -49,8 +66,6 @@ public String getReferenceID() {
 		if (opt.isEmpty()) {
 			opt = getCrossReferences().stream().filter(reference -> CrossReferencePrefix.valueOf(reference.getPrefix()) != null).findFirst();
 		}
-		return opt.map(CrossReference::getPrefix).orElse(null);
+		return opt.map(CrossReference::getCurie).orElse(null);
```
oblodgett, line 69: @cmpich are you sure this is right? getCurie is not the prefix?

PR #839, opened 2022-12-20
add fish as a subtype of affectdGenomicModel

cmpich:


oblodgett (approved):
@cmpich can you hold off on this PR until the Alleles are done indexing?

markquintontulloch:
@cmpich - This enum code has been removed, along with the column in the database, in my current PR which is waiting to be merged. I'm not sure how adding 'fish' here will help you because ,as we discussed, we don't have any code for loading subtype via the bulk loads and we need to reimplement this as a VocabularyTerm - I was planning on doing this shortly.  If you need this for some reason then let me know and I'll have to adjust my PR.

PR #845, opened 2023-01-04
Human readable data appears below field

kthorat-prog:


adamgibs (approved):

oblodgett:
@kthorat-prog @adamgibs so with this change I would suggest breaking this out into separate components. There is really no point in have one component and then having separate logic for each case. Just make 6 different components and keep the logic all internal to the component. (Component's are cheap maintenance is expensive)

oblodgett:
For instance if you were to have a `<SubjectAdditionalData subject="data.subject"/>`  Then this can be reused anywhere that we want to see "Additional Data" for the subject field.

PS The more I learn about react the more I am going to push for breaking things out into separate components.

adamgibs:
@oblodgett @kthorat-prog makes sense to me. Also, I think the UI build is failing because of the breaks after the returns in the switch. I don't think they are needed if you are already returning.

src/main/cliapp/src/components/SingleReferenceAdditionalFieldData.js
```diff
@@ -0,0 +1,26 @@
...
+				}else if(fieldData['cross_references'][i].curie.indexOf('MGI:') > -1){
+					return <div className="p-info">{fieldData['cross_references'][i].curie}</div>;
+				}else if(fieldData['cross_references'][i].curie.indexOf('RGD:') > -1){
+					return <div className="p-info">{fieldData['cross_references'][i].curie}</div>;
+				}else if(fieldData['cross_references'][i].curie.indexOf('SGD:') > -1){
+					return <div className="p-info">{fieldData['cross_references'][i].curie}</div>;
+				}else if(fieldData['cross_references'][i].curie.indexOf('WB:') > -1){
+					return <div className="p-info">{fieldData['cross_references'][i].curie}</div>;
+				}else if(fieldData['cross_references'][i].curie.indexOf('ZFIN:') > -1){
+					return <div className="p-info">{fieldData['cross_references'][i].curie}</div>;
```
oblodgett, line 20: @kthorat-prog are these not all the same?

kthorat-prog: no ....
also, I am not pushing this branch. I have created a new one and working on it. This branch seems to have a lot of changes which are already on alpha. To avoid any merge complications, I will push changes on new branch.

oblodgett: It just wasn't updated

oblodgett: Maybe I am missing something here but are not all these cases doing the same thing?

oblodgett: Couldn't we just split the `fieldData['cross_references'][i].curie` via `:` and then check that the prefix is in a list.

kthorat-prog: No. All cases have different stuff to display.

oblodgett: But the return is the same

kthorat-prog:
Created another PR for same

PR #850, opened 2023-01-09
SCRUM-2189 - Load OBI ontology

markquintontulloch:


mluypaert (commented):
I think the migration file ([src/main/resources/db/migration/v0.15.0.1__agr_curation_api.sql](https://github.com/alliance-genome/agr_curation/pull/850/files#diff-69650eaba95d39b46939b71830048103728f9a9db0d2ed7cdfa413998bc3cea7)) should be called `v0.14.1.1__agr_curation_api.sql` instead? As this is the first migration that would get applied to that version?

src/main/java/org/alliancegenome/curation_api/controllers/crud/ontology/ObiTermCrudController.java
```diff
@@ -0,0 +1,28 @@
...
+public class ObiTermCrudController extends BaseOntologyTermController<ObiTermService, OBITerm, ObiTermDAO> implements ObiTermCrudInterface {
+
+	@Inject
+	ObiTermService obiTermService;
+
+	@Override
+	@PostConstruct
+	public void init() {
+		GenericOntologyLoadConfig config = new GenericOntologyLoadConfig();
+		config.setLoadOnlyIRIPrefix("OBI");
```
oblodgett, line 24: @markquintontulloch can this work without setting the prefix?

markquintontulloch: sorry, not sure what you mean

markquintontulloch:
@mluypaert - our versioning is off.  Alpha, beta and production are all on 0.14.1.  Alpha should have incremented to 0.15.0 shouldn't it?

mluypaert:
> @mluypaert - our versioning is off. Alpha, beta and production are all on 0.14.1. Alpha should have incremented to 0.15.0 shouldn't it?

@markquintontulloch No, as we cannot be sure on the next release versions' number until we actually created it, we decided to base the migration file versioning on the version current on production, as in the version on which the migration will be applied, not the version in which it is being included. Once used in production, then both the major and the minor version number could need increasing, depending on the features implemented in the new release. See the [README](https://github.com/alliance-genome/agr_curation#data-structure-changes) for a more detailed description of this versioning.

src/main/java/org/alliancegenome/curation_api/controllers/crud/ontology/ObiTermCrudController.java
oblodgett: Leaving `config.setLoadOnlyIRIPrefix("OBI");` out?

markquintontulloch: The helper just checks that the IRI short form starts with the prefix set here:
`boolean condition 4 = currentTreeNode.getIRI().getShortForm().startsWith(config.getLoadOnlyIRIPrefix() + "_");`

oblodgett: right but that is in a OR condition if condition 1 and 2 then it doesn't need 3 and 4

markquintontulloch:
@mluypaert  - are you saying that alpha shouldn't be on v0.15.0-xxxx or that migration file versions shouldn't be based on alpha.  Historically the migration file names have been based on the alpha version, but we've had this conversation before (see a-team-code Slack channel on October 17th).

src/main/java/org/alliancegenome/curation_api/controllers/crud/ontology/ObiTermCrudController.java
markquintontulloch: ah, I see what you mean - no, it wouldn't work as there is no namespace set

mluypaert:
> @mluypaert - are you saying that alpha shouldn't be on v0.15.0-xxxx or that migration file versions shouldn't be based on alpha. Historically the migration file names have been based on the alpha version, but we've had this conversation before (see a-team-code Slack channel on October 17th).

As the most recent (prod) release is v0.14.1, alpha is (correctly) reporting version `v0.14.1-84-g1ef458f0`, which means as much as _this is v0.14.1 plus 84 commits, of which the latest is 1ef458f0_. The migration file versions shouldn't be based on alpha, but rather on the current prod version they will be applied (see README). I'm aware we've discussed this before, but it seems we didn't make an agreement on a resolution, and just continued with what was written down in the README.

src/main/java/org/alliancegenome/curation_api/controllers/crud/ontology/ObiTermCrudController.java
oblodgett: ahh ok... just checking

oblodgett:
@mluypaert @markquintontulloch to me what seems more clear and more straight forward is to have the DB version follow the LinkML version rather than the code version, and if there are changes that need to happen to the java model that are not link ML driven, then those should be the 0.0.0.1 increments.

mluypaert:
> @mluypaert @markquintontulloch to me what seems more clear and more straight forward is to have the DB version follow the LinkML version rather than the code version, and if there are changes that need to happen to the java model that are not link ML driven, then those should be the 0.0.0.1 increments.

@oblodgett @markquintontulloch Although that would make sense theoretically, I'm not sure practically because of the dispersed implementation of different linkML versions in different parts of our code. The same LinkML version will not be implemented in all classes, so you'd need to look at individual classes to figure out the relevant linkML version for each anyway anyway (regardless of the migration file version).

mluypaert (approved):

PR #851, opened 2023-01-10
SCRUM-2294 add standard and generic-control experiments as default va…

cmpich:
…lues for ZFIN annotations.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/model/entities/ConditionRelation.java
+++ b/src/main/java/org/alliancegenome/curation_api/model/entities/ConditionRelation.java
@@ -37,6 +37,9 @@ import lombok.EqualsAndHashCode;
 @AGRCurationSchemaVersion(min = "1.2.0", max = LinkMLSchemaConstants.LATEST_RELEASE, dependencies = { ExperimentalCondition.class })
 public class ConditionRelation extends UniqueIdAuditedObject {
 
+	public static final String STANDARD = "Standard";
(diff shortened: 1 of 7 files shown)


src/main/java/org/alliancegenome/curation_api/model/entities/ConditionRelation.java
```diff
@@ -37,6 +37,9 @@
 @AGRCurationSchemaVersion(min = "1.2.0", max = LinkMLSchemaConstants.LATEST_RELEASE, dependencies = { ExperimentalCondition.class })
 public class ConditionRelation extends UniqueIdAuditedObject {
 
+	public static final String STANDARD = "Standard";
+	public static final String GENERIC_CONTROL = "Generic-control";
```
oblodgett: Can these go into the constants class?

cmpich: I prefer having those constants on the class they belong to.

oblodgett: Are they used outside the class at all? If not change them to private.

cmpich: I put them into an inner class of ConditionRelation. Constant which then makes it obvious what this constant is for and that it is a constant.

oblodgett (approved):

PR #853, opened 2023-01-11
add ObjectResponse to exception

cmpich:


src/main/java/org/alliancegenome/curation_api/controllers/crud/ConditionRelationCrudController.java
```diff
@@ -48,11 +48,15 @@ public SearchResponse<ConditionRelation> findExperiments(Integer page, Integer l
...
+			ObjectResponse<ConditionRelation> response = new ObjectResponse<>();
+			response.addErrorMessage(key, "Cannot find any reference ID under map key: " + key);
+			throw new ApiErrorException(response);
 		}
 		Reference reference = referenceService.get(referenceID).getEntity();
 		if (ObjectUtils.isEmpty(reference)) {
-			throw new ApiErrorException("Cannot find reference for given reference ID: " + referenceID);
+			ObjectResponse<ConditionRelation> response = new ObjectResponse<>();
+			response.addErrorMessage(key, "Cannot find reference for given reference ID: " + referenceID);
+			throw new ApiErrorException(response);
```
oblodgett: @cmpich I didn't catch this earlier this whole method should be in the service not the controller.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/controllers/crud/ConditionRelationCrudController.java
oblodgett: PS and the update method doesn't even need to be defined.

cmpich: The method find-experiments is the endpoint method. The body of it can be moved into the service class, yes.

cmpich: The update method, I am not sure about if it is used.

cmpich: It looks like it's not used.

src/main/java/org/alliancegenome/curation_api/services/ConditionRelationService.java
```diff
@@ -71,4 +78,35 @@ public ConditionRelation getStandardExperiment(String experimentName, Reference
...
+
+	public SearchResponse<ConditionRelation> getConditionRelationSearchResponse(HashMap<String, Object> params) {
+		String key = "singleReference.curie";
+		String referenceID = (String) params.get(key);
+		if (StringUtils.isEmpty(referenceID)) {
+			ObjectResponse<ConditionRelation> response = new ObjectResponse<>();
+			response.addErrorMessage(key, "Cannot find any reference ID under map key: " + key);
+			throw new ApiErrorException(response);
+		}
+		Reference reference = referenceService.get(referenceID).getEntity();
```
oblodgett, line 90: You can probally just use the DAO here instead of the service but whatever.

src/main/java/org/alliancegenome/curation_api/interfaces/crud/ConditionRelationCrudInterface.java
```diff
@@ -45,6 +45,6 @@ public interface ConditionRelationCrudInterface extends BaseIdCrudInterface<Cond
 	@POST
 	@Path("/find-experiments")
 	@JsonView(View.FieldsAndLists.class)
-	SearchResponse<ConditionRelation> findExperiments(@DefaultValue("0") @QueryParam("page") Integer page, @DefaultValue("10") @QueryParam("limit") Integer limit, @RequestBody HashMap<String, Object> params);
+	SearchResponse<ConditionRelation> findExperiments(@RequestBody HashMap<String, Object> params);
```
oblodgett, line 48: And I assume the assumption here is that the list will always be small?

cmpich: Yes, no pagination needed.

src/main/java/org/alliancegenome/curation_api/services/ConditionRelationService.java
cmpich: Maybe. I'll leave it for now.

oblodgett (approved):

PR #862, opened 2023-01-13
SCRUM-2128 added DA settings for MODs

adamgibs:


src/main/cliapp/src/service/TableStateService.js
```diff
@@ -0,0 +1,177 @@
...
+			first: 0,
+			rows: 50,
+			multiSortMeta: [],
+			selectedColumnNames: defaultVisibleColumns ? defaultVisibleColumns : defaultColumnNames,
+			filters: {},
+			isFirst: false,
+			tableKeyName: table,
+			tableSettingsKeyName: `${table}TableSettings`
+	}
+}
```
oblodgett, line 177: @adamgibs just wondering if there might be a better way to do this given that this file is going to get huge. I know this was the initial idea. Maybe just leave it for now but just thinking.

oblodgett (approved):

src/main/cliapp/src/service/TableStateService.js
adamgibs: Yeah, it will be massive once we have all the tables included. That said, if we ever have an API set up for this, this map could go away.

PR #868, opened 2023-01-18
connect standard conditions to existing experimentcondition

cmpich:


oblodgett (approved):
Not sure about this hardcoding: `condition.setUniqueId("ZECO:0000103");` but whatever.

PR #869, opened 2023-01-18
SCRUM-2478 added check for empty okta-token-storage

adamgibs:


src/main/cliapp/src/service/BaseAuthService.js
```diff
@@ -6,7 +6,11 @@ export class BaseAuthService {
 
 	constructor() {
 		let oktaTokenStorage = localStorage.getItem('okta-token-storage');
+
+		if(!oktaTokenStorage) oktaTokenStorage = JSON.stringify({});
```
oblodgett: @adamgibs does this check for an empty string also?

adamgibs: Nope, only if exists or not. Should it also check for an empty string?

oblodgett: I would think so...actually what would be better is to catch the error of `JSON.parse` and if there is an error set it to undefined then the browser won't fail but the person won't be logged in either. Might want to think about that for the other cases of `JSON.parse` that we use.

oblodgett (approved):

PR #872, opened 2023-01-19
SCRUM-2481 added error message for duplicate disease annotations

adamgibs:
This is a draft PR. I just want to ask if there are any other situations when there would be a UniqueId error or if its a safe assumption that it's existence means that there is a duplication error?


markquintontulloch:
@adamgibs - The only error thrown on the uniqueId field is the non-unique error.  I can't see any reason for that to change in the future either.

markquintontulloch (approved):

oblodgett (approved):

PR #873, opened 2023-01-20
Added scheduled indexer

oblodgett:


markquintontulloch (approved):

src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
```diff
@@ -134,6 +134,14 @@ public void runGroupJobs() {
 			}
 		}
 	}
+	
+	@Scheduled(cron = "0 0 0 * * SUN")
```
mluypaert: Seems the cron schedule you've defined here is not valid: https://github.com/alliance-genome/agr_curation/actions/runs/3968827023/jobs/6802505838#step:11:1664

mluypaert (changes requested):

mluypaert (approved):

PR #874, opened 2023-01-20
SCRUM-2477 Prevent loading of biological entries from non-canonical species

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/controllers/crud/AGMDiseaseAnnotationCrudController.java
```diff
@@ -50,37 +50,37 @@ public APIResponse updateAgmDiseaseAnnotations(String taxonID, List<AGMDiseaseAn
...
 	@Override
 	public APIResponse updateHumanAgmDiseaseAnnotations(List<AGMDiseaseAnnotationDTO> annotations) {
-		return agmDiseaseAnnotationExecutor.runLoad("NCBITaxon:9606", annotations);
+		return agmDiseaseAnnotationExecutor.runLoad("Homo sapiens", annotations);
 	}
 
 	@Override
 	public APIResponse updateSgdAgmDiseaseAnnotations(List<AGMDiseaseAnnotationDTO> annotations) {
-		return agmDiseaseAnnotationExecutor.runLoad("NCBITaxon:559292", annotations);
+		return agmDiseaseAnnotationExecutor.runLoad("Saccharomyces cerevisiae", annotations);
```
oblodgett: @cmpich @markquintontulloch are these endpoints even used? Sense we now load all these via bulk loads?

src/main/java/org/alliancegenome/curation_api/constants/ValidationConstants.java
```diff
@@ -1,11 +1,33 @@
...
+	
+	public static final List<String> CANONICAL_TAXONS = List.of(
+			"NCBITaxon:7955",
+			"NCBITaxon:9606",
+			"NCBITaxon:10116",
+			"NCBITaxon:10090",
+			"NCBITaxon:6239",
+			"NCBITaxon:7227",
+			"NCBITaxon:559292"
+			);
```
oblodgett: Should these be a Enum instead with both values mapped to each other.

src/main/java/org/alliancegenome/curation_api/dao/AGMDiseaseAnnotationDAO.java
```diff
@@ -15,9 +15,9 @@ protected AGMDiseaseAnnotationDAO() {
 		super(AGMDiseaseAnnotation.class);
 	}
 
-	public List<Long> findAllAnnotationIds(String taxonID) {
-		Query jpqlQuery = entityManager.createQuery("SELECT annotation.id FROM AGMDiseaseAnnotation annotation WHERE annotation.subject.taxon.curie=:taxonId");
-		jpqlQuery.setParameter("taxonId", taxonID);
+	public List<Long> findAllAnnotationIds(String speciesName) {
+		Query jpqlQuery = entityManager.createQuery("SELECT annotation.id FROM AGMDiseaseAnnotation annotation WHERE annotation.subject.taxon.name LIKE CONCAT(:speciesName, '%')");
+		jpqlQuery.setParameter("speciesName", speciesName);
```
oblodgett, line 20: Doing a like query is going to be slow vs using a id field.

src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/DiseaseAnnotationCurieManager.java
```diff
@@ -5,29 +5,29 @@
...
-		curieTaxonMap.put("NCBITaxon:6239", new WormDiseaseAnnotationCurie());
-		curieTaxonMap.put("NCBITaxon:7227", new FlyDiseaseAnnotationCurie());
-		curieTaxonMap.put("NCBITaxon:559292", new SGDDiseaseAnnotationCurie());
+		curieSpeciesNameMap.put("Danio rerio", new ZFINDiseaseAnnotationCurie());
+		curieSpeciesNameMap.put("Homo sapiens", new RGDDiseaseAnnotationCurie());
+		curieSpeciesNameMap.put("Rattus norvegicus", new RGDDiseaseAnnotationCurie());
+		curieSpeciesNameMap.put("Mus musculus", new MGIDiseaseAnnotationCurie());
+		curieSpeciesNameMap.put("Caenorhabditis elegans", new WormDiseaseAnnotationCurie());
+		curieSpeciesNameMap.put("Drosophila melanogaster", new FlyDiseaseAnnotationCurie());
+		curieSpeciesNameMap.put("Saccharomyces cerevisiae", new SGDDiseaseAnnotationCurie());
```
oblodgett: This could be using the ENUM also.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/controllers/crud/AGMDiseaseAnnotationCrudController.java
markquintontulloch: Used in integration tests

src/main/java/org/alliancegenome/curation_api/constants/ValidationConstants.java
markquintontulloch: They don’t need to be mapped to each other

src/main/java/org/alliancegenome/curation_api/dao/AGMDiseaseAnnotationDAO.java
markquintontulloch: Sure, but we don’t know all the IDs (and it may not be just 1) unless we do another query and this query is only run once per load

src/main/java/org/alliancegenome/curation_api/services/helpers/diseaseAnnotations/DiseaseAnnotationCurieManager.java
markquintontulloch: This is how it was implemented with taxon IDs - any reason to change?

oblodgett: Doesn't mean the implementation was correct at the time. I was under the impression that the endpoints were going to go away.

markquintontulloch: Sure, I don’t mind changing it if there is a reason that an enum would be better. Just wondering what that reason is?

markquintontulloch: @oblodgett Given you can't have spaces in an enum constant, wouldn't it be less efficient to have to do a string conversion and then check the enum rather than just using a map?  Or is there some advantage of an enum that I'm not aware of?

oblodgett: This way the strings, ids are in 1 spot. Also you can do the lookup in the enum. This is what we do in the:
https://github.com/alliance-genome/agr_java_software/blob/master/agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/SpeciesType.java

And there is helper methods to do the lookups and conversions.

markquintontulloch: But the strings and IDs are not a 1-to-1 relationship.  I think you are suggesting that the enum constants would be the species names (converted to a form that is valid for an enum constant), but there are multiple taxon IDs for Saccharomyces cerevisiae, for example.

oblodgett: you can put a list in an enum and do that lookup that way.

markquintontulloch: But do we really want to list out all the NCBITaxon IDs for Saccharomyces cerevisiae?  There are 37 in our database, and probably a whole lot more that just haven't been loaded into our database (yet).

oblodgett: What about using a vocabularySet? Im trying to avoid hardcoded strings in the code. If this code has them, then some other code has them, then we have hardcoded lists of string all over the code and when we need to "add another one" it makes that task as well as testing very difficult

markquintontulloch: We could just lose the CANONICAL_TAXONS list in the ValidationConstants.  The only place it is used is the genomic entity validation to check the taxon ID - if it's not found then it compares species name, so the taxon ID check isn't really necessary I just wrote it like that as in the majority of cases the taxon ID will match and we don't have to do any processing of the species name.  That's probably an insignificant gain though, so if we lose that then we are only ever doing a lookup on species name, in which case an enum makes sense.

markquintontulloch:
@oblodgett - I think this latest commit addresses you concerns

src/main/java/org/alliancegenome/curation_api/controllers/crud/AGMDiseaseAnnotationCrudController.java
oblodgett: @markquintontulloch can the enum be used here also?

src/main/java/org/alliancegenome/curation_api/controllers/crud/AlleleDiseaseAnnotationCrudController.java
```diff
@@ -43,43 +43,43 @@ public ObjectResponse<AlleleDiseaseAnnotation> get(String uniqueId) {
...
 	@Override
 	public APIResponse updateHumanAlleleDiseaseAnnotations(List<AlleleDiseaseAnnotationDTO> annotations) {
-		return alleleDiseaseAnnotationExecutor.runLoad("NCBITaxon:9606", annotations);
+		return alleleDiseaseAnnotationExecutor.runLoad("Homo sapiens", annotations);
 	}
 
 	@Override
 	public APIResponse updateSgdAlleleDiseaseAnnotations(List<AlleleDiseaseAnnotationDTO> annotations) {
-		return alleleDiseaseAnnotationExecutor.runLoad("NCBITaxon:559292", annotations);
+		return alleleDiseaseAnnotationExecutor.runLoad("Saccharomyces cerevisiae", annotations);
```
oblodgett: @markquintontulloch and here?

markquintontulloch: @oblodgett  - ditched the MOD-specific endpoints in favour of just using /bulk/{dataType}/annotationFile endpoints

src/main/java/org/alliancegenome/curation_api/interfaces/crud/AlleleDiseaseAnnotationCrudInterface.java
```diff
@@ -44,42 +44,8 @@ public interface AlleleDiseaseAnnotationCrudInterface extends BaseIdCrudInterfac
...
-
-	@POST
-	@Path("/bulk/humanAnnotationFile")
-	@JsonView(View.FieldsAndLists.class)
-	public APIResponse updateHumanAlleleDiseaseAnnotations(List<AlleleDiseaseAnnotationDTO> annotationData);
-
-	@POST
-	@Path("/bulk/sgdAnnotationFile")
-	@JsonView(View.FieldsAndLists.class)
-	public APIResponse updateSgdAlleleDiseaseAnnotations(List<AlleleDiseaseAnnotationDTO> annotationData);
```
oblodgett, line 84: @markquintontulloch thanks. I was hoping we could get rid of all these one day.

oblodgett (approved):

PR #886, opened 2023-02-01
rename run_es to run_opensearch and create a temporary script

cmpich:
rename run_es to run_opensearch and create a temporary script to start ES run_elasticsearch. Can be removed after all ES instances at the Alliance are upgrade to opensearch


oblodgett (changes requested):
This should go in the agr_java_software repo as that is where it would be used.

oblodgett (approved):

PR #888, opened 2023-02-01
SCRUM-2458 & SCRUM-2173 CrossReference and DataProvider updates and autopopulation

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/entities/base/GeneratedAuditedObject.java
```diff
@@ -20,9 +20,9 @@
 import lombok.ToString;
 
 @Data
-@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = false)
```
oblodgett: @markquintontulloch this effects ALL classes. Not sure we want to set `callSuper = false`?

src/main/java/org/alliancegenome/curation_api/services/DataProviderService.java
```diff
@@ -0,0 +1,97 @@
...
+	@Inject
+	CrossReferenceDAO crossReferenceDAO;
+	
+	@Override
+	@PostConstruct
+	protected void init() {
+		setSQLDao(dataProviderDAO);
+	}
+
+	public DataProvider createAffiliatedModDataProvider() {
```
oblodgett, line 47: might want this in a @Transactional?

src/test/java/org/alliancegenome/curation_api/model/entities/ReferenceTest.java
```diff
@@ -1,13 +1,13 @@
...
 
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
+import org.jetbrains.annotations.NotNull;
+import org.junit.jupiter.api.Test;
 
 class ReferenceTest {
```
oblodgett: I know this wasn't introduced in this PR but I would think this class should be somewhere else as in not under entities because the class is not an `@Entity`

adamgibs:
@markquintontulloch @oblodgett Update on he UI test. It broke because the mock data wasn't updated with the new model changes. The mock data was in this file. `src/main/cliapp/src/containers/conditionRelationPage/mockData/mockData.js`.  Basically, I just changed `crossReferences.curie` to `crossReferences.referencedCurie`.  I'll push out the change once this gets merged in.

markquintontulloch:
@adamgibs Thanks Adam - I thought I'd checked that file for that field but I must have missed it.

oblodgett:
@adamgibs just add the change to this PR

src/main/java/org/alliancegenome/curation_api/model/entities/base/GeneratedAuditedObject.java
markquintontulloch: Agreed - I've switched the change to the CrossReference class, which was what was causing Christian's Reference identity test to fail.

src/main/java/org/alliancegenome/curation_api/services/DataProviderService.java
markquintontulloch: done

src/test/java/org/alliancegenome/curation_api/model/entities/ReferenceTest.java
markquintontulloch: moved to `src/test/java/org/alliancegenome/curation_api/identity/ReferenceIdentity.java`, but feel free to suggest something more suitable

oblodgett (approved):

oblodgett:
@chris-grove this will take alpha offline for a good amount of time while the tables update.... @mluypaert this will also be the case on the next setup of deployments. (Long running Migration)

mluypaert:
> @chris-grove this will take alpha offline for a good amount of time while the tables update.... @mluypaert this will also be the case on the next setup of deployments. (Long running Migration)

You may want to wait to merge this till the java 17 upgrade managed to deploy, GH actions seemed to do have done something funky causing the deployment to fail, so I kicked it again and it seems to continue running now (image building and deployment for that is still ongoing: https://github.com/alliance-genome/agr_curation/actions/runs/4077747074).

oblodgett:
@mluypaert java 17 looks like it deployed to alpha java fine?

mluypaert:
> @mluypaert java 17 looks like it deployed to alpha java fine?

@oblodgett Ya, the rerun just completed pretty much as I was posting my previous message 🙄 😄

PR #889, opened 2023-02-02
Upgrade java to 17

oblodgett:


.github/workflows/alpha-build-and-deploy.yml
```diff
@@ -47,8 +47,8 @@ jobs:
       - name: Set up Maven Central Repository
         uses: actions/setup-java@v3
         with:
-          java-version: '11'
-          distribution: 'adopt'
+          java-version: '17'
+          distribution: 'zulu'
```
mluypaert: It seems the natural replacement for the deprecated adoptopenjdk docker images is eclipse-temurin, can we adopt this distribution (`temurin`) instead of `zulu`, or is there a specific reason to go for azul zulu?

Dockerfile
```diff
@@ -30,10 +28,8 @@ RUN if [ "${OVERWRITE_VERSION}" != "" ]; then \
 # build the api jar
 RUN mvn -T 8 clean package -Dquarkus.package.type=uber-jar -ntp
 
-
-
 ### Stage 3: build final application image
-FROM openjdk:11-jre-slim
+FROM openjdk:17.0.1-jdk-slim
```
mluypaert: As the openjdk docker repository has been deprecated (see [here](https://hub.docker.com/_/openjdk)) in favor of eclipse temurin (see [here](https://github.com/docker-library/openjdk/issues/505)), can we adopt the eclipse temurin image instead (`eclipse-temurin:17-jre-alpine` or `eclipse-temurin:17-jre-focal`)?

Dockerfile
```diff
@@ -9,10 +9,8 @@ COPY src/main/cliapp ./cliapp
 WORKDIR /agr_curation/cliapp
 RUN make all build
 
-
-
 ### Stage 2: build API (and include UI components)
-FROM maven:3.8-openjdk-11 as BUILD_API_STAGE
+FROM maven:3.8-openjdk-17 as BUILD_API_STAGE
```
mluypaert: Can we use `maven:3.8-eclipse-temurin-17` instead? (See other comments why I think eclipse temurin is preferred over openjdk)

mluypaert (changes requested):

.github/workflows/alpha-build-and-deploy.yml
oblodgett: zulu is openjdk vs temurin which is eclipse java jdk. So considering we will be using openjdk for compiling and running. I figured it was best to keep them the same.

Dockerfile
oblodgett: OH ok... then ya lets switch them.

oblodgett:
@mluypaert Done.

mluypaert (approved):

PR #895, opened 2023-02-06
SCRUM-2512 (mk2) Use literature service keyword search

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/helpers/references/ReferenceSynchronisationHelper.java
```diff
@@ -53,9 +53,9 @@ protected LiteratureReference fetchLiteratureServiceReference(String curie) {
 
 		HashMap<String, Object> searchField = new HashMap<>();
 		if (curie.startsWith("AGR")) {
-			searchField.put("curie", searchDetails);
+			searchField.put("curie.keyword", searchDetails);
 		} else {
-			searchField.put("cross_references.curie", searchDetails);
+			searchField.put("cross_references.curie.keyword", searchDetails);
```
oblodgett: @markquintontulloch why not just search both fields and save the extra code bellow? cross_references.curie won't match AGR and curie wont match mod references?

markquintontulloch: I figured search might be faster if only searching one field - or does it not make a difference?

oblodgett: no it doesn't make a difference

markquintontulloch: @oblodgett - done

oblodgett (approved):

PR #896, opened 2023-02-06
Fix for maven deploy

oblodgett:


mluypaert (approved):

mluypaert (commented):
Small question though, do we need to fix anything on the curation deployment? The last curation deployment to maven was successful (3 days ago): https://github.com/alliance-genome/agr_curation/actions/runs/4086640441/jobs/7046309245 ?

PR #897, opened 2023-02-06
SCRUM-2173 pt. 2 Get AllianceMember from authenticatedUser

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/DataProviderService.java
```diff
@@ -45,17 +45,12 @@ protected void init() {
 
 	@Transactional
 	public DataProvider createAffiliatedModDataProvider() {
-		LoggedInPerson user = loggedInPersonService.findLoggedInPersonByOktaEmail(authenticatedPerson.getOktaEmail());
-
 		String affiliatedModAbbreviation = null;
-		if (user.getOktaEmail().equals("test@alliancegenome.org")) {
-			// Needed to enable local testing without missing dataProvider error
+		AllianceMember member = authenticatedPerson.getAllianceMember();
+		if (authenticatedPerson.getAllianceMember() == null) {
```
oblodgett: Don't you want `if(member == null) {` ??

markquintontulloch: indeed - fixed

oblodgett:
@markquintontulloch would it be more convenient for the authenticatedUser to already have a "dataProvider" object attached? Then you could just set it?

oblodgett:
And to chain onto that question, if you and or chris login being from WB wouldn't the dataProvider be the same object from the DB?

markquintontulloch:
@oblodgett dataProvider wouldn't be the same object in the database as we want to allow the xref to be independently edited

oblodgett:
Sorry help me figure this out... everytime there is a call:
`uiEntity.setDataProvider(dataProviderService.createAffiliatedModDataProvider());`
Then we are going to be creating a new dataprovider and a new crossrefernce to the same thing over and over again?

So we will have hundreds of dataproviders all pointed at WB and hundreds of crossreferences also pointed at WB?

markquintontulloch:
Yep, but I can't see a way around that if we want to be able to edit the xref part of the dataProvider.  Otherwise, if you wanted to change the xref to point to something other than the homepage for any single entry then it would change for all entries for that MOD.

oblodgett:
Ok fine... I think this is going to cause us problems in the future but we can go with it for now... on a side note the `getAllianceMember` is the organization so you don't have to do another lookup of the organization as AllianceMember is_a Organization

markquintontulloch:
@oblodgett - refactored to avoid unnecessary lookup

oblodgett (approved):

PR #898, opened 2023-02-06
Default staff added to list of mods

kthorat-prog:


src/main/cliapp/src/service/TableStateService.js
```diff
@@ -155,13 +155,33 @@ const modTableSettings = {
...
 	}
 }
 
+const modStaffList = ['RGDStaff','SGDStaff','WBStaff','FBStaff','ZFINStaff','MGIStaff'];
+
 export function getModTableState(table){
 	const oktaToken = JSON.parse(localStorage.getItem('okta-token-storage'));
 	const mod = oktaToken?.accessToken?.claims?.Groups?.filter(group => group.includes("Staff"));
-	return global.structuredClone(modTableSettings[mod][table]);
+	return global.structuredClone(modTableSettings[modStaffList.includes(mod) ? mod : 'DefaultStaff'][table]);
```
oblodgett: @kthorat-prog I would suggest looking for modTableSettings[mod] ? modTableSettings[mod] : modTableSettings['Default'] this way there is no hardcoded list.

oblodgett: Also if another option needs to be added to the modTableSettings someone doesn't have to know that it also needs to be added to this hardcoded list.

kthorat-prog: @oblodgett Yes...thats correct. I have removed the hardcoded list.

oblodgett (approved):

PR #905, opened 2023-02-08
SCRUM-2458 update to fix ontology load

markquintontulloch:


oblodgett:
It was unclear to me... but is the crossreference_id on the GenomicEnitity? If so it should be a @ManyToOne

markquintontulloch:
crossreference_id is on genomicentity_crossreference.  We're not reusing xrefs between genomic entities, so one genomic entity can have many xrefs, but not vice-versa.

oblodgett:
So with CrossReference and GenomicEnitity if you put a 

@OneToMany on the GenomicEnitity means its going to put a genomeicentity_id on the CrossReference
@OneToMany on the CrossReference means its going to put a crossrefence_id on the GenomicEnitity
@ManyToMany on either CrossReference or GenomicEnitity will create a join table with both genomeicentity_id and crossrefence_id
So this really is a ManyToMany relationship

If you have a 
GenomicEnitity with a @OneToMany its going to try to join to the CrossReference table via a genomeicentity_id on the crossReference table

Unless you create a Object that represents the relationship GenomicEnitityCrossReference then you can have a @OneToMany on both sides and on the GenomicEnitityCrossReference you would have a @ManyToOne for both the CrossReference and GenomicEnitity

markquintontulloch:
That's not what hibernate has done though.  The schema of the database is unchanged.

I recall we've discussed this before, where you define the @OneToMany without the corresponding backreference annotation then Hibernate models it as a @ManyToMany.

I guess that's a problem though - as the schema isn't enforcing a one-to-many relationship there is the potential for many-to-many relationships to end up in the database (even though the loading code shouldn't do that) and for this problem to arise again.

Okay, back to the drawing board.

oblodgett:
Once you switch a ManyToMany back to a OneToMany and try to do a query... the backend will fail because it will try to join to the other table not use join table that was created via the ManyToMany.

There are 4 basic relationships in hibernate

1. `@OneToMany` -> `List<Object2>` the FK is on the Object2 table and typically needs to have a "mapped_by" which points to the FK on the other table so the joins work, but not a join table.
2. `@ManyToOne` -> Object1 the FK is on this table and the other table doesn't need to know anything about this object (VocabularyTerms)
3. `@ManyToMany` both the FK's for both objects are in the join table and this can be put on either side or both sides of the relationship typically one side will have a mapped_by otherwise you get 2 join tables. Table1_Table2 and Table2_Table1 

4.
Table1
`@OneToMany(mapped_by=Table1_id)` -> `List<Table1Table2>`

Table2
`@OneToMany(mapped_by=Table2_id)` -> `List<Table1Table2>`

Table1_Table2 - but this is a real object in java Table1Table2 class
`@ManyToOne` -> Table1
`@ManyToOne` -> Table2

This is really a ManyToMany relationship but you have the ability to have attributes on the join table, this also gives the option for table 2 lets say to not care about the attributes on the join table and could be defined like this:

Table2
@JoinTable(name = "Table1Table2")
@ManyToMany(targetEntity = Table1) -> `List<Table1>`

markquintontulloch:
@oblodgett - so where do we go from here?  The easy option would seem to be to leave the relationship as @ManyToMany and rely on the loading code to keep the relationships to one-to-many in reality.

Perhaps the more correct option would be to reimplement as @OneToMany with mapped_by, but that would require a significant refactor of the code and seems like a headache!

markquintontulloch:
have gone with reverting to @ManyToMany for now

src/main/resources/db/migration/v0.16.0.5__agr_curation_api.sql
```diff
@@ -0,0 +1,33 @@
...
+		SELECT curie FROM informationcontententity
+			WHERE obsolete = true
+	);
+	
+DELETE FROM informationcontententity
+	WHERE obsolete = true;
+	
+DELETE FROM ontologyterm_crossreference;
+	
+DELETE FROM genomicentity_crossreference;
```
oblodgett, line 24: @markquintontulloch I am guessing you want to delete from vs drop so that they can be reloaded?

oblodgett (approved):

oblodgett:
@markquintontulloch yes that's one of the drawbacks to connecting to an object (via hibernate) that is used by several other objects. VocabularyTerm also would be like this if you needed a list on some other object, but I think this should be fine, the list is still managed the same. One question that I have is if we are creating crossreferences that need to get deleted specifically in the case of OntologyTerm I think you might want to leave that DELETE Cascade.

src/main/resources/db/migration/v0.16.0.5__agr_curation_api.sql
markquintontulloch: exactly - we're not currently loading xrefs for genomic entities (haven't been requested from DQMs yet, although some are submitting them) but there are loads that got loaded way back before we refactored all the loading and validation stuff

PR #906, opened 2023-02-09
SCRUM-2550 Duplicate key fix

markquintontulloch:


markquintontulloch:
The name fields (displayText / formatText) are not sufficient to determine identity for the nameSlotAnnotations attached to any single gene/allele for FB annotations.

src/main/java/org/alliancegenome/curation_api/services/validation/dto/GeneDTOValidator.java
```diff
@@ -55,6 +57,8 @@ public class GeneDTOValidator extends BaseDTOValidator {
 	GeneSystematicNameSlotAnnotationDTOValidator geneSystematicNameDtoValidator;
 	@Inject
 	GeneSynonymSlotAnnotationDTOValidator geneSynonymDtoValidator;
+	@Inject
+	SlotAnnotationIdentityHelper slotAnnotationIdentity;
```
oblodgett, line 61: So typically we don't inject a helper. But rather create an instance of it as in: SlotAnnotationIdentityHelper slotAnnotationIdentity = new SlotAnnotationIdentityHelper();
However looking at the code it looks like most of your methods could be static and don't need class instances, as in you could run them via `SlotAnnotationIdentityHelper.method(params);`
Just trying to be consistent in how we do things but for now I will approve it.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/services/validation/dto/GeneDTOValidator.java
markquintontulloch: noted, will address in another PR

PR #910, opened 2023-02-10
Reindex improvements

mluypaert:
See commit messages

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
+++ b/src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
@@ -146,7 +146,7 @@ public class JobScheduler {
 
 		if(reindexSchedulingEnabled){
 			Log.info("Scheduled mass reindexing initiated.");
-			bulkLoadDAO.reindexEverything(1000, 10000, 0, 4, 14400, 1);
+			bulkLoadDAO.reindexEverything(1000, 10000, 0, 4, 7200, 1);
 		}
 		else{
 			Log.info("Scheduled mass reindexing not initiated (reindex scheduling not enabled).");
--- a/src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
+++ b/src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
@@ -146,7 +146,7 @@ public class JobScheduler {
 
 		if(reindexSchedulingEnabled){
 			Log.info("Scheduled mass reindexing initiated.");
-			bulkLoadDAO.reindexEverything(1000, 10000, 0, 4, 7200, 1);
+			bulkLoadDAO.reindexEverything(1000, 10000, 0, 4, 14400, 1);
 		}
 		else{
 			Log.info("Scheduled mass reindexing not initiated (reindex scheduling not enabled).");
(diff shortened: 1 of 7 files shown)


src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
```diff
@@ -146,7 +146,7 @@ public void runMassIndexerEverything() {
 
 		if(reindexSchedulingEnabled){
 			Log.info("Scheduled mass reindexing initiated.");
-			bulkLoadDAO.reindexEverything(1000, 10000, 0, 4, 7200, 1);
+			bulkLoadDAO.reindexEverything(1000, 10000, 0, 4, 14400, 1);
```
oblodgett: Should this be 28800 like the rest?

src/main/java/org/alliancegenome/curation_api/util/DefaultProcessDisplayHandler.java
```diff
@@ -44,10 +44,13 @@ public void progressProcess(String message, String data, long startTime, long no
...
 
 		if (percent > 0) {
+			sb.append(", Mem: " + df.format(memoryPercent() * 100) + "%");
+
 			int perms = (int) (diff / percent);
 			Date end = new Date(startTime + perms);
 			String expectedDuration = ProcessDisplayHandler.getHumanReadableTimeDisplay(end.getTime() - nowTime);
-			sb.append(", Mem: " + df.format(memoryPercent() * 100) + "%, ETA: " + expectedDuration + " [" + end + "]");
+			String wallTime = ProcessDisplayHandler.getHumanReadableTimeDisplay(nowTime - startTime);
+			sb.append(", Walltime: " + wallTime + ", ETA: " + expectedDuration + " [" + end + "]");
```
oblodgett: Can we shorten Walltime to something like "WT" or "Dur" it will make the line shorter and less likely to wrap in the logs

src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
mluypaert: No, the idea would be that for the automated reindexing, we use a shorter timeout (4h here) which we deem the maximum "acceptable" duration, which would cause the automated reindexing to fail when going over that time, giving us an indication we need to create a ticket to fix it (and potentially hold a deployment until a fix is in place). The other timeouts are the defaults that go to swagger, which should be longer such that developers triggering the indexing don't need to retry several times increasing the TransactionTimeout each time (8h in my current proposal).

src/main/java/org/alliancegenome/curation_api/util/DefaultProcessDisplayHandler.java
mluypaert: I've now shortened this to `wall` instead (WT nor Dur would be clear to me what exactly is meant with that time).

src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
mluypaert: I've now set the timeout for automated reindexing back to 2h, and the default for swagger UI to 4h.

oblodgett (approved):

PR #913, opened 2023-02-13
SCRUM-2399 migration fixes (to fix v0.16.0-rc1 deployment, SCRUM-2477)

mluypaert:


oblodgett:
Changing the content of the file is not going to force it to rerun.... the version needs to be bumped.

oblodgett:
it will need to be updated to `v0.15.0.4__agr_curation_api.sql` in order for it to rerun and then it might not even rerun on alpha which is 16+

mluypaert:
> it will need to be updated to `v0.15.0.4__agr_curation_api.sql` in order for it to rerun and then it might not even rerun on alpha which is 16+

@oblodgett I'm aware that this will not rerun on alpha (as alpha moved on past this migration), but this migration is failing on beta so needs fixing there for a successful deployment. As the current beta-branch version of this migration file fails to run on the beta database, it has not been run (successfully) there yet and thus changing the content of this migration file will fix this on beta (and run it successfully) once I created a new release candidate which includes this updated migration file. I will run the necessary cleanup queries manually on alpha once merged to clean up the alpha DB as well.

oblodgett:
Ahh yes I see the error on beta... might be it good to bring the data from prod into beta and try a redeployment just to make sure this migration will work?

oblodgett (approved):

mluypaert:
> Ahh yes I see the error on beta... might be it good to bring the data from prod into beta and try a redeployment just to make sure this migration will work?

Yes, I did a prod to beta rolldown right before I initiated the v0.16.0-rc1 deployment, so the current beta database should be near identical to the prod database.

PR #917, opened 2023-02-15
SCRUM-2282: test scaffold

abecerra:
[Incomplete - WIP]
Just to get some feedback, I added an `isMounted` check in the dashboard to avoid a warning when runing the test, it does not affect its functionality


adamgibs:
@abecerra Looks good to me. One thing is that I don't think dashboardPage.test.js needs to set 'DashboardTableSettings' in local storage.

abecerra:
@adamgibs : I think now it is ready. I need to learn how to check for the mocked data

src/main/cliapp/src/containers/conditionRelationPage/__tests__/ConditionRelationTable.test.js
```diff
@@ -18,19 +19,32 @@ describe("<ConditionRelationPage />", () => {
 	});
 
 	it("Renders without crashing", async () => {
-		const result = renderWithClient(<ConditionRelationPage />);
-		expect(result);
+		var result;
+		act(() => {
```
adamgibs, line 23: These `act()` functions get rid of that warning when running the test which is awesome. Only thing I would say is change is change the `var` to `let.`

src/main/cliapp/src/containers/dashboardPage/Dashboard.js
```diff
@@ -54,31 +54,35 @@ export const Dashboard = () => {
...
 	useEffect(() => {
+		let isMounted = true;
 		const systemService = new SystemService();
 
 		let _tableData = {};
 
 		systemService.getSiteSummary().then((res) => {
-			for (const key in res.entity){
+			if (isMounted) {
+				for (const key in res.entity){
```
adamgibs, line 64: It's odd that this is needed. I would think wrapping the `render()` functions with `act()` would make this not necessary. Odd,  but works for now and doesn't affect functionality.   ¯\_(ツ)_/¯

adamgibs (approved):

src/main/cliapp/src/containers/conditionRelationPage/__tests__/ConditionRelationTable.test.js
abecerra: I'll change them in next pr

src/main/cliapp/src/containers/dashboardPage/Dashboard.js
abecerra: Maybe the isMounted can be removed now, I'll test it

PR #927, opened 2023-02-22
Release v0.17.0 rc1

abecerra:


mluypaert (changes requested):
@abecerra Please recreate this branch to start from the commit we presented at last weeks' sprint review (2f28c62) rather than the current alpha branch, we do not want to release changes made this sprint just yet.

PR #940, opened 2023-02-27
Update OpenSearch test container start timeout

mluypaert:
The PR-validation/verify-test-api gh-actions job has been failing quite a lot recently due to timeout at startup, meaning startup takes longer than the anticipated 5 minutes. I've now increased the timeout to 10 mins, which should hopefully resolve the frequent need to rerun the verify-test-api job (sometime several times) before getting a successful run and be able to merge PRs.

### DIFF
--- a/src/test/java/org/alliancegenome/curation_api/resources/OpenSearchContainer.java
+++ b/src/test/java/org/alliancegenome/curation_api/resources/OpenSearchContainer.java
@@ -28,7 +28,7 @@ public class OpenSearchContainer extends GenericContainer<OpenSearchContainer> {
 		setWaitStrategy((new HttpWaitStrategy())
 			.forPort(9200)
 			.forStatusCodeMatching(response -> response == 200 || response == 401)
-			.withStartupTimeout(Duration.ofSeconds(600))
+			.withStartupTimeout(Duration.ofSeconds(900))
 		);
 
 	}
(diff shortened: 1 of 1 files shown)


oblodgett:
@mluypaert doesn't GH limit us to 15 minute execution time?

mluypaert:
> @mluypaert doesn't GH limit us to 15 minute execution time?

Ya, I believe you could be right in that used to be the limit, though it seems like they've increase that limit if it was:
> Job execution time - Each job in a workflow can run for up to 6 hours of execution time. If a job reaches this limit, the job is terminated and fails to complete.
https://docs.github.com/en/actions/learn-github-actions/usage-limits-billing-and-administration

oblodgett:
Ahh ok... something else we may want to look into in the future. Is that we can create our own GH "runners" they run on our own servers and those don't have any limit. Think like a mix of GH and GoCD agents.

oblodgett (approved):

mluypaert:
> Ahh ok... something else we may want to look into in the future. Is that we can create our own GH "runners" they run on our own servers and those don't have any limit. Think like a mix of GH and GoCD agents.

Ya, I know we can, but as far as I'm aware we haven't hit the limits of the GH managed workers yet, and I'd definitely want to avoid needing to manage our own workers if at all possible (otherwise we're just paying for yet more constantly running mostly idle servers, which also require more time from us to manage).

PR #943, opened 2023-02-28
SCRUM-2446 WB Phenotype ontology loading

mluypaert:


oblodgett (approved):

markquintontulloch (changes requested):
Can we make the case of the various class names consistent - you have Wbphenotype for DAO and controller, but WbPhenotype for interface.  Also is WBPheno an acceptable abbreviation?  If so, we should probably use this abbreviation for the entity class for consistency (i.e. WBPhenoTerm).

mluypaert:
> Can we make the case of the various class names consistent - you have Wbphenotype for DAO and controller, but WbPhenotype for interface. Also is WBPheno an acceptable abbreviation? If so, we should probably use this abbreviation for the entity class for consistency (i.e. WBPhenoTerm).

I've now unified the casing of all WB Phenotype Term classes, the decision to abbriate all classnames or not is up to @chris-grove I think. The abbreviated form `WBPheno` makes more sense on the dashboard and the data type selection, as that benefits from shorter names.

markquintontulloch (approved):

PR #944, opened 2023-02-28
SCRUM-2526

adamgibs:


src/main/cliapp/src/containers/affectedGenomicModelPage/AffectedGenomicModelTable.js
```diff
@@ -46,30 +47,30 @@ export const AffectedGenomicModelTable = () => {
 			header: "Curie",
 			sortable: isEnabled,
 			filter: true,
-			filterElement: {type: "input", filterName: "curieFilter", fields: ["curie"]},
+			filterElement: {type: "input", filterName: "curieFilter", fields: FILTER_FIELDS.curieFilter},
```
oblodgett, line 50: @adamgibs so... you should be able to lookup the filter via name and not have to pass `fields` in all these methods.

oblodgett (commented):
Also this does not take into account any of the auto complete filters. This is where the tricky part is: `src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js`

src/main/cliapp/src/containers/affectedGenomicModelPage/AffectedGenomicModelTable.js
adamgibs: @oblodgett Sure, refactoring the filter elements and centralizing code for autocomplete makes sense. However, this was just for the local storage filter fields validation bug. I just switched the fields so that the validation logic and the filtering logic were pointing to the same map.

PR #957, opened 2023-03-09
SCRUM-2509

adamgibs:


src/main/cliapp/src/components/Filters/FilterComponentMultiSelect.js
```diff
@@ -57,6 +57,18 @@ export function FilterComponentMultiSelect({ isEnabled, filterConfig, currentFil
 		return (<div>{option.optionLabel}</div>);
 	};
 
+	const dataProviderItemTemplate = (option) => {
```
oblodgett: Instead of calling this dataproviderTemplate can we call it uppercase template?

adamgibs: @oblodgett  Name has been updated

oblodgett (approved):
This still seems like it should be its own component: "multiselectUppercase" but whatever.

PR #962, opened 2023-03-13
Added code to split out cleanup and history tracking

oblodgett:


src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmExecutor.java
```diff
@@ -57,17 +58,44 @@ public void runLoad(BulkLoadFile bulkLoadFile) {
 			String speciesName = manual.getDataType().getSpeciesName();
 			String dataType = manual.getDataType().name();
 
-			if (agms == null)
-				agms = new ArrayList<>();
-			
-			bulkLoadFile.setRecordCount(agms.size() + bulkLoadFile.getRecordCount());
-			bulkLoadFileDAO.merge(bulkLoadFile);
+			if(agms != null) {
```
markquintontulloch: This is going to undo the changes put in to fix SCRUM-2696.  We need to create an empty list if `agms` is null so that the load runs with zero entries if a file is submitted without a particular ingest set.  The same applies for all the other executors

markquintontulloch (changes requested):

src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmExecutor.java
oblodgett: Why does it need to run without any entries?

markquintontulloch: If you run a DISEASE_ANNOTATION or FULL_INGEST load (only the latter applies to AGMs) then you would expect cleanup to occur for any missing ingest sets.  In SCRUM-2696 we saw that with disease annotations as MGI had been erroneously submitting gene disease annotations - they stopped doing so but the old gene disease annotations were not cleaned up as the  GENE_DiSEASE_ANNOTATION load did not run with an empty set, it was just skipped.

oblodgett: But its not the load that needs the empty set... its the cleanup that needs the empty set?

src/main/java/org/alliancegenome/curation_api/jobs/executors/GeneExecutor.java
```diff
@@ -57,18 +58,45 @@ public void runLoad(BulkLoadFile bulkLoadFile) {
...
+			
+			log.debug("runLoad: Before: total " + geneCuriesBefore.size());
+			
+			if (genes != null) {
+				bulkLoadFile.setRecordCount(genes.size() + bulkLoadFile.getRecordCount());
+				bulkLoadFileDAO.merge(bulkLoadFile);
+
+				APIResponse resp = runLoad(speciesName, genes, dataType, geneCuriesLoaded);
+				trackHistory(resp, bulkLoadFile);
+			}
```
oblodgett: @markquintontulloch is this the correct logic?

markquintontulloch: I think that would work, although the mechanism I discussed with Chris set the record count as 0 and reported that on the data loads page.  That would then also be consistent with the behaviour of the disease annotation executors.  So here you would have
```
if (genes == null)
    genes = new ArrayList<>();
```
and then all the other code would run regardless.

markquintontulloch: Talking of disease annotation executors, doesn't this cleanup tracking also need to be applied to disease annotations?

oblodgett: Yes I would guess, on the annotations.

oblodgett: So then I would propose that we track the deleted records on the BulkLoadFileHistory. We get a blank file we delete everything and put a field on there to show it. If that part of the load "fails" then the load fails.

markquintontulloch: I guess that would be a nice addition. My main concern here though is that we are not undoing what has been implemented already for SCRUM-2696 without a different solution in place.

I had thought this PR was supposed to be a complete solution to SCRUM-2455, hence my comment about applying to disease annotations.  I see now though that the reporting part is yet to be addressed though.

oblodgett: Ya... so I am working on a more complete fix as in tracking that number of items deleted and if any fail capturing the failures of "deletes" also. As in more incorporating the deletes into the load.

markquintontulloch: Cool - so presumably for empty ingest sets you would have something like "Records completed - 0; Records failed - 0; Records deleted - 100".  Keeping the logic of running the load with an empty list seems like an easy way of achieving the records completed/failed part, at least for now.

oblodgett: Ya... and I am also adding a Records Failed Deleting also which will capture the exception vs print to the logs and store it as an exception like the rest. Then once we get the exception download working you should have all the info to fix the file, or in our case fix the foreign key issues. (Feature) I think it would also be nice to have "category's" of exceptions and them be grouped. So you could see the different kinds of errors instead of having to scroll through everything.. but anyway.

markquintontulloch (approved):

PR #966, opened 2023-03-14
SCRUM-2711 Add DataProvider to BiologicalEntity

markquintontulloch:
A lot of this will get reverted when we implement the LinkML changes to require DQMs to submit data provider for biological entities.


src/main/java/org/alliancegenome/curation_api/services/base/BaseDTOCrudService.java
```diff
@@ -9,7 +9,8 @@
 
 	protected abstract void init();
 
-	public abstract E upsert(T dto) throws ObjectUpdateException;
+	public abstract E upsert(T dto, String dataType) throws ObjectUpdateException;
+
```
oblodgett, line 13: @markquintontulloch so what would be better is to create two methods one with the params and one without. That way you don't have to change as much code, and don't have to pass nulls all over the place. As in:
```java
public E upsert(T dto) throws ObjectUpdateException {
     return upsert(dto, null);
}
```

src/main/java/org/alliancegenome/curation_api/services/DataProviderService.java
```diff
@@ -54,15 +54,19 @@ public DataProvider createAffiliatedModDataProvider() {
 	}
 	
 	@Transactional
-	public DataProvider createAllianceDataProvider() {
-		SearchResponse<Organization> orgResponse = organizationDAO.findByField("abbreviation", "Alliance");
+	public DataProvider createOrganizationDataProvider(String dataType) {
+		SearchResponse<Organization> orgResponse = organizationDAO.findByField("abbreviation", dataType);
```
oblodgett: Also instead of calling it `dataType` can we call it `organization`?

oblodgett: Plus couldn't this be set on the Biological Entity at load time as in the processor, on the Biological Entity DTO and then it doesn't have to be passed around everywhere?

src/main/java/org/alliancegenome/curation_api/services/base/BaseDTOCrudService.java
markquintontulloch: Wouldn't that require implementing both methods in all service classes that extend this?  This will be reverted (along with all the null passing) as soon as we implement the LinkML changes in the Java code and DQMs are supplying the data provider.

oblodgett: Thats why I took off the abstract... I realize all this code is going to get ripped out thats why I am suggesting changes that change the least amount of code.

src/main/java/org/alliancegenome/curation_api/services/DataProviderService.java
markquintontulloch: Good point - will change

markquintontulloch:
@oblodgett - Rewritten to add the data provider to the incoming DTO objects in the executors.  Now there's very little code to be ripped out once data provider loading is fully implemented.

oblodgett:
@markquintontulloch ya this look a lot better

oblodgett (approved):

PR #967, opened 2023-03-15
Added error messages on restarted server

oblodgett:


src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
```diff
@@ -69,6 +70,7 @@ public void init() {
 						bulkLoadDAO.merge(b);
 					}
 					if (b.getBulkloadStatus().isRunning()) {
+						b.setErrorMessage("Failed due to server start up: Process never finshed before the server restarted");
```
markquintontulloch: Typo on finished (also above)

markquintontulloch (approved):

src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
oblodgett: oops

PR #969, opened 2023-03-16
added migration PR into SCRUM-2661

oblodgett:


oblodgett:
@markquintontulloch I have tried this and its "relatively" fast

PR #970, opened 2023-03-16
Added another option (this disables FK's which also speeds it up) PR into PR #969

oblodgett:


oblodgett:
@markquintontulloch I have not tried this but I think it will work.

PR #972, opened 2023-03-17
Evidence code filter changed to multiselect dropdown

kthorat-prog:


oblodgett:
@kthorat-prog were we going to switch to a vocabularySet? if not this looks fine

kthorat-prog:
> vocabularySet

Like not from aggregationfields?

oblodgett:
Sorry... this is what Christ said on March 9 `Options in select list will represent ALL values in entire table regardless of applied filters` so you are right. Agg query.

oblodgett (approved):

PR #973, opened 2023-03-20
SCRUM-2727 serialize referenceID of Reference object

cmpich:


oblodgett:
@cmpich this code is not going out for 5.4.0

cmpich:
Can we fast-track this change to prod?

PR #976, opened 2023-03-21
Superscript display improved for name & symbol field for Genes Table

kthorat-prog:


src/main/cliapp/src/components/EllipsisTableCell.js
```diff
@@ -2,8 +2,7 @@ import React from 'react';
 
 export const EllipsisTableCell = ({ children, otherClasses = '' }) => {
 	return (
-		<div className={`overflow-hidden text-overflow-ellipsis ${otherClasses}`} >
-			{children}
+		<div className={`overflow-hidden text-overflow-ellipsis ${otherClasses}`} dangerouslySetInnerHTML={{__html: children}}>
```
oblodgett: @kthorat-prog I think this might cause us problems in the future doing at this level.

src/main/cliapp/src/containers/genesPage/GenesTable.js
```diff
@@ -27,6 +27,18 @@ export const GenesTable = () => {
 		}
 	};
 
+	const symbolBodyTemplate = (rowData) => {
+		if (rowData?.geneFullName) {
+			return (
+				<>
+					<EllipsisTableCell otherClasses={`a${rowData.curie.replace(':', '')}`}>
+						{rowData.geneFullName.displayText}
```
oblodgett: I think it would be better to do it at this level?

kthorat-prog: But then we will have to do it everywhere like in all templates.

oblodgett: But its only for Allele Symbols and Gene names?

kthorat-prog: The ticket is for Gene symbol & name columns. Where else will this change needs to be added? I change it at individual table level as suggested by you.

kthorat-prog: Alleles, DA table is already taken care of . So, I have changed it in Genes page only.

oblodgett (approved):

src/main/cliapp/src/containers/genesPage/GenesTable.js
oblodgett: There is a reason why its called: "dangerouslySetInnerHTML" we want to make sure and use only when we know what the text is thats got HTML in it. Which I think is only in 4 places, AGM.name, Allele.symbol, Gene.name, and Gene.symbol.

PR #978, opened 2023-03-23
SCRUM-2679 Enable loading of non-canonical biological entities

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmDiseaseAnnotationExecutor.java
```diff
@@ -75,11 +77,26 @@ public void runLoad(BulkLoadFile bulkLoadFile) {
...
+	public APIResponse runLoad(String dataProvider, List<AGMDiseaseAnnotationDTO> annotations) {
+
+		List<Long> annotationIdsLoaded = new ArrayList<>();
+		List<Long> annotationIdsBefore = new ArrayList<>();
+		annotationIdsBefore.addAll(agmDiseaseAnnotationDAO.findAllAnnotationIdsByDataProvider(dataProvider));
+		annotationIdsBefore.removeIf(Objects::isNull);
+		
+		BulkLoadFileHistory history = new BulkLoadFileHistory(annotations.size());
+		runLoad(history, dataProvider, annotations, annotationIdsLoaded);
+		runCleanup(diseaseAnnotationService, history, dataProvider, annotationIdsBefore, annotationIdsLoaded);
```
oblodgett: @markquintontulloch the only draw back to this is if someone calls the API with a small set of Annotations then clean runs and deletes everything else. This was why I didn't have this runLoad do a cleanup.

oblodgett: I kinda made the assumption that the API was only for bulk updates and not "loading"

markquintontulloch: yeah, I was in two minds about this - maybe one for discussion with the group, but this is how it was previously

oblodgett: And again these API were only suppose to serve the Integration tests, but I think that @cmpich ZFIN might be using them to do updates. And before they did not run the cleanup.

oblodgett: `runLoad(history, speciesNames, alleles, null, null);` when the list is null no cleanup happens.

markquintontulloch: I thought they did do cleanups (before you refactored the cleanup stuff) - that's certainly how they were working for the integration tests before those got refactored

oblodgett: Ya... I guess when the cleanup code went in it did it on both.

markquintontulloch:
Just realised I need to add more changes to this to sort out disease annotation curies

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmDiseaseAnnotationExecutor.java
markquintontulloch: @oblodgett - cleanup now removed for direct API calls

PR #979, opened 2023-03-23
Remove datatype

oblodgett:
@markquintontulloch this is what I was trying to suggest the other day.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkManualLoad.java
+++ b/src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkManualLoad.java
@@ -5,7 +5,7 @@ import javax.persistence.EnumType;
 import javax.persistence.Enumerated;
 
 import org.alliancegenome.curation_api.constants.LinkMLSchemaConstants;
-import org.alliancegenome.curation_api.enums.BackendBulkDataType;
+import org.alliancegenome.curation_api.enums.BackendBulkDataProvider;
 import org.alliancegenome.curation_api.interfaces.AGRCurationSchemaVersion;
 import org.alliancegenome.curation_api.view.View;
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
@@ -30,5 +30,5 @@ public class BulkManualLoad extends BulkLoad {
 
 	@JsonView({ View.FieldsOnly.class })
 	@Enumerated(EnumType.STRING)
-	private BackendBulkDataType dataType;
+	private BackendBulkDataProvider dataProvider;
 }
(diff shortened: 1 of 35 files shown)


src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkManualLoad.java
```diff
@@ -30,5 +30,5 @@ public class BulkManualLoad extends BulkLoad {
 
 	@JsonView({ View.FieldsOnly.class })
 	@Enumerated(EnumType.STRING)
-	private BackendBulkDataType dataType;
+	private BackendBulkDataProvider dataProvider;
```
oblodgett, line 33: @markquintontulloch also I think this should be switched out for the object in the database.

markquintontulloch: Not sure I follow

markquintontulloch (approved):

src/main/java/org/alliancegenome/curation_api/model/entities/bulkloads/BulkManualLoad.java
markquintontulloch: @oblodgett Do you mean the dataProvider object? - I don't think that would work as we don't reuse dataProvider objects, so there isn't a single object for RGD, for example

oblodgett: Rather than being an ENUM stored as a string, should it be an DataProvider entity object?

oblodgett: Ok what about an Organization then?

markquintontulloch: I'm inclined to leave it as an enum for now as I'm not sure how that would play with the UI - we don't want all organizations to be in any dropdowns.  Talking of which, aren't these changes going to break the UI?

oblodgett: This may I will need to check, but everything else shouldn't.

oblodgett: Yes just updated the UI

markquintontulloch: I think we need a change in DataLoadsComponent.js too.  Also, does the change from "HUMAN" to "OMIM" require a database migration? It will need a change to the enum as that's currently HUMAN.

oblodgett: yes you are right... just added it.

markquintontulloch: cool, I think that covers everything - I'm happy to merge if you are

oblodgett: yes I think so

PR #980, opened 2023-03-23
SCRUM-2741 Delete MGI gene annotations

markquintontulloch:


oblodgett:
@markquintontulloch can this migration get run twice without error?

oblodgett:
That way we can run it in production manually and then when this rolls through it won't be an issue?

markquintontulloch:
Yes, will run twice fine

oblodgett (approved):

PR #983, opened 2023-03-28
SCRUM-2761

markquintontulloch:


src/main/cliapp/src/containers/dataLoadsPage/DataLoadsComponent.js
```diff
@@ -317,7 +317,7 @@ export const DataLoadsComponent = () => {
 			}
 		}
 		if (showManualLoad) {
-			ret.push(<Column key="fmsDataType2" field="fmsDataType" header="Load Data Type" />);
```
oblodgett, line 320: The field name for the FMS load did not change, not should it. See `BulkFMSLoad.java`

src/main/java/org/alliancegenome/curation_api/jobs/executors/LoadFileExecutor.java
```diff
@@ -131,22 +131,23 @@ protected boolean validSchemaVersion(String submittedSchemaVersion, Class<?> dto
 	
 
 	// The following methods are for bulk validation
-	public void runCleanup(DiseaseAnnotationService service, BulkLoadFileHistory history, String speciesName, List<Long> annotationIdsBefore, List<Long> annotationIdsAfter) {
```
oblodgett, line 134: Thanks for catching this I thought I got all of them.

src/main/java/org/alliancegenome/curation_api/jobs/executors/LoadFileExecutor.java
```diff
@@ -172,7 +173,7 @@ public void runCleanup(DiseaseAnnotationService service, BulkLoadFileHistory his
 		ph.startProcess("Deletion/deprecation of primary objects " + dataProvider, curiesToRemove.size());
 		for (String curie : curiesToRemove) {
 			try {
-				service.removeOrDeprecateNonUpdated(curie, dataProvider);
```
oblodgett, line 175: @markquintontulloch what may be better here is to pass the history object down, vs the md5 sum. And this method doesn't need to have the md5 passed in as it is available in the history object:  history.bulkLoadFile.md5Sum if there isn't a bulkLoadFile in the case of the API call with a list, then it can just be ignored.

oblodgett (changes requested):

src/main/cliapp/src/containers/dataLoadsPage/DataLoadsComponent.js
markquintontulloch: This is for the manual loads and the data type was missing from the column until I made this change

src/main/java/org/alliancegenome/curation_api/jobs/executors/LoadFileExecutor.java
markquintontulloch: I tried that initially and it was throwing errors.  I’m not sure the back reference from history to file has been created at this point as I was getting a null pointer error on history.getBulkLoadFile()

markquintontulloch:
@oblodgett  - do you require any changes?  see comments above in response to your change requests

oblodgett (approved):

PR #994, opened 2023-04-26
Fixed migration to use indexes

oblodgett:


markquintontulloch (approved):

mluypaert:
@oblodgett This branch either needs to be called `prodfix/...` or `release/v0.17.1` (I'd vote the latter as you'll probably need to create release `v0.17.1`) for it to pass the PR checks

PR #995, opened 2023-04-26
SCRUM-2450 load PATO into database

cmpich:


src/main/java/org/alliancegenome/curation_api/model/entities/ontology/PATOTerm.java
```diff
@@ -0,0 +1,23 @@
...
+import javax.persistence.InheritanceType;
+
+@Audited
+@Entity
+@Data
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
+@ToString(callSuper = true)
+@Inheritance(strategy = InheritanceType.JOINED)
+@AGRCurationSchemaVersion(min = LinkMLSchemaConstants.MIN_ONTOLOGY_RELEASE, max = LinkMLSchemaConstants.MAX_ONTOLOGY_RELEASE, dependencies = { OntologyTerm.class })
+public class PATOTerm extends OntologyTerm {
```
oblodgett, line 21: Shouldn't this have a different parent class?

oblodgett:
@cmpich also can we name the branches the same as they are in jira as in `SCRUM` vs `scrum`

src/main/java/org/alliancegenome/curation_api/model/entities/ontology/PATOTerm.java
cmpich: I am not sure what you mean? I think PATOTerm should extend OntologyTerm, I don't know of any parent class in between PATOTerm and OntologyTerm.

src/main/java/org/alliancegenome/curation_api/controllers/crud/ontology/PatoTermCrudController.java
```diff
@@ -0,0 +1,28 @@
...
+public class PatoTermCrudController extends BaseOntologyTermController<PatoTermService, PATOTerm, PatoTermDAO> implements PatoTermCrudInterface {
+
+	@Inject
+	PatoTermService patoTermService;
+
+	@Override
+	@PostConstruct
+	public void init() {
+		GenericOntologyLoadConfig config = new GenericOntologyLoadConfig();
+		config.setLoadOnlyIRIPrefix("FBbt");
```
oblodgett: Is this correct that the terms start with FBbt? This can be left out if there is a default namespace

src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadJobExecutor.java
```diff
@@ -187,107 +164,84 @@ public void process(BulkLoadFile bulkLoadFile) throws Exception {
...
 					config.setLoadOnlyIRIPrefix("OBI");
 					processTerms(bulkLoadFile, obiTermService, config);
-					break;
-				case WBPheno:
-					processTerms(bulkLoadFile, wbPhenotypeTermService, config);
-					break;
-				default:
+				}
+				case WBPheno -> processTerms(bulkLoadFile, wbPhenotypeTermService, config);
+				case PATO -> processTerms(bulkLoadFile, patoTermService, config);
```
oblodgett, line 240: This config needs to match the config in PatoTermCrudController

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/controllers/crud/ontology/PatoTermCrudController.java
```diff
@@ -0,0 +1,28 @@
...
+public class PatoTermCrudController extends BaseOntologyTermController<PatoTermService, PATOTerm, PatoTermDAO> implements PatoTermCrudInterface {
+
+	@Inject
+	PatoTermService patoTermService;
+
+	@Override
+	@PostConstruct
+	public void init() {
+		GenericOntologyLoadConfig config = new GenericOntologyLoadConfig();
+		config.setLoadOnlyIRIPrefix("PATO");
```
oblodgett: If there is a default_namespace you don't want to set the prefix.

src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadJobExecutor.java
oblodgett: If you do set a prefix it needs to be set here also.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadJobExecutor.java
cmpich: Ok, removed the prefix as there is a default namespace 'quality' in the file.

oblodgett (approved):

PR #996, opened 2023-04-26
Fixed migration to use indexes

oblodgett:


markquintontulloch (approved):

oblodgett:
@markquintontulloch you can merge this and create a release as needed.

markquintontulloch:
Thanks @oblodgett

oblodgett:
@markquintontulloch by the way the first chunk of queries took 12 minutes on my machine and the second chunk took 16 minutes so I didn't run everything but I think that will be faster.

PR #998, opened 2023-04-27
Code for quarkus upgrade 3.2

oblodgett:


oblodgett:
@mluypaert we need better checking than `github.event.action != 'edited'` these actions need to get rerun but now they can't be.

markquintontulloch (approved):

mluypaert:
> @mluypaert we need better checking than `github.event.action != 'edited'` these actions need to get rerun but now they can't be.

@oblodgett Why did the actions needed to be rerun? If you need a specific workflow to be rerun then you can always manually trigger a rerun by going to the workflow page (in Github) and hitting the "Re-run all jobs" button. Not getting rerun on title changes is exactly the intended behavior of `github.event.action != 'edited'`

oblodgett:
The reruns were not triggering.

PR #1003, opened 2023-05-02
SCRUM-2449 adding HP ontology

cmpich:


oblodgett (approved):

markquintontulloch:
@cmpich - your migration file is named `v0.18.0.5__hptermsql` as opposed to `v0.18.0.5__hpterm.sql` so the migration isn't running

PR #1008, opened 2023-05-05
SCRUM-2449 added HP Ontology

adamgibs:


oblodgett (commented):
@adamgibs was setLoadOnlyIRIPrefix preferable to default_namespace?

src/main/resources/db/migration/v0.18.0.5__hpterm.sql
```diff
@@ -13,9 +13,9 @@ CREATE TABLE public.hpterm_aud (
 ALTER TABLE ONLY public.hpterm
     ADD CONSTRAINT hpterm_pkey PRIMARY KEY (curie);
 ALTER TABLE ONLY public.hpterm
-    ADD CONSTRAINT hpterm_curie_fk FOREIGN KEY (curie) REFERENCES public.phenotypeterm(curie);
```
oblodgett: Was this an existing migration? If so you need to create a new migration for this change.

adamgibs: Ah, I see. Will update. @cmpich suggested `setLoadOnlyIRIPrefix` and it seemed to work.

oblodgett: It might... but its better to use the default namespace... if there is one.

oblodgett: On this migration it will fail because hpterm_curie_fk already exists in the database. So you have to remove it via a migration and then add it back. In other words:
`DROP CONSTRAINT hpterm_curie_fk`
`ADD CONSTRAINT hpterm_curie_fk FOREIGN KEY (curie) REFERENCES public.ontologyterm(curie);`

adamgibs: Actually, maybe the sql file on alpha is fine and the Java class needs to change. Maybe HPTerm needs to extend PhenotypeTerm instead of OntologyTerm?

oblodgett: Sure thats fine... but once a migration has been applied. We can't edit it, because it will not be rerun. So a new migration has to be created to do this change.

adamgibs: That makes sense. I think potentially my SQL change just doesn't need to happen. I'll play around with it on my machine.

oblodgett: Does HPTerm extend PhenotypeTerm? If so then this migration doesn't need to happen. But if not then the HP terms need to get deleted from the phenoterm ontology table and the this migration needs to happen.

oblodgett: I see:
```java
public class HPTerm extends OntologyTerm {

}
```
So this migration does need to happen along with a cleanup from the PhototypeTerm table.

markquintontulloch: I think this was resolved in Daily Scrum, but just to make sure:

In the LinkML model:
```
  HPTerm:
    is_a: PhenotypeTerm
```
So, it's the HPTerm entity class that needs to change to inherit from PhenotypeTerm, rather than the migration changing.

adamgibs: Thanks, Mark!

markquintontulloch (approved):

PR #1011, opened 2023-05-05
PRmerge/production

markquintontulloch:
Merge various v17 hotfixes into beta


oblodgett (commented):
Im a little concerned about the changes in the migrations? Is this expected?

markquintontulloch:
They were from the previous hotfixes, so I think they are expected in this.  They change the files to how they should be, and won't be rerun anyway.

oblodgett:
Ahh ok.

oblodgett (approved):

PR #1020, opened 2023-05-09
SCRUM-2838 updated row index for Autocomplete editors

adamgibs:


oblodgett:
@adamgibs this looks like the same code as before?

adamgibs:
@oblodgett Same change added in more places

oblodgett (approved):

PR #1027, opened 2023-05-16
SCRUM-2800 Loading of gene secondary IDs and display of gene slot annotation fields

markquintontulloch:


adamgibs (approved):
JS parts look good. @oblodgett will probably need to look at the rest

oblodgett (commented):
@adamgibs the java side looks good.

oblodgett (approved):

PR #1034, opened 2023-05-18
SCRUM-2898: agr_curation beta to prod release/v0.18.0

abecerra:


markquintontulloch (changes requested):
You need to add release notes

PR #1035, opened 2023-05-18
SCRUM-2898: agr_curation beta to prod release/v0.18.0

abecerra:


mluypaert (approved):
@abecerra Looks good now, just note that you don't have to close a PR when a negative review is made, unless you need to change the branch you want to merge from or to. Just push to the branch for which you opened a PR to merge and it will automatically show up in the open PR.

PR #1036, opened 2023-05-19
SCRUM-2341 (update) Increase number of terms returned by search for CV dropdown

markquintontulloch:


src/main/cliapp/src/service/useControlledVocabularyService.js
```diff
@@ -28,7 +28,7 @@ export function useControlledVocabularyService(termType) {
 
 	useQuery(['terms', termType],
 		() => {
-			return searchService.find("vocabularyterm", 15, 0, {"vocabulary.name" : termType } )
+			return searchService.find("vocabularyterm", 16, 0, {"vocabulary.name" : termType } )
```
oblodgett: @markquintontulloch I am curious about this change? This is the pull down for all vocabs? Would 20 or 30 be better?

markquintontulloch: @oblodgett Allele notes have 16 terms.  I asked Chris if we wanted the dropdown to return all 16 or replace with an autocomplete - he went for increasing the number of returned terms.

oblodgett: Well this is for ALL vocabs... why not 20 or 30?

markquintontulloch: because maybe that's too many? For other vocabularies with lots of terms we have used autosuggest instead.  If we want to increase further for future vocabularies then we can do that at the time.

oblodgett: if Gene notes needs 17 and Allele notes need 16. In other words changing it for Allele's is changing it for ALL, and I would guess he doesn't know that nuance?

markquintontulloch: How about I bump it to 20 then?  If a particular number of terms is acceptable for the dropdown for one vocabulary then I'm sure it would be acceptable for all.  And if it's not then we are not forced to use the dropdown and would use the autocomplete anyway.

oblodgett: In other words why not just set it to 30, and if there is a pulldown somewhere that has to many, then a new story to switch it out for an autocomplete.

oblodgett: Maybe this is just gnat soup, but I could see the next vocab that comes along that has 17 items and we are making changes to this again.

markquintontulloch: Doesn't the search service also get used for autosuggest though?  I don't think we want to return 30 terms in the autosuggest.  That's why I went with 16 - the smallest possible change in functionality to accommodate the current use case.

oblodgett: I guess my point is that so long as Chris knows this is effecting ALL, then great lets make the change. Otherwise can we choose an acceptable amount that will work for ALL and then not have to change this again.

markquintontulloch: I've asked Chris on the ticket - will wait for his response

markquintontulloch: @oblodgett - bumped to 20 on Chris' request after explaining the global nature of this change

abecerra (approved):

oblodgett (approved):

PR #1044, opened 2023-05-25
Fix for indexes and constraints

oblodgett:


src/main/java/org/alliancegenome/curation_api/model/entities/Allele.java
```diff
@@ -74,57 +74,49 @@ public class Allele extends GenomicEntity {
 	@OneToMany(mappedBy = "subject", cascade = CascadeType.ALL)
 	private List<AlleleDiseaseAnnotation> alleleDiseaseAnnotations;
 
-	@IndexedEmbedded(includePaths = { "mutationTypes.curie", "mutationTypes.name", "evidence.curie",
-			"mutationTypes.curie_keyword", "mutationTypes.name_keyword", "evidence.curie_keyword"})
+	@IndexedEmbedded(includePaths = { "mutationTypes.curie", "mutationTypes.name", "evidence.curie", "mutationTypes.curie_keyword", "mutationTypes.name_keyword", "evidence.curie_keyword"})
```
cmpich, line 77: why not leaving two lines? One line is too long.

oblodgett: to me its easier to read when not wrapped... plus I have a big montior

cmpich: We have the coding convention: no lines longer than 120 characters.

cmpich (approved):
Looks good. We need to agree on a coding style for eclipse and IntelliJ so we are not going to overwrite each other's changes.

PR #1063, opened 2023-06-06
SCRUM-2883: Added database user and privileges instructions to README

mluypaert:


oblodgett (approved):

oblodgett:
@mluypaert with the `no-deploy` label can we get the actions to skip the tests also?

mluypaert:
> @mluypaert with the `no-deploy` label can we get the actions to skip the tests also?

We could, but we should probably use a different label (like `no-test` or something) if we want to accomplish that, to ensure we can create PRs that need to be able to build, but which we simply don't want to deploy at current due to other priorities.

oblodgett:
Well changes to a README don't need to be tested. (To me the no-deploy and no-test always have gone hand in hand)

mluypaert:
> > @mluypaert with the `no-deploy` label can we get the actions to skip the tests also?
> 
> We could, but we should probably use a different label (like `no-test` or something) if we want to accomplish that, to ensure we can create PRs that need to be able to build, but which we simply don't want to deploy at current due to other priorities.

Also, we should be careful in considering whether that's really needed as a functionality, because that potentially opens the gates for untested changes or changes failing testing to get merged anyway (and then cause problems further down the line). Implementing this possibility would move the responsibility of safe merges strongly to the reviewer (whom currently often do a semi-blind approval).

PR #1064, opened 2023-06-06
SCRUM-2917: left aligning the column contents

abecerra:


oblodgett (changes requested):
Not sure where my comment went but it would be better to use bootstrap classes to do this kind of thing vs hardcoding css styles: https://primeflex.org/position

oblodgett (changes requested):
PS Also this code has not been tested locally. He is requesting the synonyms column.

oblodgett:
https://www.primefaces.org/primeflex/alignitems
Plus the styleClass is going to go on the `ListTableCell`

abecerra:
> https://www.primefaces.org/primeflex/alignitems Plus the styleClass is going to go on the `ListTableCell`

How about making the style a parameter, with left alignment as default,  and allowing to center or right align with a style argument? [last commit]

src/main/cliapp/src/components/ListTableCell.js
```diff
@@ -1,13 +1,16 @@
...
+	const leftAlign = { textAlign: 'left' };
+	let alignment = style ?? leftAlign;
 	let suppress = 'list-none'
 	if(showBullets)
 		suppress = ''
 	return (
 		<ul className={`pl-0 ${suppress}`}>
 			{listData?.map((item, index) =>
-				<li key={index}>
+				<li key={index} style={alignment}>
```
oblodgett: We don't need this code... we should no hardcode styles.

oblodgett (changes requested):

src/main/cliapp/src/components/ListTableCell.js
oblodgett: Plus this is going to change ALL ListTableCell usage... the code needs to be on the ListTableCell component not in here, styling that is.

oblodgett: This is a single line change:
```<ListTableCell template={listTemplate} listData={sortedSynonyms}/>```
change to:
 ```<ListTableCell styleClass="align-items-start" template={listTemplate} listData={sortedSynonyms}/>```

oblodgett:
This is a single line change:
```<ListTableCell template={listTemplate} listData={sortedSynonyms}/>```
change to:
 ```<ListTableCell styleClass="align-items-start" template={listTemplate} listData={sortedSynonyms}/>```

abecerra:
> This is a single line change: `<ListTableCell template={listTemplate} listData={sortedSynonyms}/>` change to: `<ListTableCell styleClass="align-items-start" template={listTemplate} listData={sortedSynonyms}/>`

The styleClass="align-items-start" in the ListTableCell left-aligns the list,

> This is a single line change: `<ListTableCell template={listTemplate} listData={sortedSynonyms}/>` change to: `<ListTableCell styleClass="align-items-start" template={listTemplate} listData={sortedSynonyms}/>`

The change produces the result in the attached capture:
Screenshot_20230612_122250

abecerra:
Using styleClass="align-items-start" in the  <ul> inside the ListTableCell, or even in each <li> does not change the result. The only way I was able to align the texts at the left was with { textAlign: 'left' }

oblodgett:
Sorry was wrong on which one to put the class:
On the list template:
`className='overflow-hidden text-overflow-ellipsis text-left'`

abecerra:
> Sorry was wrong on which one to put the class: On the list template: `className='overflow-hidden text-overflow-ellipsis text-left'`


it works now!

oblodgett (approved):

PR #1065, opened 2023-06-06
Fix issue with header column

oblodgett:


oblodgett:
@markquintontulloch @adamgibs in this case the column name change from "Genetic Modifier" to "Genetic Modifiers" is not just a simple column name change, but rather we have to think about the settings that are in each persons browser for this column.

markquintontulloch (approved):

adamgibs:
@oblodgett I think one potential issue with this solution is that if the column isn't found due to a name change, the column won't be displayed until the table state is changed. So in this case Genetic Modifiers would be unselected not because that was user's intended or saved state, but because the column name was changed.

oblodgett:
@adamgibs thats better than the table crashing.

adamgibs:
@oblodgett Sure, but we may need another ticket to address the issue further at some point. On the other hand, I don't know how much column names and fields will be changing once production is in use and stable so it may not much of an issue at that point anyway.

PR #1072, opened 2023-06-07
PRmerge/beta

adamgibs:


oblodgett:
@adamgibs I would guess that this PR should have changes?

adamgibs:
@oblodgett No, it's just the change in utils that pushed into alpha yesterday for the column header stuff. So the change is already in alpha

oblodgett:
I would guess that we don't need this PR then.

adamgibs:
@oblodgett Per the README, "After prerelease creation and deployment, merge the beta branch back to alpha, to make the (pre)release tag reachable from alpha (and report the correct version number through git describe --tags)". Is that relevant here?

oblodgett:
ya ok

oblodgett (approved):

PR #1077, opened 2023-06-09
Attempt to get better performance

oblodgett:
Not sure if the r5.large -> m5.2xlarge stays on the same architecture... I would assume so.

### DIFF
--- a/.ebextensions/aws-autoscaling.config
+++ b/.ebextensions/aws-autoscaling.config
@@ -1,6 +1,6 @@
 option_settings:
   aws:ec2:instances:
-    InstanceTypes: m5.2xlarge
+    InstanceTypes: m6i.2xlarge
   aws:autoscaling:asg:
     Availability Zones: Any
     MaxSize: 1
--- a/.ebextensions/aws-autoscaling.config
+++ b/.ebextensions/aws-autoscaling.config
@@ -1,6 +1,6 @@
 option_settings:
   aws:ec2:instances:
-    InstanceTypes: r5.large
+    InstanceTypes: m5.2xlarge
   aws:autoscaling:asg:
     Availability Zones: Any
     MaxSize: 1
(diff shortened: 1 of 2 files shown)


mluypaert (changes requested):
Those do use the same architecture, but if you want to improve performance then you may want to upgrade to the current intel instance families (`r6i.*` and `m6i.*`). Also, with 32 GiB mem at instance level, you may want to keep at least 10% for system (so assign at most 29 to the application).

oblodgett:
Typically for single application type server the recommendation is for 1GB for the system and the rest for the app, in this case leaving 2gb for the system should be fine.

oblodgett:
That is odd they are the same price but the network is better on the m6i.

mluypaert:
> That is odd they are the same price but the network is better on the m6i.

New generations always get either better performance or cheaper prices (or both).

.ebextensions/aws-autoscaling.config
```diff
@@ -1,6 +1,6 @@
 option_settings:
   aws:ec2:instances:
-    InstanceTypes: r5.large
+    InstanceTypes: m6i.2xlarge
```
mluypaert, line 3: @oblodgett `r5.large` to `m6i.2xlarge` means going from 16 GiB and 2vCPU to 32 GiB and and 8vCPU. Wouldn't you want to upgrade one or the other, rather than both (and definitely not going from 2vCPU to 8 while we don't event see a bottleneck at vCPU level on the app server)?

oblodgett: Im looking for both more memory and more cpu... there is something in this chain that is a bottle neck and right now its not easy to identify so I would like to open up the pipe to see the better performance. And then start dropping pieces of it to understand what is the effecting thing.

mluypaert (approved):

.ebextensions/aws-autoscaling.config
oblodgett: Right now ES is on a 16 node cluster, which is massive compared to what it was on. The database and its IO have been doubled, and yet no change. So it has to be the app server, and my guess is that we might have several rounds of sizing to get the right instance type.

mluypaert: Alright, if upsizing vCPU and RAM on the app server doesn't work, then you might want to look at trying to improve network and disk. `m6id.2xlarge` would be the equivalent disk optimised instance, `m6in.2xlarge` the network optimised one and `m6idn.2xlarge` both network and disk optimised. We may also need to check if we need to change any system configurations to make use of that improved network and/or disk (not sure).

PR #1079, opened 2023-06-09
SCRUM-2983 added indexedEmbedded include depth to gene and allele DAs

adamgibs:
subjects


markquintontulloch (approved):

oblodgett (commented):
Do we want a depth of 2 or just 1?

markquintontulloch:
We need 2 due to the slot annotation fields

oblodgett:
Ahh ok

oblodgett (approved):

PR #1080, opened 2023-06-12
SCRUM-2946 - Improve error reporting for failed data loads

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/jobs/executors/GeneExecutor.java
```diff
@@ -72,6 +72,7 @@ public void runLoad(BulkLoadFile bulkLoadFile, Boolean cleanUp) {
 			trackHistory(history, bulkLoadFile);
 
 		} catch (Exception e) {
+			failLoad(bulkLoadFile, e.getMessage());
```
oblodgett: `e.getMessage()` might not be the best message to put as the error message. There are a few others that could be added. There are several others:
```
			e.getLocalizedMessage();
			e.getCause().getMessage();
```
These being added to the error message might be more informative.

markquintontulloch: @oblodgett - went with `e.getLocalizedMessage()` as `e.getCause()` returns null for the exception in question.  Also changed the code to pass the Exception itself rather than just the message so that we only have to change this in one place if we decide something else is more appropriate.

src/main/java/org/alliancegenome/curation_api/jobs/executors/LoadFileExecutor.java
```diff
@@ -190,4 +189,10 @@ public void runCleanup(DiseaseAnnotationService service, BulkLoadFileHistory his
 		ph.finishProcess();
 		
 	}
+	
+	protected void failLoad(BulkLoadFile bulkLoadFile, Exception e) {
+		bulkLoadFile.setErrorMessage(e.getLocalizedMessage());
```
oblodgett: I guess my point is that you will want to add these messages together. As in e.getMessage + e.getLocalizedMessage() + e.getCause().getMessage() doing null checks in case. This will give the best message otherwise someone is going to have to look in the logs anyway.

oblodgett: There is also a "RootCasuse" somewhere that give an even better message... Something like this:
```
Throwable rootCause = e;
while (rootCause.getCause() != null && rootCause.getCause() != rootCause) {
     rootCause = rootCause.getCause();
}
rootCause.getMessage() + rootCause.getLocalizedMessage().;
```
but you will also want e.getMessage() and e.getLocalizedMessage() if they exist.

markquintontulloch: Ah, ok - done.  Used a Set to avoid duplication as in the test case e.getMessage() and e.getLocalizedMessage() return the same string.

src/main/java/org/alliancegenome/curation_api/jobs/executors/LoadFileExecutor.java
```diff
@@ -190,4 +191,15 @@ public void runCleanup(DiseaseAnnotationService service, BulkLoadFileHistory his
 		ph.finishProcess();
 		
 	}
+	
+	protected void failLoad(BulkLoadFile bulkLoadFile, Exception e) {
+		Set<String> errorMessages = new HashSet<String>();
+		errorMessages.add(e.getMessage());
+		errorMessages.add(e.getLocalizedMessage());
+		if (e.getCause() != null)
+			errorMessages.add(e.getCause().getMessage());
```
oblodgett: Not sure if you will need to do the while loop to get the "rootCause" but we can go with this.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/jobs/executors/LoadFileExecutor.java
oblodgett: PS... good idea with the set.

markquintontulloch: Added the while loop and also changed the Set type to LinkedHashSet in order to preserve order of insertion

oblodgett: Nice.

PR #1083, opened 2023-06-12
SCRUM-2991 update - Update with field index and remove filtering on undisplayed fields

markquintontulloch:


markquintontulloch:
@oblodgett - will hold off on merging this to allow allele indexing to run

oblodgett (approved):

PR #1091, opened 2023-06-14
SCRUM-2600: Upgrade to Node 18

abecerra:
- Adding babel/plugin-proposal-private-property-in-object silences a long warning in the UI tests, maybe it can be removed later in a ticket to upgrade/remove dependencies

### DIFF
--- a/src/main/cliapp/package.json
+++ b/src/main/cliapp/package.json
@@ -9,6 +9,7 @@
     },
     "license": "MIT",
     "dependencies": {
+	"@babel/plugin-proposal-private-property-in-object": "7.21.0-placeholder-for-preset-env.2",
         "@fullcalendar/core": "^5.7.2",
         "@fullcalendar/daygrid": "^5.7.2",
         "@fullcalendar/interaction": "^5.7.2",
--- a/src/main/cliapp/package.json
+++ b/src/main/cliapp/package.json
@@ -9,7 +9,7 @@
     },
     "license": "MIT",
     "dependencies": {
-        "@babel/plugin-proposal-private-property-in-object": "7.21.0-placeholder-for-preset-env.2",
+        "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
         "@fullcalendar/core": "^5.7.2",
         "@fullcalendar/daygrid": "^5.7.2",
         "@fullcalendar/interaction": "^5.7.2",
@@ -62,7 +62,7 @@
         "not op_mini all"
     ],
     "devDependencies": {
-        "@babel/plugin-proposal-private-property-in-object": "7.21.0-placeholder-for-preset-env.2",
+        "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
         "msw": "^0.48.1"
     }
 }
--- a/src/main/cliapp/package.json
+++ b/src/main/cliapp/package.json
@@ -9,7 +9,7 @@
     },
     "license": "MIT",
     "dependencies": {
-	"@babel/plugin-proposal-private-property-in-object": "7.21.0-placeholder-for-preset-env.2",
+        "@babel/plugin-proposal-private-property-in-object": "7.21.0-placeholder-for-preset-env.2",
         "@fullcalendar/core": "^5.7.2",
         "@fullcalendar/daygrid": "^5.7.2",
         "@fullcalendar/interaction": "^5.7.2",
@@ -62,6 +62,7 @@
         "not op_mini all"
     ],
     "devDependencies": {
+        "@babel/plugin-proposal-private-property-in-object": "7.21.0-placeholder-for-preset-env.2",
         "msw": "^0.48.1"
     }
 }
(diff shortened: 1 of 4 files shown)


oblodgett (dismissed):
This does not compile locally:

```
Compiled with problems:
×
ERROR in ./src/components/GenericDataTable/GenericDataTable.js
Module build failed (from ./node_modules/babel-loader/lib/index.js):
Error: [BABEL] /Users/olinblodgett/git/agr_curation/src/main/cliapp/src/components/GenericDataTable/GenericDataTable.js: --- PLACEHOLDER PACKAGE ---
This @babel/plugin-proposal-private-property-in-object version is not meant to
be imported. Something is importing
@babel/plugin-proposal-private-property-in-object without declaring it in its
dependencies (or devDependencies) in the package.json file.
Add "@babel/plugin-proposal-private-property-in-object" to your devDependencies
to work around this error. This will make this message go away.
```

oblodgett:
This also disrupts all local development for the UI and this story would be better pulled in in the off release.

src/main/cliapp/package.json
```diff
@@ -9,6 +9,7 @@
     },
     "license": "MIT",
     "dependencies": {
+	"@babel/plugin-proposal-private-property-in-object": "7.21.0-placeholder-for-preset-env.2",
```
oblodgett: Plus the indention is wrong with this.

mluypaert: To add in to this, this version reference looks like a development version, which should not be used for importing in our project (use stable versions only). The current version on npm is `7.21.11` so I would suggest using that.

src/main/cliapp/package.json
```diff
@@ -9,6 +9,7 @@
     },
     "license": "MIT",
     "dependencies": {
+        "@babel/plugin-proposal-private-property-in-object": "7.21.0-placeholder-for-preset-env.2",
```
mluypaert: `babel/plugin-proposal-private-property-in-object` must be imported at a production-grade version (I'd propose `^7.21.11`), instead of the development version used here.

mluypaert (changes requested):

src/main/cliapp/package.json
```diff
@@ -61,6 +62,7 @@
         "not op_mini all"
     ],
     "devDependencies": {
+        "@babel/plugin-proposal-private-property-in-object": "7.21.0-placeholder-for-preset-env.2",
```
mluypaert: Same here, production version please.

mluypaert (changes requested):

src/main/cliapp/package.json
abecerra: set in last commit

src/main/cliapp/package.json
abecerra: set in last commit

src/main/cliapp/package.json
abecerra: set in last commit

mluypaert:
> This does not compile locally:
> 
> ```
> Compiled with problems:
> ×
> ERROR in ./src/components/GenericDataTable/GenericDataTable.js
> Module build failed (from ./node_modules/babel-loader/lib/index.js):
> Error: [BABEL] /Users/olinblodgett/git/agr_curation/src/main/cliapp/src/components/GenericDataTable/GenericDataTable.js: --- PLACEHOLDER PACKAGE ---
> This @babel/plugin-proposal-private-property-in-object version is not meant to
> be imported. Something is importing
> @babel/plugin-proposal-private-property-in-object without declaring it in its
> dependencies (or devDependencies) in the package.json file.
> Add "@babel/plugin-proposal-private-property-in-object" to your devDependencies
> to work around this error. This will make this message go away.
> ```

@oblodgett What version of Node are you using? Seems to compile fine on my local system using Node `v18.14.1`

oblodgett:
Yes it probably works on the correct node version but this update now requires all developers to stop and update node. Again this needs to wait till the off release sprint.

abecerra:
> This also disrupts all local development for the UI and this story would be better pulled in in the off release.

To merge it needs in local environments:

- delete package-lock.json node_modules
- nvm use v18.16.0 
- npm install

So it could be merged tomorrow morning (UK)

oblodgett:
No... the team did not agree to pull in this story.

mluypaert:
I disagree that this can only get merge during off-release (updating your node version for this project really is a trivial task if you're using `nvm`), but I agree that this tickets need approval by the team before merge.

oblodgett:
@markquintontulloch and @adamgibs are the primary UI developers, they are the ones that need to drive this ticket. If they are fine with it and can get the right version of node installed sure bring it in, but we did not discuss this as a team.

adamgibs:
@oblodgett @abecerra @mluypaert @markquintontulloch This seems to work for me locally. Component test seem to run as well.

markquintontulloch:
I'm fine with pulling it in, but agree that the timing of breaking changes to the development environment should really be discussed with the team in advance.

adamgibs (approved):

abecerra:
> I'm fine with pulling it in, but agree that the timing of breaking changes to the development environment should really be discussed with the team in advance.

Yes, but the PR helps testing the github actions changes. Once approved the merging can wait discussion of course.

mluypaert (approved):

PR #1094, opened 2023-06-16
SCRUM-2338 Load and display allele_database_status attribute

markquintontulloch:


adamgibs (approved):
UI stuff looks good

src/main/java/org/alliancegenome/curation_api/model/entities/slotAnnotations/alleleSlotAnnotations/AlleleDatabaseStatusSlotAnnotation.java
```diff
@@ -0,0 +1,56 @@
...
+
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@Audited
+@Entity
+@Data
+@EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
+@Inheritance(strategy = InheritanceType.JOINED)
```
oblodgett: This (`@Inheritance(strategy = InheritanceType.JOINED)`) only needs to be on here if there is a inheriting subclass.

oblodgett (changes requested):
Still have some other checks to do.

src/main/resources/db/migration/v0.21.0.1__add_allele_database_status.sql
```diff
@@ -0,0 +1,30 @@
...
+
+ALTER TABLE alleledatabasestatusslotannotation_aud ADD CONSTRAINT alleledatabasestatus_aud_id_rev_fk FOREIGN KEY (id, rev) REFERENCES slotannotation_aud (id, rev);
+
+INSERT INTO vocabulary (id, name) VALUES (nextval('hibernate_sequence'), 'Allele database status vocabulary');
+	
+INSERT INTO vocabularyterm (id, name, vocabulary_id) SELECT nextval('hibernate_sequence'), 'approved', id FROM vocabulary WHERE name = 'Allele database status vocabulary';
+INSERT INTO vocabularyterm (id, name, vocabulary_id) SELECT nextval('hibernate_sequence'), 'deleted', id FROM vocabulary WHERE name = 'Allele database status vocabulary';
+INSERT INTO vocabularyterm (id, name, vocabulary_id) SELECT nextval('hibernate_sequence'), 'reserved', id FROM vocabulary WHERE name = 'Allele database status vocabulary';
+INSERT INTO vocabularyterm (id, name, vocabulary_id) SELECT nextval('hibernate_sequence'), 'autoloaded', id FROM vocabulary WHERE name = 'Allele database status vocabulary';
+INSERT INTO vocabularyterm (id, name, vocabulary_id) SELECT nextval('hibernate_sequence'), 'in_progress', id FROM vocabulary WHERE name = 'Allele database status vocabulary';
```
oblodgett, line 31: Also looks like there are two missing FK's:

```
ALTER TABLE public.alleledatabasestatusslotannotation_aud
    ADD CONSTRAINT fko47ewqwil249eqgi82tcc3wxn FOREIGN KEY (id, rev) REFERENCES public.nameslotannotation_aud(id, rev);
```

and

```
ALTER TABLE public.alleledatabasestatusslotannotation
    ADD CONSTRAINT fkbjl9t9iwmxl7jgxuwpi8r3atc FOREIGN KEY (id) REFERENCES public.nameslotannotation(id);
```
   
Always need to test against what hibernate is going to do.

oblodgett (changes requested):

src/main/resources/db/migration/v0.21.0.1__add_allele_database_status.sql
markquintontulloch: Actually, it was the class definition that was wrong - it shouldn't be extending NameSlotAnnotation, just SlotAnnotation

src/main/java/org/alliancegenome/curation_api/model/entities/slotAnnotations/alleleSlotAnnotations/AlleleDatabaseStatusSlotAnnotation.java
markquintontulloch: Will remove from this and other allele/gene slot annotation classes

src/main/resources/db/migration/v0.21.0.1__add_allele_database_status.sql
markquintontulloch: Was missing an FK constraint on the id field to the slotAnnotation table though - fixed now

src/main/java/org/alliancegenome/curation_api/model/entities/slotAnnotations/alleleSlotAnnotations/AlleleDatabaseStatusSlotAnnotation.java
markquintontulloch: Removed

src/main/resources/db/migration/v0.21.0.1__add_allele_database_status.sql
oblodgett: Ahh ok.

oblodgett (approved):

oblodgett:
PS... I have found some duplicate indexing on some of the tables on alpha. This would cause slower insert times, if you have a few maybe you could go through some of the slot annotation tables and create a migration to remove some of the extra (duplicate) indexes. But I would suggest that in a separate PR.

markquintontulloch:
Sure, will take a look at them in another PR

PR #1098, opened 2023-06-19
SCRUM-2448 - Load DPO Ontology

markquintontulloch:
@oblodgett - standard ontology stuff, but note the change to the FileTransferHelper in order to deal with redirects and enable use of the PURL

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -358,6 +358,12 @@
 			<artifactId>reflections</artifactId>
 			<version>0.10.2</version>
 		</dependency>
+		
+		<dependency>
+    		<groupId>org.jsoup</groupId>
+    		<artifactId>jsoup</artifactId>
+    		<version>1.13.1</version>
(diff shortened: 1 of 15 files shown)


oblodgett:
@markquintontulloch the only issue is that it only handles 1 redirect i have seen at least one of the purls that has multiple, i think it would be better to find a library that does this than code it ourselves.

oblodgett:
[stackoverflow.com/questions/13193595/java-get-filename-of-download-from-redirected-friendly-url](https://stackoverflow.com/questions/13193595/java-get-filename-of-download-from-redirected-friendly-url)
If you look at the Jsoup example, it has an option like followRedirects(true)

markquintontulloch:
@oblodgett - Switched to Jsoup for URL redirection

oblodgett (approved):

PR #1099, opened 2023-06-19
Correct migration version

markquintontulloch:
Had wrong version number on migration, so previous PR failed to deploy


oblodgett:
Will this cause alpha to rerun it?

markquintontulloch:
It hasn't run on alpha as the incorrect version was prior to the latest on alpha

oblodgett (approved):

PR #1103, opened 2023-06-19
Added code to remove many elastic search fields

oblodgett:


markquintontulloch (approved):

oblodgett:
@markquintontulloch @adamgibs not sure if this might cause some errors in the UI... but we will go with these fields for now... may need to add more depending on which fields we are searching.

PR #1108, opened 2023-06-20
SCRUM-2828 Prevent loading of duplicate notes

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/helpers/notes/NoteIdentityHelper.java
```diff
@@ -0,0 +1,33 @@
...
+import java.util.Collections;
+import java.util.List;
+
+import javax.enterprise.context.RequestScoped;
+
+import org.alliancegenome.curation_api.model.ingest.dto.NoteDTO;
+import org.apache.commons.collections.CollectionUtils;
+import org.apache.commons.lang.StringUtils;
+
+@RequestScoped
```
oblodgett: If the class is going to be statically referenced then it doesn't need the @RequestScoped, @RequestScoped  is only for `@Inject`ion

markquintontulloch: removed

oblodgett (approved):

PR #1109, opened 2023-06-21
SCRUM-3007 Add DataProvider to Gene and AGM tables

markquintontulloch:


oblodgett (approved):
May want to `no-deploy` one of these PR's so you don't get two deployments.

PR #1113, opened 2023-06-21
SCRUM-3005 Related Note dialog display updates

markquintontulloch:


src/main/cliapp/src/containers/allelesPage/RelatedNotesDialog.js
```diff
@@ -387,15 +387,16 @@ export const RelatedNotesDialog = ({
...
-					<Column editor={internalEditor} field="internal" header="Internal" body={internalTemplate} headerClassName='surface-0'/>
 					<Column
 						editor={(props) => freeTextEditor(props, "freeText", errorMessages)}
 						field="freeText"
 						header="Text"
 						body={textTemplate}
 						headerClassName='surface-0'
+						style={{overflowWrap: 'break-word', maxWidth: '35rem'}}
 					/>
+					<Column editor={(props) => evidenceEditorTemplate(props, errorMessages)} field="evidence.curie" header="Evidence" headerClassName='surface-0' body={(rowData) => evidenceTemplate(rowData)} style={{overflowWrap: 'break-word', maxWidth: '25rem'}}/>
```
oblodgett: Instead of hardcoding the styles can we find bootstrap styles that will work? @adamgibs

markquintontulloch: I couldn't find any, but could have been looking in the wrong places!

oblodgett: https://www.primefaces.org/primeflex/textoverflow

markquintontulloch: that just truncates the text (with or without elipsis).   Chris wanted wrapping of text across multiple lines where the string was too long and didn't include whitespace.

oblodgett: yes... I know there should be other options for style classes at that site.

markquintontulloch: That’s where I looked before.  The closest I could find was an option for wrapping on white space characters, but nothing for wrapping mid-word.

oblodgett: So.... if its going to be custom then it would be better to add a RelatedNotesDialog.css file with:
```
.wrap-word {
    overflowWrap: 'break-word'
}
```
And also you can add `w-25rem` to the `styleClass="wrap-word w-25rem"` this way its using the built in styles.... this also helps when it comes to resizing fonts etc.

oblodgett: Not fully sure on the file name maybe need to be <file>.scss @adamgibs ? Also will need to import it in the Dialog.

markquintontulloch: ok - where should that css file live in the repo? Edit: didn’t see your previous reply!

adamgibs: We don't really have that much custom CSS and what we do is inline so I don't think we have a precedent set with regard to CSS file location. It looks like there is a small App.scss file that Ketaki used. Perhaps it could go in there? Also, I'm not really sure why it's SCSS instead of CSS. Does need to be? It looks like plain CSS to me.

oblodgett: @adamgibs SCSS gets processed like javascript and that is the react way. Also we need to stop doing the inline css and start using the build in styles. This is going to work better change changing themes, fonts sizes, etc.

oblodgett: @markquintontulloch the App.scss file might be best then we will have those custom classes for use everywhere.

markquintontulloch: @oblodgett @adamgibs - done, I think

adamgibs (approved):
Looks good to me. As a side note, it might make sense at some point to refactor these RelatedNotesDialog components since they have so much in common. Maybe just pull out the shared functions and templates and import them in.

oblodgett (approved):

PR #1118, opened 2023-06-22
SCRUM-3008 Use consensus set of defining properties for DAs

markquintontulloch:
Includes endpoint to regenerate uniqueId for all disease annotations


oblodgett (approved):
Looks good... just wait on the indexer to finish.

PR #1128, opened 2023-06-27
SCRUM-3005 (update) Custom SCSS class reimplementation

markquintontulloch:
@adamgibs - This keeps the custom classes but avoids the weird scrolling bug


adamgibs (approved):
Cool, sounds good

PR #1130, opened 2023-06-29
SCRUM-2621_mem-monitoring

mluypaert:
Added code to enable memory (and cpu, disk and net) monitoring of the application server (the EC2 instance Elastic Beanstalk spins up) + miscelaneous changes to enable working-directory code deployment to EB (to a dedicated EB environment for testing purposes).


abecerra (approved):
Looks good, no idea about the cloudwatch.config though!

mluypaert:
> Looks good, no idea about the cloudwatch.config though!

Rest assured that this file has been tested thoroughly 🙂

Dockerfile
```diff
@@ -48,4 +50,4 @@ ENV QUARKUS_HIBERNATE_SEARCH_ORM_ELASTICSEARCH_HOSTS opensearch:9200
 ENV QUARKUS_HIBERNATE_SEARCH_ORM_ELASTICSEARCH_PROTOCOL http
 
 # Start the application
-CMD ["java", "-Xms30g", "-Xmx30g", "-jar", "agr_curation_api-runner.jar"]
+CMD ["java", "-XX:InitialRAMPercentage=50", "-XX:MaxRAMPercentage=90", "-jar", "agr_curation_api-runner.jar"]
```
oblodgett, line 53: @mluypaert whenever you read about java performance you always want to preallocate all the memory and not allow it to dynamically resize. So InitialRAMPercentage should be set to 90 also.

mluypaert: Looking at the current reported RAM usage, we're not even using the initially assigned 50% (we're currently using about 25% of the total available RAM), so let's see how much memory we actually need first and whether there's a lot of resizing happening in reality.

oblodgett: Yes but from the Oracle docs: "If the optimal heap size for the application is known, we recommend that you set -Xms and -Xmx to the same value. This gives you a controlled environment where you get a good heap size right from the start." https://docs.oracle.com/cd/E13150_01/jrockit_jvm/jrockit/geninfo/diagnos/memman.html

mluypaert: Sure, I'd argue the optimal heap size for our application is not know though, so let's set the minimum size to the same as the maximum only once we have a known optimal heap size (let the metrics collect for a bit and then look at max used in reality) and then we can also use the optimal instance for it.

PR #1131, opened 2023-06-30
SCRUM-2710 Implement loading of orthology data

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/jobs/executors/OrthologyExecutor.java
```diff
@@ -0,0 +1,138 @@
...
+	GeneToGeneOrthologyGeneratedService generatedOrthologyService;
+
+	public void runLoad(BulkLoadFile bulkLoadFile) {
+		try {
+			BulkFMSLoad fms = (BulkFMSLoad) bulkLoadFile.getBulkLoad();
+			
+			OrthologyIngestFmsDTO orthologyData = mapper.readValue(new GZIPInputStream(new FileInputStream(bulkLoadFile.getLocalFilePath())), OrthologyIngestFmsDTO.class);
+			bulkLoadFile.setRecordCount(orthologyData.getData().size());
+			
+			AGRCurationSchemaVersion version = Molecule.class.getAnnotation(AGRCurationSchemaVersion.class);
```
oblodgett: @markquintontulloch shouldn't the version be coming from the Orthology class vs the Molecule class?

src/main/java/org/alliancegenome/curation_api/model/entities/orthology/GeneToGeneOrthologyCurated.java
```diff
@@ -24,6 +25,7 @@
 import lombok.ToString;
 
 @Audited
+@Indexed
```
oblodgett: @markquintontulloch is there going to be a table to view all Gene to Gene ortho? This may cause more indexing issues. To me this would be another candidate for a gene page that has a table to show them?

src/main/java/org/alliancegenome/curation_api/jobs/executors/OrthologyExecutor.java
markquintontulloch: Yep, good spot - fixed

src/main/java/org/alliancegenome/curation_api/model/entities/orthology/GeneToGeneOrthologyCurated.java
markquintontulloch: I had assumed so - will ask Chris what he wants to do.  Would they not need to be indexed for a gene page view anyway?  There's obviously a lot of orthology pairs, but we index a pretty small number of fields so I'm not sure it would cause too many problems.

oblodgett: If we are looking at a Gene then it would be gene.orthology.fields which would be indexed off the gene object. And would be available as a list off the gene. Basically the table we are looking is a list of those objects. So if the table is to show GeneToGeneOrthology's then sure we need to index them but otherwise no. @Indexed makes a index for those objects directly. But they can be indexed off other objects as needed by doing the index path, without being indexed directly. In other words if we are going to have a table for the object or a autocomplete for the object then it needs to be @Indexed otherwise it can be included on other objects via the `@IndexedEmbedded(includePaths = {})`

markquintontulloch: @oblodgett  - UI code and @Indexed annotations removed

oblodgett (approved):

oblodgett (approved):

PR #1139, opened 2023-07-05
SCRUM-2710 (update) Use term map for validation

markquintontulloch:
@oblodgett - On my local system, loads are about 3 times faster with this method


oblodgett:
@markquintontulloch I'm thinking something a little different. This gets into the java app server internals a bit more... given that we are inside a `@RequestScope` means that the class will be around for the duration of the request (meaning all items that are in the load vs just an API request), so we could check a "requestTime" variable if null no issue just run the request as normal if its set however then we would check for the cache. If there is no cache (2nd request) then we load it. 3rd request and all the ones after it would use the cache. This prevents any passing around of lists and this would also work for ALL things (Gene load, Allele load, etc) that use Vocabulary terms over and over again. This all being done in the `VocabularyTermValidationHelper` so its seamless to whoever is using it. (We can chat about this tomorrow)

markquintontulloch:
@oblodgett - sounds complicated!  Makes sense though, and definitely good to not be passing maps around.  Will close this PR and open a new one just for the logging for now - can sort out the cache stuff for all loads in a separate ticket.

PR #1147, opened 2023-07-07
Allele load speedup

oblodgett:


markquintontulloch:
Looks great - just need to change the endpoint in the `createPerson` method in `org.agr_curation_api.src.test.java.org.alliancegenome.curation_api.base.BaseITCase.java` as it's pointing to `/loggedinperson`

markquintontulloch:
We'll need to update the LinkML model for the Person / LoggedInPerson changes too (and update the version in the @AGRCurationSchemaVersion annotation on the entity class once a release is created).  What's the reason for this change?

oblodgett:
@markquintontulloch when you look at the queries that get generated, some queries join to the person / loggedinperson 10's of times. The removal of this class causes a speedup in the queries given that EVERYTHING has a updatedbBy and createdBy fields which have to link out to both tables twice.

markquintontulloch (approved):

PR #1159, opened 2023-07-11
SCRUM-3062 Load Biological Spatial Ontology (BSPO)

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
```diff
@@ -239,7 +239,7 @@ private boolean isNodeInOntology(OWLClass currentTreeNode, T currentTerm, HashSe
 		boolean condition5 = !config.getIgnoreEntitiesWithChebiXref();
 		boolean condition6 = !hasChebiXref(currentTerm);
 
-		return ((condition1 && condition2) || (condition3 && condition4)) && (condition5 || condition6);
+		return ((condition1 && condition2 && !condition3) || (condition3 && condition4)) && (condition5 || condition6);
```
abecerra, line 242: Why is negated now?

abecerra (approved):

oblodgett:
@markquintontulloch can we get the other PR in first?

src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
oblodgett: @abecerra its not negated its added to the condition.

markquintontulloch: If we apply the IRI prefix filter we don’t want terms to load if they match the default namespace but don’t have the correct prefix. ie. Passing condition1 and condition2 is not sufficient, checkIRIPrefix must be null

markquintontulloch:
> 

Sure. I’m at my kids’ sports day so can’t review the other PR at the moment.  Will be back at my desk in about an hour.

oblodgett:
@markquintontulloch this will probably fail... you will need to switch out the imports.

oblodgett:
@markquintontulloch you will need to revert back to the javax imports and you can get this in for tomorrow.

PR #1168, opened 2023-07-12
SCRUM-3037 (cleanup) Remove temp endpoint

markquintontulloch:


oblodgett (approved):
I always like to see code getting removed.

PR #1178, opened 2023-07-24
SCRUM-2784

cmpich:
Add short citation to reference table and JSON


oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
```diff
@@ -47,6 +47,11 @@ public class Reference extends InformationContentEntity {
 	@EqualsAndHashCode.Include
 	private List<CrossReference> crossReferences;
 
+	@IndexedEmbedded(includeDepth = 1)
+	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
```
oblodgett: @cmpich this is not an object... copy and paste from a string field.

oblodgett (changes requested):
Needs to have the annotations changed.

oblodgett (approved):

PR #1180, opened 2023-08-03
SCRUM-2248 Fail loads with invalid URL with correct status

markquintontulloch:


oblodgett (changes requested):
So the issue needs to be caught before getting into that method. Also in that method we need to check for the file and that its not  zero bytes. Rather then just catching a null pointer.

oblodgett (approved):

oblodgett:
@markquintontulloch not sure if this fully fixes the issue... or what happens when there is an error... but at least there will not be any null pointers.

PR #1182, opened 2023-08-03
SCRUM-2248 update

markquintontulloch:
@oblodgett Needed to stop `compressInputFile()` from running if `filePath` is null otherwise get uncaught NullPointerException and status doesn't get changed to FAILED


oblodgett:
it should ok with a null? I was trying to achieve that.

oblodgett:
I think we need to fix the compress file method if it receives null it needs to return null.

oblodgett:
May want to close this one? as the other code will work?

PR #1184, opened 2023-08-04
SCRUM-3117 Delay bulk load FINISHED status until file load complete

markquintontulloch:


oblodgett:
So... not driving into this, i know the statuses are important on how things run. I would suggest not making this change but rather tweak the ui to display the load status at the collection of all the file statuses under it. I would need to look into this further before approving this PR.

PR #1186, opened 2023-08-08
SCRUM-3143 RGD obsoletion fix

mluypaert:


markquintontulloch:
Don't we need to also update the logic for the disease annotation loads as these currently use the dataProvider of the subject?

You'll also need to switch out OMIM for HUMAN in the UI DataLoadsService (and notify Marek that he needs to use HUMAN for submission of files rather than OMIM once this reaches production).

Personally, I'm not a huge fan of specifying canonical species taxon for RGD rat loads as there are multiple rat species that may well be submitted in the future (I'd prefer dataProvider = RGD and taxon NOT NCBITaxon:9606), but I guess this works for now.

Finally, we should really add in the validation (and integration tests) that we talked about for loads to ensure the data is what we are expecting and we don't get caught out by this again.

mluypaert:
> Don't we need to also update the logic for the disease annotation loads as these currently use the dataProvider of the subject?

Yes, I realised that after I opened the PR, working on that now.

> 
> You'll also need to switch out OMIM for HUMAN in the UI DataLoadsService (and notify Marek that he needs to use HUMAN for submission of files rather than OMIM once this reaches production).

Ah right, thanks, I didn't think of the UI part. I'll change that as well.

> 
> Personally, I'm not a huge fan of specifying canonical species taxon for RGD rat loads as there are multiple rat species that may well be submitted in the future (I'd prefer dataProvider = RGD and taxon NOT NCBITaxon:9606), but I guess this works for now.
> 
> Finally, we should really add in the validation (and integration tests) that we talked about for loads to ensure the data is what we are expecting and we don't get caught out by this again.

src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadProcessor.java
```diff
@@ -108,7 +108,7 @@ public void syncWithS3(BulkLoadFile bulkLoadFile) {
 		log.info("S3: " + bulkLoadFile.getS3Path());
 
 		if ((bulkLoadFile.getS3Path() != null || bulkLoadFile.generateS3MD5Path() != null) && bulkLoadFile.getLocalFilePath() == null) {
-			File outfile = fileHelper.downloadFileFromS3(s3AccessKey, s3SecretKey, s3Bucket, s3PathPrefix, bulkLoadFile.generateS3MD5Path());
```
oblodgett, line 111: @mluypaert was there an issue with this? s3PathPrefix should be "alpha" "beta" or "production" so that files can be all stored in the same bucket.

mluypaert: Yes there was an issue with the old code: when doing a data-rolldown from production to beta (as an example), the S3Paths also get rolled down. When a file then had been loaded to production, but not to beta, the download would fail when using the prefix, while the file is available in the production S3Path. Downloading directly from the S3Path rather than generating the path using the S3PathPrefix solved this issue (while still storing new files using the prefix).

markquintontulloch:
@mluypaert  - also need to update the logic for finding previously loaded orthology pairs (again, currently uses subject gene data provider)

src/main/java/org/alliancegenome/curation_api/jobs/BulkLoadProcessor.java
oblodgett: Ok... I just wasn't sure at what point that would be null vs it being populated. But based on if statement this is only when there is no local file as in a rerun of a previous load, which would mean the path is populated.

mluypaert: > Ok... I just wasn't sure at what point that would be null vs it being populated. But based on if statement this is only when there is no local file as in a rerun of a previous load, which would mean the path is populated.

Exactly

mluypaert:
@markquintontulloch Disease annotation classes have now been updated as well and UI has been updated so this PR should now be good to go (additional validation will be ticketed in another ticket as it is out of scope for this ticket). Orthology classes were already fixed ([here](https://github.com/alliance-genome/agr_curation/pull/1186/files#diff-3acc522d1cd7d791b5fb35de79229a9fc3d1309769fa5a70350eedd6122833b3R43) and [here](https://github.com/alliance-genome/agr_curation/pull/1186/files#diff-187fb2fa74fed987072d1c6fafc2c833bf1c4b33865b57e6577d6d2389327ca1R49)).

markquintontulloch (approved):

PR #1187, opened 2023-08-08
SCRUM-3167 Add label to Vocabulary and VocabularyTermSet

markquintontulloch:
And make vocabularies editable in UI


src/main/java/org/alliancegenome/curation_api/constants/VocabularyConstants.java
```diff
@@ -2,38 +2,38 @@
...
+	public static final String AGM_SUBTYPE_VOCABULARY = "agm_subtype";
 	
-	public static final String ORTHOLOGY_BEST_SCORE_VOCABULARY = "Orthology best score";
-	public static final String ORTHOLOGY_CONFIDENCE_VOCABULARY = "Orthology confidence";
-	public static final String ORTHOLOGY_BEST_REVERSE_SCORE_VOCABULARY_TERM_SET = "Orthology best reverse score";
-	public static final String ORTHOLOGY_PREDICTION_METHOD_VOCABULARY = "Orthology prediction methods";
+	public static final String ORTHOLOGY_BEST_SCORE_VOCABULARY = "ortho_best_score";
+	public static final String ORTHOLOGY_CONFIDENCE_VOCABULARY = "ortho_confidence";
+	public static final String ORTHOLOGY_BEST_REVERSE_SCORE_VOCABULARY_TERM_SET = "ortho_best_rev_score";
+	public static final String ORTHOLOGY_PREDICTION_METHOD_VOCABULARY = "ortho_prediction_method";
```
oblodgett, line 37: @markquintontulloch might want to change the variable names also to add _LABEL on the end?

src/main/java/org/alliancegenome/curation_api/model/entities/VocabularyTermSet.java
```diff
@@ -50,6 +52,12 @@ public class VocabularyTermSet extends GeneratedAuditedObject {
 	@KeywordField(name = "name_keyword", aggregable = Aggregable.YES, sortable = Sortable.YES, searchable = Searchable.YES, normalizer = "sortNormalizer")
 	@JsonView({ View.FieldsOnly.class })
 	private String name;
+	
+	@FullTextField(analyzer = "autocompleteAnalyzer", searchAnalyzer = "autocompleteSearchAnalyzer")
+	@KeywordField(name = "vocabularyLabel_keyword", aggregable = Aggregable.YES, sortable = Sortable.YES, searchable = Searchable.YES, normalizer = "sortNormalizer")
+	@Column(unique = true)
+	@JsonView({ View.FieldsOnly.class })
+	private String vocabularyLabel;
```
oblodgett, line 60: Should we call this termSetLabel? Rather than vocabularyLabel? Sorry should have caught this over on the linkml side.

src/main/java/org/alliancegenome/curation_api/services/VocabularyTermService.java
```diff
@@ -105,18 +105,18 @@ public ObjectResponse<VocabularyTerm> getTermInVocabularyTermSet(String vocabula
 		
 	}
 	
-	private VocabularyTerm getTermInVocabularyFromDB(String vocabularyName, String termName) {
+	private VocabularyTerm getTermInVocabularyFromDB(String vocabularyLabel, String termLabel) {
 		Map<String, Object> params = new HashMap<>();
-		params.put("name", termName);
```
oblodgett, line 110: Does term name need to be changed here?

adamgibs (approved):

oblodgett:
@markquintontulloch looks like my phone only put these comments in "pending" and I would guess you didn't see them till now?

src/main/java/org/alliancegenome/curation_api/model/entities/VocabularyTermSet.java
markquintontulloch: We could, but I'm inclined to keep it the same for both Vocabulary and VocabularyTermSet - avoids duplicated code for a field that has the same function

src/main/java/org/alliancegenome/curation_api/services/VocabularyTermService.java
markquintontulloch: yeah, `"name"` is correct but should have stuck with `termName` - will put in a new PR to fix

src/main/java/org/alliancegenome/curation_api/constants/VocabularyConstants.java
markquintontulloch: I'm inclined to keep them as they are - they didn't have _NAME previously and the vocabularyLabel field is the only constant field (plus the variable names are already pretty lengthy)

src/main/java/org/alliancegenome/curation_api/model/entities/VocabularyTermSet.java
markquintontulloch: I'd planned to just use `label` but that's already used in the Image class in LinkML and since I wanted the field to be required for Vocabulary and VocabularyTermSet it seemed clearer to define a new slot rather than using the slot_usage properties to specify required on each class.

oblodgett: What is the usage for "label" is it a string field? Maybe we could just use it?

src/main/java/org/alliancegenome/curation_api/constants/VocabularyConstants.java
oblodgett: Ya... good point.

src/main/java/org/alliancegenome/curation_api/model/entities/VocabularyTermSet.java
markquintontulloch: We could, but then we would have to define slot_usage properties on the Vocabulary and VocabularyTermSet classes in order to make them required.  Also, I wanted to keep this slot separate so that the slot description in the LinkML was clear as to what it's purpose was as opposed to the `name` slot.  If `vocabularyLabel` proves to be confusing then we can change it later, but I'd rather keep it as is for now in order to avoid another LinkML release and more code changes.

PR #1189, opened 2023-08-08
PRmerge/beta (v0.22.0-rc2, SCRUM-3143) mergeback to alpha

mluypaert:


markquintontulloch (approved):

markquintontulloch:
@mluypaert - Is this going to work?  Won't the previous migration attempt to run again and the new one not run?

mluypaert:
> @mluypaert - Is this going to work? Won't the previous migration attempt to run again and the new one not run?

Not sure, this might temporarily break migrations on alpha, but can be fixed manually if so (I will monitor the deployment to alpha). It will definitely work correctly when the database roll-down from production comes (after v0.22.0 is released and deployed to production).

PR #1194, opened 2023-08-14
SCRUM-3172  - Ensure submitted data meets expectations for cleanup code parameters

markquintontulloch:


oblodgett (commented):
@markquintontulloch my gut seems to sense that this isn't the best approach, considering there is a lot of infrastructure code that is changing (passing DP around everywhere). Also in the future this is not going to be needed as we will be doing everything through the API. And this is all just so we can know what to delete from the DB after a load? I am uneasy with this change but can't see a strong reason of why not to, so will leave it up to you.

markquintontulloch:
@oblodgett The future of everything through the API is a long way off though.  If we have expectations of the data specific to the load type then we need to validate that the data meets those expectations - we've been caught out by this several times now and wasted a lot of time on investigations and cleanup.

oblodgett (approved):

PR #1209, opened 2023-08-22
SCRUM-2784 make short citation show up next to the entry box when selected

cmpich:


oblodgett (changes requested):
The story doesn't outright say but he wants this to be a new column in the table.

cmpich:
I don't see the requirement to be a new column. I just added it underneath the IDs as requested.

PR #1223, opened 2023-09-01
SCRUM-3027: increase readibility for experimental conditions edition

abecerra:


src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js
```diff
@@ -463,7 +463,7 @@ export const NewAnnotationForm = ({
 		<div>
 			<Toast ref={toast_error} position="top-left" />
 			<Toast ref={toast_success} position="top-right" />
-			<Dialog visible={newAnnotationDialog} style={{ width: '900px' }} header={dialogHeader} modal className="p-fluid" footer={dialogFooter} onHide={hideDialog} resizeable>
+			<Dialog visible={newAnnotationDialog} style={{ width: '1200px' }} header={dialogHeader} modal className="p-fluid" footer={dialogFooter} onHide={hideDialog} maximizable resizeable="true">
```
adamgibs: Resizable is a boolean type so I think it would need to be set to like this `resizable={true}`. However, if it's true, I don't think it needs to be set equal to anything. Also, I think the e needs to be removed., "resizable" not "resizeable". Link to the docs https://www.primefaces.org/primereact-v8/dialog/

oblodgett: Also these widths need to be not hardcoded they need to be relative via bootstrap styles. 1200 for me is not the same as 1200 for you.

oblodgett (dismissed):

src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js
oblodgett: More Generally we should not be using `style=` anywhere, that is why we are using a framework. So that the framework can provide our styling for us.

abecerra: - I am going to remove the resizable (s), as the default value is true. The forms then get an icon to maximize
- I am going to remove the style and let the new disease annotation form start maximized

abecerra: resizable removed!

oblodgett (commented):
I still would like to see the style width removed and we use prime react classes. But looks good to me.

abecerra:
> I still would like to see the style width removed and we use prime react classes. But looks good to me.

It was removed in the new annotation form, now it is starting maximized to accomodate more space for experimental conditions.

oblodgett:
@adamgibs are there more changes here?

adamgibs (approved):

PR #1229, opened 2023-09-08
SCRUM-3027: no maximize new dis. anot. form and fix column expansion

abecerra:


oblodgett (approved):

oblodgett:
Nice use of the prime flex classes 👍

PR #1240, opened 2023-09-13
SCRUM-3268 Display latest FMS release as default

markquintontulloch:


adamgibs (approved):

src/main/cliapp/src/components/FMSComponent.js
```diff
@@ -10,26 +10,32 @@ export const FMSComponent = () => {
...
 	useEffect(() => {
 		const fmsService = new FMSService();
 		fmsService.getReleases().then(results => {
 			//console.log(results);
 			setReleases(results.reverse());
-			for(let idx of results) {
-				if(idx.releaseVersion === selectedRelease.releaseVersion) {
-					setSelectedRelease(idx);
+			if (selectedRelease.releaseVersion === '0') {
+				setSelectedRelease(results[0]);
```
oblodgett, line 22: Not sure that results[0] is always going to be the "current" release, it may be the future release, but we can go with this.

oblodgett (approved):

PR #1244, opened 2023-09-14
Initial Code for backend child counts

oblodgett:
There is a PR in the schemas repo to add these fields: https://github.com/alliance-genome/agr_curation_schema/pull/233


src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
```diff
@@ -196,6 +196,9 @@ public T traverse(OWLClass currentTreeNode, int depth, HashSet<String> requiredN
 				}
 			}
 		}
+		
+		currentTerm.setChildCount(currentTerm.getIsaChildren().size());
+		currentTerm.setDescendantsCount(currentTerm.getIsaDescendants().size());
```
oblodgett: These might need null checks

markquintontulloch (approved):

PR #1254, opened 2023-09-22
SCRUM-199 Load and display constructs

markquintontulloch:


adamgibs (approved):
JS looks good!

src/main/java/org/alliancegenome/curation_api/dao/ConstructDAO.java
```diff
@@ -0,0 +1,28 @@
...
+	@Inject
+	NoteDAO noteDAO;
+
+	protected ConstructDAO() {
+		super(Construct.class);
+	}
+
+	public List<Long> findAllIdsByDataProvider(String dataProvider) {
+		Query jpqlQuery = entityManager.createQuery("SELECT construct.id FROM Construct construct WHERE construct.dataProvider.sourceOrganization.abbreviation = :dataProvider");
+		jpqlQuery.setParameter("dataProvider", dataProvider);
```
oblodgett, line 24: @markquintontulloch I think we may need to create a method or pass a flag to the find() method so that we get back a list of id's rather than objects. I keep seeing an increase for this need. Then it prevents us from having to write these methods everywhere.

src/main/java/org/alliancegenome/curation_api/services/validation/ReagentValidator.java
```diff
@@ -0,0 +1,99 @@
...
+		}
+		
+		DataProvider validatedDataProvider = dpResponse.getEntity();
+		if (validatedDataProvider.getObsolete() && (dbDataProvider == null || !validatedDataProvider.getId().equals(dbDataProvider.getId()))) {
+			addMessageResponse(field, ValidationConstants.OBSOLETE_MESSAGE);
+			return null;
+		}
+		
+		if (validatedDataProvider.getId() == null)
+			validatedDataProvider = dataProviderDAO.persist(validatedDataProvider);
```
oblodgett: Ideally you would use the dataProvider service here.

src/main/resources/db/migration/v0.27.0.1__add_constructs.sql
```diff
@@ -0,0 +1,144 @@
...
+	rev integer NOT NULL,
+	revtype smallint
+);	
+	
+ALTER TABLE constructcomponentslotannotation_note_aud ADD PRIMARY KEY (constructcomponentslotannotation_id, relatednotes_id, rev);
+
+ALTER TABLE constructcomponentslotannotation_note_aud ADD CONSTRAINT constructcomponentslotannotation_note_aud_rev_fk FOREIGN KEY (rev) REFERENCES revinfo (rev);
+
+	
+INSERT INTO vocabulary (id, name, vocabularylabel) VALUES (nextval('hibernate_sequence'), 'Construct Component Note Type', 'construct_component_note_type');
```
oblodgett, line 141: After the quarkus 3.0 upgrade `nextval('hibernate_sequence')` will need to be changed to the specific sequence for the table.

oblodgett (approved):
Good job this is a complicated change.

src/main/java/org/alliancegenome/curation_api/dao/ConstructDAO.java
markquintontulloch: Agreed - will put in another PR for this at some point

src/main/java/org/alliancegenome/curation_api/services/validation/ReagentValidator.java
markquintontulloch: Refactored

src/main/resources/db/migration/v0.27.0.1__add_constructs.sql
markquintontulloch: Yup - we have a lot of these in the migrations

PR #1259, opened 2023-09-25
SCRUM-3132 (update) Deal with removed or non-selected terms

markquintontulloch:
@adamgibs Is there a better place for this logic, or an easy way to avoid having to duplicate for other non-required autocomplete fields.  This was put in place to avoid a 400 error when a term was removed from a non-required field and saved in the detail page  - the caused a conversion error on the backend as it as trying to convert and empty string to a PhenotypeTerm.  The same error was being thrown if a valid curie was typed but not selected from the autocomplete list.

### DIFF
--- a/src/main/cliapp/src/containers/allelesPage/inheritanceModes/InheritanceModesForm.js
+++ b/src/main/cliapp/src/containers/allelesPage/inheritanceModes/InheritanceModesForm.js
@@ -41,13 +41,22 @@ export const InheritanceModesForm = ({ state, dispatch }) => {
   
   const phenotypeTermOnChangeHandler = (event, setFieldValue, props) => {
     //updates value in table input box
-    setFieldValue(event.target.value);
+    let value = null;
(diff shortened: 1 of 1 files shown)


adamgibs:
@markquintontulloch For the not selecting from the autosuggest issue, we could add UI validation code like we have for subject fields that disallows submission without selecting from the dropdown. However, that might be weird for a field that is optional. In terms of avoiding duplication, you could just pull this code out into a utility function an add it to the utils file and import it. It would still need to be added to all optional fields onChange functions like this, but at least it would be a one-liner.

adamgibs:
@markquintontulloch I have the same issue in my current ticket. I can add the code to the utils file if you would rather just go ahead and merge this in. Either way.

markquintontulloch:
@adamgibs Sounds good - I'll go ahead and merge.  Thanks.

adamgibs (approved):

PR #1262, opened 2023-09-26
SCRUM-3316 update rowData directly to fix dropdown bug

adamgibs:
@markquintontulloch We're going to have to go with your solution from  earlier today for now. The DataTable component props provides an editorCallback that is used to update the editor state. The problem is, it fails in tables that use row edit and open in edit mode.  I put together a demo and asked the PrimeReact people about it this morning and they said that it's a bug in PrimeReact version 8.x but is fixed in 9.6.2. However, that upgrade will have to wait. I've added comments as a reminder for now.


markquintontulloch (approved):

oblodgett:
@adamgibs we are only on prime react version 7 right?

adamgibs:
@oblodgett We're on version 8.7.3

PR #1267, opened 2023-09-28
SCRUM-3125: remove the DELETE action for symbol in Allele details

abecerra:


src/main/cliapp/src/containers/allelesPage/symbols/SymbolForm.js
```diff
@@ -83,12 +83,10 @@ export const SymbolForm = ({ labelColumnSize, state, dispatch }) => {
     });
   }
 
-  const deletionHandler  = (e) => {
-    e.preventDefault();
-    dispatch({type: "DELETE_OBJECT", entityType: "alleleSymbol"});
-    dispatch({type: "UPDATE_TABLE_ERROR_MESSAGES", entityType: "alleleSymbol", errorMessages: []});
-  };
 
+  const button = state.allele?.alleleSymbol ? <></> :
```
markquintontulloch: There shouldn't be an 'Add Symbol' button.  It's a required field so the row is always there and cannot be added or deleted.  The detail page is not used for creating new annotations so there cannot be a case of there not being a symbol since validation would have failed.

markquintontulloch (changes requested):

src/main/cliapp/src/containers/allelesPage/symbols/SymbolForm.js
abecerra: I can remove it completely

markquintontulloch (approved):

PR #1270, opened 2023-09-29
SCRUM-3319 (LinkML v1.10.0 update) & SCRUM-3320 (Load & display constructs, 2nd iteration)

markquintontulloch:


adamgibs (approved):
JS looks good

src/main/java/org/alliancegenome/curation_api/jobs/executors/AlleleDiseaseAnnotationExecutor.java
```diff
@@ -47,6 +48,8 @@ public void runLoad(BulkLoadFile bulkLoadFile, Boolean cleanUp) {
 
 			IngestDTO ingestDto = mapper.readValue(new GZIPInputStream(new FileInputStream(bulkLoadFile.getLocalFilePath())), IngestDTO.class);
 			bulkLoadFile.setLinkMLSchemaVersion(getVersionNumber(ingestDto.getLinkMLVersion()));
+			if (StringUtils.isNotBlank(ingestDto.getAllianceMemberReleaseVersion()))
+				bulkLoadFile.setAllianceMemberReleaseVersion(ingestDto.getAllianceMemberReleaseVersion());
```
oblodgett, line 52: Wondering if we can get rid of this duplication of logic?

src/main/java/org/alliancegenome/curation_api/services/validation/slotAnnotations/constructSlotAnnotations/ConstructFullNameSlotAnnotationValidator.java
```diff
@@ -0,0 +1,70 @@
...
+		if (response.hasErrors()) {
+			if (throwError) {
+				response.setErrorMessage(errorTitle);
+				throw new ApiErrorException(response);
+			} else {
+				return null;
+			}
+		}
+
+		return dbEntity;
```
oblodgett, line 67: This logic looks very copy and pasteish? Wondering if we can consolidate these three validate methods into a single method?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/jobs/executors/AlleleDiseaseAnnotationExecutor.java
markquintontulloch: I can't think of a clean way of doing this, especially as the logic is different for FMS files.  Open to suggestions though.

src/main/java/org/alliancegenome/curation_api/services/validation/slotAnnotations/constructSlotAnnotations/ConstructFullNameSlotAnnotationValidator.java
markquintontulloch: It's maybe not as copy-and-paste as it looks.  Each name type validates against a different vocabularyTermSet (or Vocabulary rather than set in the case of Synonyms).  All the actual validation code is already elsewhere in shared methods.

PR #1275, opened 2023-10-02
Hotfix for UI tests

markquintontulloch:


mluypaert (approved):

oblodgett (commented):
@adamgibs still wondering if we shouldn't be ignoring this file.

mluypaert:
> @adamgibs still wondering if we shouldn't be ignoring this file.

Ya, me too. I'm sure there must be a better way to fix the UI-tests than including that file (and essentially freezing all our package versions), but I'm letting it pass for now until we found a better solution in order to be able to continue with deployments.

adamgibs:
@oblodgett @mluypaert From my research, best practice is to commit the package-lock.json. The idea is that all developers and automated environments are using the same versions of everything. Here's the [docs](https://docs.npmjs.com/cli/v9/configuring-npm/package-lock-js) and here's another [article](https://medium.com/helpshift-engineering/package-lock-json-the-complete-guide-2ae40175ebdd) explaining it and why `npm ci` should be used in automated environments or when you want to ensure that the dependency tree isn't updated. Also,  [here's a Stackoverflow discussion](https://stackoverflow.com/questions/71691283/is-using-npm-ci-by-developers-a-good-npm-working-process) about whether to use `npm ci` locally instead of `npm insall`. That said, if you guys come across something that I've missed that would be better, then we should consider that then.

PR #1288, opened 2023-10-10
SCRUM-2248 Load Allele-Gene associations

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/dao/associations/alleleAssociations/AlleleGeneAssociationDAO.java
```diff
@@ -0,0 +1,24 @@
...
+
+@ApplicationScoped
+public class AlleleGeneAssociationDAO extends BaseSQLDAO<AlleleGeneAssociation> {
+
+	protected AlleleGeneAssociationDAO() {
+		super(AlleleGeneAssociation.class);
+	}
+
+	public List<Long> findAllIdsByDataProvider(String dataProvider) {
+		Query jpqlQuery = entityManager.createQuery("SELECT aga.id FROM AlleleGeneAssociation aga WHERE aga.subject.dataProvider.sourceOrganization.abbreviation = :dataProvider");
```
oblodgett, line 19: As a point for the future we need to create a generic method that does this.

markquintontulloch: Yeah, I'll have a think about that.  It's slightly different for different classes, but there's a lot of duplication and will be more so once we start adding more association classes.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/dao/associations/alleleAssociations/AlleleGeneAssociationDAO.java
oblodgett: Ya we could have a method that is like the other one where you pass in the params but then you could also pass in the "id" field then all the same code could be run...

PR #1290, opened 2023-10-11
SCRUM-2248 update - AlleleGeneAssociation loading - bulk load deprecation & link to gene

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/entities/Allele.java
```diff
@@ -58,7 +58,6 @@
 @ToString(exclude = { "alleleGeneAssociations", "alleleDiseaseAnnotations", "alleleMutationTypes", "alleleSymbol", "alleleFullName", "alleleSynonyms", "alleleSecondaryIds", "alleleInheritanceModes", "alleleFunctionalImpacts", "alleleGermlineTransmissionStatus", "alleleDatabaseStatus", "alleleNomenclatureEvents" }, callSuper = true)
 @AGRCurationSchemaVersion(min = "1.7.3", max = LinkMLSchemaConstants.LATEST_RELEASE, dependencies = { GenomicEntity.class }, partial = true)
 @Table(indexes = { @Index(name = "allele_inCollection_index", columnList = "inCollection_id"), })
-@JsonIdentityInfo(generator = ObjectIdGenerators.PropertyGenerator.class, property = "curie")
```
markquintontulloch, line 61: This annotation shouldn't have been kept in and is causing the error on the disease annotation table at the moment.

src/main/java/org/alliancegenome/curation_api/services/associations/alleleAssociations/AlleleGeneAssociationService.java
```diff
@@ -80,20 +91,40 @@ public List<Long> getAlleleGeneAssociationsByDataProvider(BackendBulkDataProvide
...
+				throw new ApiErrorException(response);
+			}
+			log.error(errorMessage);
+			return null;
+		}
+		if (deprecate) {
+			if (!association.getObsolete()) {
+				association.setObsolete(true);
+				if (authenticatedPerson.getOktaEmail() != null) {
+					association.setUpdatedBy(personService.findPersonByOktaEmail(authenticatedPerson.getOktaEmail()));
```
oblodgett: You should be able to use the authenticatedPerson as the object for setUpdatedBy... if not you should be able to use the authenticatedPerson.id for a personDao.get(id) as in authenticatedPerson is the object from the database. So it should be able to be used as such without another lookup.

src/main/java/org/alliancegenome/curation_api/services/associations/alleleAssociations/AlleleGeneAssociationService.java
```diff
@@ -80,20 +91,40 @@ public List<Long> getAlleleGeneAssociationsByDataProvider(BackendBulkDataProvide
...
+				} else {
+					association.setUpdatedBy(personService.fetchByUniqueIdOrCreate(loadDescription));
```
oblodgett, line 113: Are we creating a bunch of "persons" based on loadDescriptions? I don't think this is a good idea as in the person table is not really "persons" anymore?

markquintontulloch: Yeah, not sure what the best solution is without a model change though.  We've been doing the same for disease annotations as we want the "UpdatedBy" to indicate why the annotation was changed to obsolete and we don't have a person associated with a bulk load.

src/main/java/org/alliancegenome/curation_api/services/associations/alleleAssociations/AlleleGeneAssociationService.java
markquintontulloch: Updated here and in the DiseaseAnnotationService

src/main/java/org/alliancegenome/curation_api/services/associations/alleleAssociations/AlleleGeneAssociationService.java
markquintontulloch: We also create "people" from whatever gets submitted in the updatedBy and createdBy fields in bulk submissions, which are often not really people either.

oblodgett (approved):

PR #1291, opened 2023-10-11
add checks for inferredAllele

adamgibs:


markquintontulloch (approved):

markquintontulloch:
@adamgibs I think this error is being caused by my allele association PR that went in.  There's an annotation that I forgot to remove on the allele entity class that means the allele object isn't being returned in some of the results, but just the curie: e.g. `"inferredAllele": "MGI:5575656"`.  The error will still be thrown after this PR goes in as curie won't be found either.  My current open PR should fix this.

adamgibs:
@markquintontulloch Ah, okay. That makes sense

PR #1292, opened 2023-10-12
SCRUM-3321 Load and display Variants

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/dao/base/BaseSQLDAO.java
```diff
@@ -111,6 +111,22 @@ public E find(Long id) {
...
+	
+	public List<String> findFilteredIds(Map<String, Object> params) {
+		List<String> primaryKeys = new ArrayList<>();
+		SearchResponse<E> results = findByParams(params);
+		for (E entity : results.getResults()) {
+			String pkString;
+			try {
+				pkString = (String) entityManager.getEntityManagerFactory().getPersistenceUnitUtil().getIdentifier(entity);
+			} catch (ClassCastException e) {
+				pkString = Long.toString((Long) entityManager.getEntityManagerFactory().getPersistenceUnitUtil().getIdentifier(entity));
```
oblodgett: May want to catch any errors here also. (I can't think of where we would have a different type of PK)

src/main/java/org/alliancegenome/curation_api/interfaces/crud/VariantCrudInterface.java
```diff
@@ -0,0 +1,73 @@
...
+	@Tag(name = "Relational Database Browsing Endpoints")
+	@JsonView(View.VariantView.class)
+	public SearchResponse<Variant> find(@DefaultValue("0") @QueryParam("page") Integer page, @DefaultValue("10") @QueryParam("limit") Integer limit, @RequestBody HashMap<String, Object> params);
+
+	@Override
+	@POST
+	@Path("/search")
+	@Tag(name = "Search Database Browsing Endpoints")
+	@JsonView({ View.VariantView.class })
+	public SearchResponse<Variant> search(@DefaultValue("0") @QueryParam("page") Integer page, @DefaultValue("10") @QueryParam("limit") Integer limit, @RequestBody HashMap<String, Object> params);
```
oblodgett, line 71: We shouldn't need to override all these methods as they should be coming from `BaseCurieCrudInterface`

src/main/java/org/alliancegenome/curation_api/services/GeneDiseaseAnnotationService.java
```diff
@@ -92,18 +95,14 @@ public ObjectResponse<GeneDiseaseAnnotation> delete(Long id) {
...
-		if( StringUtils.equals(sourceOrg, "RGD") ){
-			annotationIds = geneDiseaseAnnotationDAO.findAllAnnotationIdsByDataProvider(dataProvider.sourceOrganization, dataProvider.canonicalTaxonCurie);
-		} else {
-			annotationIds = geneDiseaseAnnotationDAO.findAllAnnotationIdsByDataProvider(dataProvider.sourceOrganization);
-		}
-
-		annotationIds.removeIf(Objects::isNull);
-
+		Map<String, Object> params = new HashMap<>();
+		params.put("dataProvider.sourceOrganization.abbreviation", dataProvider.sourceOrganization);
```
oblodgett: Given the amount that we are using this field name... might be good to put them into a constants file... it would just make it easier to change in the future if this field name changes.

src/main/java/org/alliancegenome/curation_api/interfaces/crud/VariantCrudInterface.java
markquintontulloch: We do this all over the place - I thought it was because we wanted to override the view, but maybe I've misunderstood.

src/main/java/org/alliancegenome/curation_api/dao/base/BaseSQLDAO.java
```diff
@@ -111,6 +111,22 @@ public E find(Long id) {
 			return null;
 		}
 	}
+	
+	public List<String> findFilteredIds(Map<String, Object> params) {
+		List<String> primaryKeys = new ArrayList<>();
+		SearchResponse<E> results = findByParams(params);
+		for (E entity : results.getResults()) {
```
oblodgett, line 118: The only concern I have with this is if the result set is large. We may not get an indication of why the server is "hanging". This pulls the whole object and then pulls out the PK. I think there is a way to only ask for the PK. Basically a `select id from Table` vs `select * from Table` we can go with this for now, but it might bite us on large result sets in the future.

src/main/java/org/alliancegenome/curation_api/services/VariantService.java
```diff
@@ -0,0 +1,78 @@
...
+import org.apache.commons.collections.CollectionUtils;
+
+import lombok.extern.jbosslog.JBossLog;
+
+@JBossLog
+@RequestScoped
+public class VariantService extends BaseDTOCrudService<Variant, VariantDTO, VariantDAO> {
+
+	@Inject VariantDAO variantDAO;
+	@Inject PersonService personService;
```
oblodgett: Is this being used here?

src/main/java/org/alliancegenome/curation_api/interfaces/crud/VariantCrudInterface.java
oblodgett: Ahh yes the view... sorry wasn't thinking

src/main/java/org/alliancegenome/curation_api/services/VariantService.java
markquintontulloch: Nope, removed

adamgibs (approved):
JS looks good. Using a common related notes dialog is a good idea.

src/main/java/org/alliancegenome/curation_api/dao/base/BaseSQLDAO.java
markquintontulloch: What would we do if we caught the exception?  As you say, I can't think of any other types of PK we're going to encounter.

src/main/java/org/alliancegenome/curation_api/services/GeneDiseaseAnnotationService.java
markquintontulloch: @oblodgett Done

src/main/java/org/alliancegenome/curation_api/dao/base/BaseSQLDAO.java
markquintontulloch: Agreed - at least we now have a shared method that we can update with a better solution though.

src/main/java/org/alliancegenome/curation_api/constants/EntityConstants.java
```diff
@@ -0,0 +1,8 @@
+package org.alliancegenome.curation_api.constants;
+
+public final class EntityConstants {
```
oblodgett: Sorry to be weird about names but can we name this `EntityFieldNameConstants`?

src/main/java/org/alliancegenome/curation_api/dao/base/BaseSQLDAO.java
oblodgett: So I guess maybe a 
`String pkString = null;`

and later:

if(pkString != null) {
primaryKeys.add(pkString);
}

That way if it does fail... things can still move forward maybe this would happen anyway not sure.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/constants/EntityConstants.java
markquintontulloch: Sure, I didn't like the name either! Done.

src/main/java/org/alliancegenome/curation_api/dao/base/BaseSQLDAO.java
markquintontulloch: Done

PR #1294, opened 2023-10-12
SCRUM-2248 update - more fixes

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
```diff
@@ -99,7 +99,7 @@ public DiseaseAnnotation deprecateOrDeleteAnnotationAndNotes(Long id, Boolean th
 		if (deprecateAnnotation) {
 			if (!annotation.getObsolete()) {
 				annotation.setObsolete(true);
-				if (authenticatedPerson != null) {
+				if (authenticatedPerson.getOktaEmail() != null) {
```
markquintontulloch: Can't just check for null authenticatedPerson as on bulk loads authenticatedPerson is a non-null object full of null fields, so updatedBy just gets set to null.

src/main/java/org/alliancegenome/curation_api/services/associations/alleleAssociations/AlleleGeneAssociationService.java
```diff
@@ -111,14 +111,15 @@ public AlleleGeneAssociation deprecateOrDeleteAssociation(Long id, Boolean throw
...
-				if (authenticatedPerson != null) {
+				if (authenticatedPerson.getOktaEmail() != null) {
 					association.setUpdatedBy(personDAO.find(authenticatedPerson.getId()));
 				} else {
 					association.setUpdatedBy(personService.fetchByUniqueIdOrCreate(loadDescription));
 				}
 				association.setDateUpdated(OffsetDateTime.now());
 				return alleleGeneAssociationDAO.persist(association);
 			}
+			return association;
```
markquintontulloch, line 126: Need this to avoid deleting association if already obsolete.

src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
oblodgett: Ahh I didn't think of this case.  I think it would be better to have it null vs an object with all null fields?

markquintontulloch: Have changed this to `if (authenticatedPerson.getId() != null)` - don't know why I used OktaEmail field to check!

src/main/java/org/alliancegenome/curation_api/jobs/executors/LoadFileExecutor.java
```diff
@@ -210,7 +210,7 @@ public void runCleanup(DiseaseAnnotationService service, BulkLoadFileHistory his
 		for (Long id : idsToRemove) {
 			try {
 				String loadDescription = dataProviderName + " association bulk load (" + md5sum + ")";
-				service.deprecateOrDeleteAssociation(id, false, loadDescription, true);
+				service.deprecateOrDeleteAssociation(id, false, loadDescription, false);
```
markquintontulloch, line 213: This should have been false as want to delete rather than deprecate when doing cleanup of association loads - only want to deprecate if the method is called by the gene/allele load cleanup code

src/main/java/org/alliancegenome/curation_api/services/DiseaseAnnotationService.java
oblodgett: We can change it so that authenticatedPerson is null? Or leave it?

markquintontulloch: Not sure that I want to start tinkering with authenticatedPerson at this point (the details of the whole authenticatedPerson creation are still a little hazy to me) - maybe a separate PR

oblodgett (approved):

PR #1296, opened 2023-10-13
Give correct log message

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/jobs/executors/LoadFileExecutor.java
```diff
@@ -181,7 +184,7 @@ public void runCleanup(DiseaseAnnotationService service, BulkLoadFileHistory his
 		ph.startProcess("Deletion/deprecation of primary objects " + dataProviderName, curiesToRemove.size());
 		for (String curie : curiesToRemove) {
 			try {
-				String loadDescription = dataProviderName + " association bulk load (" + md5sum + ")";
+				String loadDescription = dataProviderName + " " + manual.getBackendBulkLoadType() + " bulk load (" + bulkLoadFile.getMd5Sum() + ")";
```
oblodgett, line 187: Just to be consistent you might want to change `bulkLoadFile.getMd5Sum()` to `manual.getMd5Sum()` as in after "conversion" `BulkManualLoad manual = (BulkManualLoad) bulkLoadFile.getBulkLoad();` always refer to it as manual from that point forward...

markquintontulloch: `BulkManualLoad` doesn't have the `getMd5Sum()` method

oblodgett: I'm sorry you are capturing the upper class not just casting it... didn't catch that.

oblodgett (approved):

PR #1298, opened 2023-10-16
SCRUM-3325 Load and display ConstructGenomicEntityAssociation

markquintontulloch:


adamgibs (approved):

src/main/java/org/alliancegenome/curation_api/dao/GenomicEntityDAO.java
```diff
@@ -0,0 +1,15 @@
+package org.alliancegenome.curation_api.dao;
+
+import javax.enterprise.context.ApplicationScoped;
+
+import org.alliancegenome.curation_api.dao.base.BaseSQLDAO;
+import org.alliancegenome.curation_api.model.entities.GenomicEntity;
+
+@ApplicationScoped
+public class GenomicEntityDAO extends BaseSQLDAO<GenomicEntity> {
+
```
cmpich, line 10: What do you need this class for?

src/main/java/org/alliancegenome/curation_api/services/ConstructService.java
```diff
@@ -119,8 +123,13 @@ public Construct removeOrDeprecateNonUpdated(Long id, Boolean throwApiError, Str
 		}
 		
 		Boolean anyReferencingEntities = false;
-		// TODO: implement check for any referencing entities and code for their deprecation
-		// once links between constructs and other entities have been established
+		if (CollectionUtils.isNotEmpty(construct.getConstructGenomicEntityAssociations())) {
```
cmpich, line 126: This piece of code is duplicated in multiple places. Could that be re-used in a single class?

src/main/java/org/alliancegenome/curation_api/services/associations/constructAssociations/ConstructGenomicEntityAssociationService.java
```diff
@@ -0,0 +1,179 @@
...
+		return response;
+	}
+	
+	private void addAssociationToConstruct(ConstructGenomicEntityAssociation association) {
+		Construct construct = association.getSubject();
+		List<ConstructGenomicEntityAssociation> currentAssociations = construct.getConstructGenomicEntityAssociations();
+		if (currentAssociations == null)
+			currentAssociations = new ArrayList<>();
+		List<Long> currentAssociationIds = currentAssociations.stream().map(ConstructGenomicEntityAssociation::getId).collect(Collectors.toList());
+		if (!currentAssociationIds.contains(association.getId()));
```
cmpich, line 162: Why not using a Set that only collects unique items?

markquintontulloch: It's retrieving a list of IDs for associations attached to a single Construct object, so they will be unique by default.

src/main/java/org/alliancegenome/curation_api/services/ConstructService.java
markquintontulloch: There are class-dependent differences

src/main/java/org/alliancegenome/curation_api/dao/GenomicEntityDAO.java
markquintontulloch: It's used in the validation code for ConstructGenomicEntity associations.  It's the first time we've had a GenomicEntity field and not the BiologicalEntity parent class or one of the child classes.

cmpich: Ok.

src/main/java/org/alliancegenome/curation_api/services/ConstructService.java
cmpich: With the List<ConstructGenomicEntityAssociation> as an argument you could write a convenience method that return a boolean. That method checks if the collection has deprecated of deleted Associations. Maybe, it could live in a shared service class.

markquintontulloch: Sure, will take a look at it later and put in a separate PR so we can get this stuff in and the associations loaded before sprint review.

src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmDiseaseAnnotationExecutor.java
```diff
@@ -41,43 +38,33 @@ public class AgmDiseaseAnnotationExecutor extends LoadFileExecutor {
...
-			
-			runLoad(history, dataProvider, annotations, annotationIdsLoaded);
-			
-			if(cleanUp) runCleanup(diseaseAnnotationService, history, dataProvider.name(), annotationIdsBefore, annotationIdsLoaded, bulkLoadFile.getMd5Sum());
-
-			history.finishLoad();
-			
-			trackHistory(history, bulkLoadFile);
-
-		} catch (Exception e) {
```
oblodgett, line 77: @markquintontulloch was there a reason to get rid of the try catch? These loads can fail for many different reasons, other than the obvious?

markquintontulloch: The try/catch moved to the new `readIngestFile()` method in `LoadFileExecutor` that does the actual loading of the file

PR #1306, opened 2023-10-25
Fixed tab issues and change to amazon cert for curation sites

oblodgett:


oblodgett:
This has not changed anything fundamentally. So not sure why the tests are failing.

.github/workflows/github-actions-PR-validation.yml
```diff
@@ -54,6 +54,9 @@ jobs:
       - name: Package API
         run: |
           mvn clean package -Dquarkus.package.type=uber-jar -ntp
+      - name: Create test containers properties file
+        run: |
+          touch ~/.testcontainers.properties
```
oblodgett, line 59: Looks like github now requires this file for test containers... thanks github for letting us know.

.ebextensions/loadbalancer.config
```diff
@@ -25,7 +25,7 @@ option_settings:
   # https://docs.aws.amazon.com/elasticbeanstalk/latest/dg/environments-cfg-alb.html#environments-cfg-alb-namespaces
 
   aws:elbv2:listener:443:
-    SSLCertificateArns: arn:aws:acm:us-east-1:100225593120:certificate/70263776-5e62-4454-9997-bb7e560d7c63
+    SSLCertificateArns: arn:aws:acm:us-east-1:100225593120:certificate/047a56a2-09dd-4857-9f28-32d23650d4da
```
oblodgett, line 28: @mluypaert will this take effect or are we going to need to delete the ENV?

oblodgett: PS @abecerra this will need to be done in Mati and the FMS.

oblodgett: Switches from our own manually updated cert to amazons auto renewed cert.

mluypaert: > @mluypaert will this take effect or are we going to need to delete the ENV?

Not sure, this will be a load balancer change so we'll need to see if EB can handle this as an update. I would expect this not to be a problem as it it a single property on the load balancer, nothing would change from the application instance's point of view. The deployment GH actions workflow will fail with clear errors otherwise, in which case we will know we need to delete and recreate.

mluypaert (approved):

.ebextensions/loadbalancer.config
oblodgett: @mluypaert ok... I am also thinking that maybe we can go and just update them manually in the LB's rather than deleting the ENV

PR #1309, opened 2023-10-30
Update SiteLayout.js

VarunReddy1111:
Made changes to menu that sorts Data tables alphabetically


src/main/cliapp/src/containers/layout/SiteLayout.js
```diff
@@ -340,6 +340,14 @@ export const SiteLayout = (props) => {
 				}
 		];
 
+		const sortedMenu = [...menu];
+		sortedMenu[0].items.forEach((item) => {
+			if (item.label.match("Data Tables")) {
+				item.items = item.items.sort((a, b) => a.label.localeCompare(b.label));
+			}
+		});
```
oblodgett: It would probally be better to just hardcode the order rather than using a sort method. This way if the lists get rearranged or changed in some way there is not unexpected behavior.

oblodgett (changes requested):

oblodgett (approved):
Looks great!!!

oblodgett:
@mluypaert not sure why these are not updating?

mluypaert:
> @mluypaert not sure why these are not updating?

@oblodgett There was a github outage around that time, so that was probably causing this as well. https://www.githubstatus.com/history?page=1

oblodgett:
@mluypaert ya didn't think to check, after a long while just changed to title and everything fired again

PR #1316, opened 2023-11-09
Fixed issue with failing vocab query, however frontend needs to send …

oblodgett:
…in the vocab to be searched

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/VocabularyTermService.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/VocabularyTermService.java
@@ -106,7 +106,7 @@ public class VocabularyTermService extends BaseEntityCrudService<VocabularyTerm,
 	private VocabularyTerm getTermInVocabularyFromDB(String vocabularyLabel, String termName) {
 		Map<String, Object> params = new HashMap<>();
 		params.put("name", termName);
-		params.put("vocabulary.vocabularyLabel", vocabularyLabel);
+		if(vocabularyLabel != null) params.put("vocabulary.vocabularyLabel", vocabularyLabel);
 		SearchResponse<VocabularyTerm> resp = vocabularyTermDAO.findByParams(params);
 		return resp.getSingleResult();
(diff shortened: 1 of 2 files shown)


src/main/java/org/alliancegenome/curation_api/services/VocabularyTermService.java
```diff
@@ -106,16 +106,15 @@ public ObjectResponse<VocabularyTerm> getTermInVocabularyTermSet(String vocabula
 	private VocabularyTerm getTermInVocabularyFromDB(String vocabularyLabel, String termName) {
 		Map<String, Object> params = new HashMap<>();
 		params.put("name", termName);
-		params.put("vocabulary.vocabularyLabel", vocabularyLabel);
+		if(vocabularyLabel != null) params.put("vocabulary.vocabularyLabel", vocabularyLabel);
```
oblodgett, line 109: @adamgibs @markquintontulloch for some reason the UI is only sending in the "termName" and not the vocabularyLabel so the lookup was failing. I fixed it by doing a null check but I think the UI should be fixed as in the backend will look to make sure this field is null if it is send in without a value.

markquintontulloch: Where is the UI code calling this method? I thought it just got called by the validation code, which passes in the label using the Vocabulary constants.

oblodgett: There were errors in the log files showing just the termName was being sent to the endpoint... not sure what is causing that but those were errors I was seeing.

markquintontulloch: Weird.  Only thing I can think of is someone calling the  /vocabularyterm/findBy endpoint without specifying a vocabularyLabel.  Well, this solves that at least and would give an appropriate response.

markquintontulloch (approved):

PR #1318, opened 2023-11-10
Reorged yet again the base sql query stuff

oblodgett:


markquintontulloch:
Looks like this has reintroduced the issue with searching lists of strings

markquintontulloch (approved):

PR #1330, opened 2023-11-15
Use enum for bulkload file extensions

markquintontulloch:


oblodgett:
The only issue with this change is that we will not be able to load the same data from two different types of files. We don't have this case right now. But we will when it comes to interaction data. Two different file formats but loading the same data at the end of the day. Maybe those can be implemented as two loads but this change will force them to be implemented as two loads.

oblodgett (approved):

markquintontulloch:
I think we'd want that as two separate load types anyway as we'd have different executors.

oblodgett:
It would be the same executor... we would just load the file two different ways.

PR #1334, opened 2023-11-15
Data loads - Remove polling and add event driven

oblodgett:


oblodgett:
This is to help diagnose the other issues related to quarkus and not have sql query polling to the database.

markquintontulloch (approved):

PR #1336, opened 2023-11-15
SCRUM-3507 - Possible fix for 'failed to initialize collection...' error

markquintontulloch:
Note sure this will work as cannot reproduce the error locally, but we had the same error for VocabularyTerm synonyms and fixed it like this.

Current error:
`failed to lazily initialize a collection of role: org.alliancegenome.curation_api.model.entities.Person.emails: could not initialize proxy - no Session (through reference chain: org.alliancegenome.curation_api.response.ObjectResponse["entity"]->org.alliancegenome.curation_api.model.entities.VocabularyTerm["vocabulary"]->org.alliancegenome.curation_api.model.entities.Vocabulary["updatedBy"]->org.alliancegenome.curation_api.model.entities.Person["emails"])
Nov 15 23:02:23 [agr.curation.alpha.api.server](http://logs.alliancegenome.org:5601/app/logtrail) : 	at com.fasterxml.jackson.databind.JsonMappingException.wrapWithPath(JsonMappingException.java:361)`

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/PersonService.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/PersonService.java
@@ -43,6 +43,8 @@ public class PersonService extends BaseEntityCrudService<Person, PersonDAO> {
 			} else {
 				Log.debug("Person not cached, caching uniqueId: (" + uniqueId + ")");
 				person = findPersonByUniqueIdOrCreateDB(uniqueId);
+				if (person != null)
+					person.getEmails().size();
 				personCacheMap.put(uniqueId, person);
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


markquintontulloch:
Or it could just kill all the bulk load integration tests!

oblodgett (approved):
Yes this saves a huge amounts of joins during other query operations.

oblodgett:
I typically don't like calling the underlying API (hibernate) directly but rather use the upper API (entitymanager). However I see this:

```
Use Hibernate.initialize() within @Transactional to initialize lazy objects.

 start Transaction 
      Hibernate.initialize(entity.getAddresses());
      Hibernate.initialize(entity.getPersons());
 end Transaction 
Now out side of the Transaction you are able to get lazy objects.

entity.getAddresses().size();
entity.getPersons().size();
```

So I guess this would be the best option.

PR #1337, opened 2023-11-16
SCRUM-3507 Eager loading of Person emails and oldEmails

markquintontulloch:
I've tried a bunch of stuff to solve the `failed to lazily initialize a collection of role...` error, but this is the only solution I can get to work.  I don't think that eager loading is too bad in this case.


markquintontulloch:
Actually eager loading only removes the error (which actually we can do with views), it doesn't fix the indexing on insertion - closing PR

PR #1343, opened 2023-11-20
SCRUM-3524

adamgibs:


oblodgett:
@adamgibs I was thinking after we got off the call... it might be better to just default it to true, that way you can just leave it out on all the columns and if you don't want it you can override by setting it to false.

oblodgett (approved):
@adamgibs I was thinking after we got off the call... it might be better to just default it to true, that way you can just leave it out on all the columns and if you don't want it you can override by setting it to false.

adamgibs:
@oblodgett Yeah, sure. That works. Either way is fine with me

PR #1344, opened 2023-11-20
SCRUM-3473 Add supplemental data error map to API response for alleles

markquintontulloch:


oblodgett:
@markquintontulloch I think a bunch of that logic can be moved into the APIRepsonce object in order to simplify the other places that are using the errorMap... I'll put in a PR to your PR.

oblodgett (approved):
Before merging can you double check that I didn't cause any major issue?

markquintontulloch:
Will do

PR #1347, opened 2023-11-21
Refactored map logic into one place

oblodgett:


oblodgett:
@markquintontulloch I think this puts everything into one place

src/main/java/org/alliancegenome/curation_api/response/SupplementalDatahelper.java
```diff
@@ -0,0 +1,79 @@
...
+public class SupplementalDatahelper {
+
+	public static final String ERROR_MAP = "errorMap";
+	
+	public static void addFieldErrorMessages(Map<String, Object> supplementalData, String fieldName, Object fieldErrorMessages) {
+		Map<String, Object> errorMap = getErrorMap(supplementalData);
+		if(!errorMap.containsKey(fieldName)) {
+			errorMap.put(fieldName, fieldErrorMessages);
+		} else {
+			Log.warn("Error Map already contains errors for field: " + fieldName);
```
oblodgett, line 23: @markquintontulloch this is the only question I have is this now happens globally as in if there is already a error message we check and throw this warning if it already exists.

markquintontulloch (approved):

markquintontulloch:
Looks good. Definitely neater.

PR #1350, opened 2023-11-24
SCRUM-3519 Prevent creation of orphaned slot annotation entries & cleanup and validation refactor

markquintontulloch:


oblodgett (approved):
@markquintontulloch did the new version of hibernate help make this work... or was CASCADE just not used before?

markquintontulloch:
We just hadn't used cascade before (or if we had, it was removed at some point for some reason)

markquintontulloch:
Actually, we had used cascade in many places (but not orphanRemoval).  I don't know if this just wasn't working for persist, or if we just had unnecessary additional DAO calls in the code.

PR #1353, opened 2023-11-28
SCRUM-3529 Temporary workaround for broken disease annotation table

markquintontulloch:


src/main/resources/db/migration/v0.29.0.4__rename_construct_association_fields.sql
```diff
@@ -0,0 +1,8 @@
+ALTER TABLE constructgenomicentityassociation RENAME subject_id TO subjectconstruct_id;
+ALTER TABLE constructgenomicentityassociation_aud RENAME subject_id TO subjectconstruct_id;
+
+ALTER TABLE constructgenomicentityassociation RENAME object_curie TO objectgenomicentity_curie;
+ALTER TABLE constructgenomicentityassociation_aud RENAME object_curie TO objectgenomicentity_curie;
+
+ALTER TABLE allelegeneassociation RENAME object_curie TO objectgene_curie;
+ALTER TABLE allelegeneassociation_aud RENAME object_curie TO objectgene_curie;
```
oblodgett, line 8: @markquintontulloch was there FK's or indexes that need to be renamed also?

oblodgett (approved):

src/main/resources/db/migration/v0.29.0.4__rename_construct_association_fields.sql
markquintontulloch: Postgres should automatically update them to point at the new column name.  No point renaming the constraints and indexes themselves as this is going to be reverted shortly anyway.

PR #1355, opened 2023-11-30
SCRUM-3536 add isLoading boolean to AlleleDetailPage

adamgibs:


oblodgett (commented):
aInstead of baking this in everywhere couldn't a component be created that wraps everything to disable it?

oblodgett:
Here is the example i am referring to https://stackoverflow.com/questions/69263080/how-to-hide-show-all-react-component-when-loading-screen-is-showing-up Just do the hiding at a higher level

adamgibs:
@oblodgett In the meeting we talked about a solution that involved greying everything out while still being able to see the components. You mentioned using the individual component boolean properties to do this, which is why I went down this route. Yes, hiding everything with a spinner or other component while loading is simple (that's already happening with the Get request), but that's not what we talked about in the meeting.

oblodgett:
My idea was to use an overlay at that level that could cover everything while keeping it all on the page

adamgibs:
What's the down side of leaving it as it?

oblodgett:
Its a lot more code and the maintenance is much higher, its not a "simple" solution. If you don't implement loading in a new component it's shown where everything else is still hidden.

adamgibs:
Okay, I'll work on it.

oblodgett:
@adamgibs I thought this PR was closed?

adamgibs:
@oblodgett Odd. It was merged in days ago and the code is on alpha. I'll go ahead and close it again

adamgibs:
@oblodgett Ah, I just used a different branch and different PR to merge in what ended up on alpha and I forgot to close this PR

PR #1358, opened 2023-12-01
SCRUM-3536 add LoadingOverlay to AlleleDetailPage

adamgibs:


abecerra (approved):

oblodgett:
@adamgibs does this take over the whole page or only the portion under the allele symbol?

adamgibs:
@oblodgett Whole page

adamgibs:
Whole page under the main site header that is

oblodgett (approved):
I guess if the curator really want to do something else while an allele is saving they could just have another window to the site open.

adamgibs:
Side bar is still functional. Just the page including the sticky header

PR #1367, opened 2023-12-12
SCRUM-3354

adamgibs:


markquintontulloch (approved):
Sorry, I realise you're having to jump through some hoops here to get around the association operations being separated from the Allele validation and saving.  I think this is good for now but I need to sort out the API side ASAP to handle it all just using the allele endpoint - I don't think that's too complicated, but won't happen this sprint.

adamgibs:
@markquintontulloch No worries! Yeah, I assumed most this alleleGeneAssociation service code would be temporary.

PR #1375, opened 2024-01-03
SCRUM-3515 LinkML v2.0.0 refactor

markquintontulloch:
Known issues:

1. Genomic/BiologicalEntity DAO searches expecting an NCBITaxonTerm for some reason!
2. Persistence exception for multivalued allele slot annotations - could just move code that attaches subject Allele, but the same logic is working fine for multivalued allele name slot annotations (would like to understand why there's a difference before just changing the code).
3. Still seem to be having issues resolving subject/object of different types
4. UI changes untested and probably more required


src/main/java/org/alliancegenome/curation_api/interfaces/crud/ReferenceCrudInterface.java
```diff
@@ -19,14 +19,14 @@
...
 	@GET
 	@Path("/sync")
 	public void synchroniseReferences();
 
 	@GET
-	@Path("/sync/{curie}")
+	@Path("/sync/{id}")
 	@JsonView(View.FieldsAndLists.class)
-	public ObjectResponse<Reference> synchroniseReference(@PathParam("curie") String curie);
+	public ObjectResponse<Reference> synchroniseReference(@PathParam("curie") Long id);
```
oblodgett: Need to change the PathParam here to "id"

src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmDiseaseAnnotationExecutor.java
```diff
@@ -92,9 +93,11 @@ public void runLoad(BulkLoadFileHistory history, BackendBulkDataProvider dataPro
 					idsAdded.add(annotation.getId());
 				}
 			} catch (ObjectUpdateException e) {
+				Log.info("EXCEPTION 1: " + e.getData());
```
oblodgett: I would guess these should be Log.error() which will turn it red in the logs.

src/main/java/org/alliancegenome/curation_api/interfaces/crud/AffectedGenomicModelCrudInterface.java
```diff
@@ -24,11 +24,11 @@
 @Tag(name = "CRUD - Affected Genomic Models")
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
-public interface AffectedGenomicModelCrudInterface extends BaseCurieCrudInterface<AffectedGenomicModel>, BaseDTOCrudControllerInterface<AffectedGenomicModel, AffectedGenomicModelDTO> {
+public interface AffectedGenomicModelCrudInterface extends SubmittedObjectCrudInterface<AffectedGenomicModel>, BaseDTOCrudControllerInterface<AffectedGenomicModel, AffectedGenomicModelDTO> {
 
 	@Override
 	@JsonView(View.FieldsAndLists.class)
-	public ObjectResponse<AffectedGenomicModel> get(@PathParam("curie") String curie);
+	public ObjectResponse<AffectedGenomicModel> get(@PathParam("identifierString") String identifierString);
```
oblodgett, line 31: What is the purpose of changing this over to identifierString? Because now we will have both "curie" and "identifierString"

src/main/java/org/alliancegenome/curation_api/model/entities/ontology/OntologyTerm.java
```diff
@@ -41,14 +38,9 @@
 @Data
 @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
 @Entity
-@Inheritance(strategy = InheritanceType.JOINED)
```
oblodgett, line 44: This could be the answer to the slow ontologies load.

src/main/java/org/alliancegenome/curation_api/model/entities/Allele.java
```diff
@@ -59,8 +60,13 @@ public class Allele extends GenomicEntity {
 
 	@IndexedEmbedded(includePaths = {"primaryCrossReferenceCurie", "crossReferences.referencedCurie", "crossReferences.displayName", "curie", "primaryCrossReferenceCurie_keyword", "crossReferences.referencedCurie_keyword", "crossReferences.displayName_keyword", "curie_keyword"})
 	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
-	@ManyToMany
-	@JoinTable(indexes = { @Index(columnList = "allele_curie"), @Index(columnList = "references_curie") })
+	@ManyToMany(fetch = FetchType.EAGER)
```
oblodgett: Do we need this fetch Type to be EAGER? as it will cause alleles to get loaded with all the references and I would guess slow things down even more.

src/main/java/org/alliancegenome/curation_api/interfaces/crud/GeneCrudInterface.java
```diff
@@ -31,13 +31,13 @@
...
-public interface GeneCrudInterface extends BaseCurieCrudInterface<Gene>, BaseDTOCrudControllerInterface<Gene, GeneDTO> {
+public interface GeneCrudInterface extends SubmittedObjectCrudInterface<Gene>, BaseDTOCrudControllerInterface<Gene, GeneDTO> {
 
 	@Override
 	@GET
-	@Path("/{curie}")
+	@Path("/{identifierString}")
 	@JsonView(View.GeneDetailView.class)
-	public ObjectResponse<Gene> get(@PathParam("curie") String curie);
+	public ObjectResponse<Gene> get(@PathParam("identifierString") String identifierString);
```
oblodgett, line 40: I think this method could get moved into another interface... IdentifiableObjectInterface and then the other interfaces could extend it so that it doesn't have to be defined a bunch of places.... unless the view is important here?

src/main/java/org/alliancegenome/curation_api/dao/ontology/NcbiTaxonTermDAO.java
```diff
@@ -1,65 +1,15 @@
...
-		if (taxonMap.get("status").equals("active")) {
-			taxon.setObsolete(false);
-		} else {
-			taxon.setObsolete(true);
-		}
-		persist(taxon);
-
-		return taxon;
-	}
-
```
oblodgett, line 64: I'm guessing this code got moved somewhere else... just haven't seen it yet?

src/main/java/org/alliancegenome/curation_api/model/entities/GenomicEntity.java
```diff
@@ -36,20 +35,21 @@ public class GenomicEntity extends BiologicalEntity {
 
 	@IndexedEmbedded(includePaths = {"referencedCurie", "displayName", "referencedCurie_keyword", "displayName_keyword"})
 	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
-	@ManyToMany
+	@OneToMany(cascade = CascadeType.ALL, orphanRemoval=true)
```
oblodgett, line 33: This should be a many to many with the join table

src/main/java/org/alliancegenome/curation_api/model/ingest/dto/GeneDiseaseAnnotationDTO.java
```diff
@@ -12,15 +12,15 @@
...
 @Getter
-@AGRCurationSchemaVersion(min = "1.4.0", max = LinkMLSchemaConstants.LATEST_RELEASE, dependencies = { DiseaseAnnotationDTO.class }, submitted = true)
+@AGRCurationSchemaVersion(min = "2.0.0", max = LinkMLSchemaConstants.LATEST_RELEASE, dependencies = { DiseaseAnnotationDTO.class }, submitted = true)
 public class GeneDiseaseAnnotationDTO extends DiseaseAnnotationDTO {
 
 	@JsonView({ View.FieldsOnly.class })
-	@JsonProperty("gene_curie")
-	private String geneCurie;
+	@JsonProperty("gene_identifier")
+	private String geneIdentifier;
```
oblodgett, line 20: Are these identifier names reflected in the LinkML?

src/main/java/org/alliancegenome/curation_api/services/ontology/NcbiTaxonTermService.java
```diff
@@ -51,15 +60,55 @@ public ObjectResponse<NCBITaxonTerm> get(String taxonCurie) {
...
+			taxon = downloadAndSave(taxonCurie);
 			if (taxon == null) {
 				Log.warn("Taxon ID could not be found");
 			}
 		}
 		return taxon;
 	}
+	
+	public NCBITaxonTerm downloadAndSave(String taxonCurie) {
+
```
oblodgett, line 75: Ahh ok found it.

src/main/java/org/alliancegenome/curation_api/services/ontology/NcbiTaxonTermService.java
```diff
@@ -51,15 +60,55 @@ public ObjectResponse<NCBITaxonTerm> get(String taxonCurie) {
...
+		String name = (String) taxonMap.get("scientificname");
+		NCBITaxonTerm taxon = new NCBITaxonTerm();
+		taxon.setName(name);
+		taxon.setCurie(taxonCurie);
+		if (taxonMap.get("status").equals("active")) {
+			taxon.setObsolete(false);
+		} else {
+			taxon.setObsolete(true);
+		}
+		ncbiTaxonTermDAO.persist(taxon);
```
oblodgett: you will want to return `ncbiTaxonTermDAO.persist(taxon);` as it will have the new ID I am not sure it will get updated in the taxon object?

src/main/java/org/alliancegenome/curation_api/services/base/CurieObjectCrudService.java
```diff
@@ -0,0 +1,44 @@
...
+	@Transactional
+	public ObjectResponse<E> delete(String curie) {
+		E object = findByCurie(curie);
+		if (object != null)
+			dao.remove(object.getId());
+		ObjectResponse<E> ret = new ObjectResponse<>(object);
+		return ret;
+	}
+	
+	public E findByCurie(String curie) {
```
oblodgett, line 28: This might be better moved into the BaseDAO that way these other classes don't have to inject both the service and the dao.

src/main/java/org/alliancegenome/curation_api/services/OrganizationService.java
```diff
@@ -38,11 +37,11 @@ public ObjectResponse<Organization> get(String orgId) {
 				org = orgCacheMap.get(orgId);
 			} else {
 				Log.debug("Org not cached, caching org: (" + orgId + ")");
-				org = organizationDAO.find(orgId);
+				org = organizationDAO.find(Long.parseLong(orgId));
```
oblodgett, line 41: Might want to change the definition of this method to take a Long vs String?

src/main/java/org/alliancegenome/curation_api/services/ExperimentalConditionService.java
```diff
@@ -41,18 +41,18 @@ public ObjectResponse<ExperimentalCondition> update(ExperimentalCondition uiEnti
 	@Transactional
 	public ObjectResponse<ExperimentalCondition> create(ExperimentalCondition uiEntity) {
 		ExperimentalCondition dbEntity = experimentalConditionValidator.validateExperimentalConditionCreate(uiEntity);
-		return new ObjectResponse<>(experimentalConditionDAO.persist(dbEntity));
+			 return new ObjectResponse<>(experimentalConditionDAO.persist(dbEntity));
```
oblodgett: Long nights, low sleep, extra tab and space... seems reasonable.

oblodgett (commented):
Ok took my most of the day to get through this PR. I have not looked at the migration yet. I am going to pull this branch and give it a shot. Maybe can speed up the migration... will run again current version of alpha.

src/main/java/org/alliancegenome/curation_api/interfaces/crud/AffectedGenomicModelCrudInterface.java
oblodgett: After writing this I think this is used to lookup "curie", or mod ids? I made a comment I guess further down about putting this into its own Interface.

src/main/resources/db/migration/v0.30.0.1__linkml_v2.0.0_update.sql
```diff
@@ -0,0 +1,2574 @@
...
+CREATE TABLE submittedobject_aud (
+	id bigint NOT NULL,
+	modentityid varchar(255),
+	modinternalid varchar(255),
+	dataprovider_id bigint,
+	rev integer NOT NULL,
+	PRIMARY KEY (id, rev)
+	);
+	
+CREATE SEQUENCE auditedobject_seq START WITH 200250000 INCREMENT BY 50 NO MINVALUE NO MAXVALUE CACHE 1;
```
oblodgett: For this migration we might want to set the increment to 1 and change it back to 50 when we finish. Hibernate caches 50 ids so that it doesn't have to call the seq all the time. If we don't I think we will have huge gaps in ids here. Not sure yet... going to run this on my system now.

src/main/resources/db/migration/v0.30.0.1__linkml_v2.0.0_update.sql
```diff
@@ -0,0 +1,2574 @@
...
+ALTER TABLE xsmoterm DROP CONSTRAINT fk36v85u2ghsukqljkkh1jxmtlx;
+ALTER TABLE xsmoterm_aud DROP CONSTRAINT fkjdlhxyvw79i14932dqdb1bx8f;
+ALTER TABLE zecoterm DROP CONSTRAINT fkqd3f6hcopl67fwai6viq07t88;
+ALTER TABLE zecoterm_aud DROP CONSTRAINT fke5wuchgyjhb2orgvht50q2dah;
+ALTER TABLE zfaterm DROP CONSTRAINT fk572s3xiqi0y4gjblq8xjyk3f7;
+ALTER TABLE zfaterm_aud DROP CONSTRAINT fks66s1k4fon0to2kk7qfsm1xon;
+ALTER TABLE zfsterm DROP CONSTRAINT fknkrcnh3l23ol126v9w61nk079;
+ALTER TABLE zfsterm_aud DROP CONSTRAINT fk27dwwh5ekpa60ug4ystuwwn6w;
+
+ALTER TABLE affectedgenomicmodel DROP CONSTRAINT affectedgenomicmodel_pkey;
```
oblodgett, line 434: Dropping the primary keys does not help with the later query of: `UPDATE affectedgenomicmodel t SET id = a.id FROM auditedobject a WHERE t.curie = a.old_curie AND a.tablename = 'biologicalentity';` as t.curie now has no index.

src/main/resources/db/migration/v0.30.0.1__linkml_v2.0.0_update.sql
```diff
@@ -0,0 +1,2574 @@
...
+UPDATE zfaterm t SET id = a.id FROM auditedobject a WHERE t.curie = a.old_curie AND a.tablename = 'ontologyterm';
+UPDATE zfsterm t SET id = a.id FROM auditedobject a WHERE t.curie = a.old_curie AND a.tablename = 'ontologyterm';
+
+INSERT INTO submittedobject (id, modentityid, dataprovider_id) SELECT id, curie, dataprovider_id FROM biologicalentity;
+INSERT INTO submittedobject (id, modentityid, modinternalid, dataprovider_id) SELECT id, modentityid, modinternalid, dataprovider_id FROM reagent;
+
+INSERT INTO curieobject (id) SELECT id FROM submittedobject;
+INSERT INTO curieobject (id, curie) SELECT id, curie FROM ontologyterm;
+INSERT INTO curieobject (id, curie) SELECT id, curie FROM informationcontententity;
+
```
oblodgett, line 877: I would guess you could remove the pkeys here or just let them drop when the column gets dropped.

src/main/resources/db/migration/v0.30.0.1__linkml_v2.0.0_update.sql
```diff
@@ -0,0 +1,2574 @@
...
+UPDATE agmdiseaseannotation t SET inferredallele_id = a.id FROM auditedobject a WHERE t.inferredallele_curie = a.old_curie AND a.tablename = 'biologicalentity';
+UPDATE agmdiseaseannotation t SET assertedallele_id = a.id FROM auditedobject a WHERE t.assertedallele_curie = a.old_curie AND a.tablename = 'biologicalentity';
+UPDATE agmdiseaseannotation t SET inferredgene_id = a.id FROM auditedobject a WHERE t.inferredgene_curie = a.old_curie AND a.tablename = 'biologicalentity';
+UPDATE agmdiseaseannotation t SET subject_id = a.id FROM auditedobject a WHERE t.subject_curie = a.old_curie AND a.tablename = 'biologicalentity';
+UPDATE agmdiseaseannotation_gene t SET agmdiseaseannotation_id = a.id FROM auditedobject a WHERE t.agmdiseaseannotation_id = a.old_id AND a.tablename = 'association';
+UPDATE agmdiseaseannotation_gene t SET assertedgenes_id = a.id FROM auditedobject a WHERE t.assertedgenes_curie = a.old_curie AND a.tablename = 'biologicalentity';
+UPDATE allele t SET id = a.id FROM auditedobject a WHERE t.curie = a.old_curie AND a.tablename = 'biologicalentity';
```
oblodgett: Ya so playing around with this on my system it looks like the index / pkey doesn't help:
```
UPDATE allele t SET id = a.id FROM auditedobject a WHERE t.curie = a.old_curie AND a.tablename = 'biologicalentity';
-- index t.curie:  1:49
-- pkey  t.curie:  1:27
-- none  t.curie:  1:00
```

src/main/java/org/alliancegenome/curation_api/interfaces/crud/ReferenceCrudInterface.java
markquintontulloch: Done

src/main/java/org/alliancegenome/curation_api/jobs/executors/AgmDiseaseAnnotationExecutor.java
markquintontulloch: Have just deleted these - they were only ever intended for my debugging locally

src/main/java/org/alliancegenome/curation_api/interfaces/crud/AffectedGenomicModelCrudInterface.java
markquintontulloch: Yep, could be AGR curie, MOD curie or MOD internal ID

src/main/java/org/alliancegenome/curation_api/model/entities/Allele.java
markquintontulloch: I have no idea why I added this! Removed now

src/main/java/org/alliancegenome/curation_api/services/ontology/NcbiTaxonTermService.java
markquintontulloch: Done

src/main/java/org/alliancegenome/curation_api/services/base/CurieObjectCrudService.java
markquintontulloch: I thought this kind of logic was better kept in the service classes rather than DAO.  Also, I only wanted it accessible by the classes that inherit from CurieObject

src/main/java/org/alliancegenome/curation_api/services/OrganizationService.java
markquintontulloch: orgCacheMap can also be keyed by abbreviation (String), so I kept it like this.  Could refactor though

src/main/java/org/alliancegenome/curation_api/services/ExperimentalConditionService.java
markquintontulloch: Fixed

markquintontulloch:
> Ok took my most of the day to get through this PR. I have not looked at the migration yet. I am going to pull this branch and give it a shot. Maybe can speed up the migration... will run again current version of alpha.

@oblodgett If you're going to run against alpha then make sure you pull in my latest changes.  I was getting a PK error on the inserts into auditedobject_aud as there is a clash between IDs that hadn't been updated in the audit tables due to those IDs no longer being present in the (non-audit) tables, and therefore not having had new IDs assigned.

src/main/resources/db/migration/v0.30.0.1__linkml_v2.0.0_update.sql
markquintontulloch: Done

src/main/java/org/alliancegenome/curation_api/model/ingest/dto/GeneDiseaseAnnotationDTO.java
markquintontulloch: yep

src/main/java/org/alliancegenome/curation_api/model/entities/GenomicEntity.java
markquintontulloch: We've probably had this conversation before, but I think this needs to be OneToMany for orphanRemoval to work - otherwise we need to manually delete xrefs when we delete a GenomicEntity

markquintontulloch:
@oblodgett ES indexes are smaller than those on alpha, so that doesn't explain slow performance on my machine.  Might be good if you could check this locally to see how performance is for you.  I can create a dump of the migrated DB again (changed since last one).

oblodgett:
ya if you can create the dump file... i'll give it a shot.

PR #1379, opened 2024-01-08
SCRUM-3511: bring production to beta

abecerra:


oblodgett:
What needs to happen is a branch is taken from production called release/v0.27.4 and that branch is merged back to beta, and then deal with the conflicts in that branch.

PR #1387, opened 2024-01-08
SCRUM-3511: deployment to production  

abecerra:


oblodgett (approved):

oblodgett (changes requested):
You need to create a branch v0.28.0 from beta and not use this one.

PR #1391, opened 2024-01-09
Made modifications to the dataloads components and dataload service

VarunReddy1111:


oblodgett (changes requested):
node_modules should not be included in the PR

oblodgett (changes requested):
package-lock.json and package.json need to get removed also

adamgibs:
@oblodgett The package.json and package-lock.json are needed to build the node_modules.

oblodgett:
@adamgibs not in the root of the project

adamgibs:
@oblodgett Ah, didn't catch that. Yeah, they shouldn't be there.

package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "file-saver": "^2.0.5"
```
oblodgett: @adamgibs @VarunReddy1111 this file needs to be removed.

package-lock.json
```diff
@@ -0,0 +1,17 @@
+{
```
oblodgett: @adamgibs @VarunReddy1111 this file needs to be removed.

adamgibs (approved):

oblodgett (approved):

PR #1393, opened 2024-01-10
SCRUM-3605 serialize sgdStrainBackground, add to ForPublic view

cmpich:


src/main/java/org/alliancegenome/curation_api/model/entities/GeneDiseaseAnnotation.java
```diff
@@ -46,7 +46,7 @@ public class GeneDiseaseAnnotation extends DiseaseAnnotation {
 	@IndexedEmbedded(includeDepth = 1)
 	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
 	@ManyToOne
-	@JsonView({ View.FieldsOnly.class })
+	@JsonView({ View.FieldsOnly.class, View.ForPublic.class })
```
devopsagr, line 49: @cmpich this is not going to work unless its on production.

devopsagr (changes requested):

src/main/java/org/alliancegenome/curation_api/model/entities/GeneDiseaseAnnotation.java
devopsagr: Sorry logged in at the wrong user. (oblodgett)

oblodgett, line 49: @cmpich this is not going to work unless its on production.

oblodgett (changes requested):

cmpich:
I also will create a new release branch with this and a corresponding PR for production. Will still need this change on alpha, though.

oblodgett:
Changes put in production get rolled back to alpha there is no need for this PR then.

PR #1395, opened 2024-01-11
Fix JsonMappingException for allele/updateDetail endpoint

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/entities/InformationContentEntity.java
```diff
@@ -23,6 +24,7 @@
 @Data
 @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
 @ToString(callSuper = true)
+@JsonIgnoreProperties(ignoreUnknown = true)
```
oblodgett, line 27: This should be the default for the API. If you look in the `RestDefaultObjectMapper` you will see `mapper.configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);` if that is not working then we need to investigate as to why and not just override here.

src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
```diff
@@ -49,7 +51,8 @@ public class Reference extends InformationContentEntity {
 
 	@IndexedEmbedded(includeDepth = 1)
 	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
-	@ManyToMany
+	@OneToMany
```
oblodgett: The reason for this change as effectively onetomany and manytomany here are going to be the same it should stay as a manytomany as that describes the relationship between the two tables.

oblodgett: If you are trying to make sure that the cross refs are always loaded then put a fetch = EAGER on the @ManyToMany

markquintontulloch: It's not many-to-many though - we don't reuse/share xrefs.

oblodgett: yes that may be the case, but the relationship doesn't change, its more of a policy in the code. if it was truly a oneto many the a referenceId would need to be on the cross reference and its not so from a object to object point of view a cross reference can be on many data types and a data type can have many cross references we just happen to enforce a single entry in the list, so relationship wise in the db its a manytomany policy wise in the software in the software its a onetomany. The other way to enforce this is to put a unique constraint on the crossrefid column in the join table

src/main/java/org/alliancegenome/curation_api/model/entities/InformationContentEntity.java
markquintontulloch: I get multiple failures in the integration tests without this, e.g.
`com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: Unrecognized field "crossReferences" (class org.alliancegenome.curation_api.model.entities.InformationContentEntity), not marked as ignorable (9 known properties: "updatedBy", "obsolete", "dateCreated", "dbDateCreated", "dateUpdated", "createdBy", "internal", "dbDateUpdated", "curie"])
 at [Source: (String)"{"entity":{"type":"Gene","createdBy":{"internal":false,"obsolete":false,"dbDateCreated":"2024-01-11T12:58:36.965568Z","dbDateUpdated":"2024-01-11T12:58:36.965638Z","id":200000000,"uniqueId":"Local|Dev User|test@alliancegenome.org","firstName":"Local","lastName":"Dev User"},"updatedBy":{"internal":false,"obsolete":false,"dbDateCreated":"2024-01-11T12:58:36.965568Z","dbDateUpdated":"2024-01-11T12:58:36.965638Z","id":200000000,"uniqueId":"Local|Dev User|test@alliancegenome.org","firstName":"Local","[truncated 8557 chars]; line: 1, column: 3033] (through reference chain: org.alliancegenome.curation_api.response.ObjectResponse["entity"]->org.alliancegenome.curation_api.model.entities.Gene["geneSymbol"]->org.alliancegenome.curation_api.model.entities.slotAnnotations.geneSlotAnnotations.GeneSymbolSlotAnnotation["evidence"]->java.util.ArrayList[0]->org.alliancegenome.curation_api.model.entities.InformationContentEntity["crossReferences"])
	at org.alliancegenome.curation_api.crud.controllers.GeneITCase.editGene(GeneITCase.java:192)
Caused by: com.fasterxml.jackson.databind.exc.UnrecognizedPropertyException: 
Unrecognized field "crossReferences" (class org.alliancegenome.curation_api.model.entities.InformationContentEntity), not marked as ignorable (9 known properties: "updatedBy", "obsolete", "dateCreated", "dbDateCreated", "dateUpdated", "createdBy", "internal", "dbDateUpdated", "curie"])
 at [Source: (String)"{"entity":{"type":"Gene","createdBy":{"internal":false,"obsolete":false,"dbDateCreated":"2024-01-11T12:58:36.965568Z","dbDateUpdated":"2024-01-11T12:58:36.965638Z","id":200000000,"uniqueId":"Local|Dev User|test@alliancegenome.org","firstName":"Local","lastName":"Dev User"},"updatedBy":{"internal":false,"obsolete":false,"dbDateCreated":"2024-01-11T12:58:36.965568Z","dbDateUpdated":"2024-01-11T12:58:36.965638Z","id":200000000,"uniqueId":"Local|Dev User|test@alliancegenome.org","firstName":"Local","[truncated 8557 chars]; line: 1, column: 3033] (through reference chain: org.alliancegenome.curation_api.response.ObjectResponse["entity"]->org.alliancegenome.curation_api.model.entities.Gene["geneSymbol"]->org.alliancegenome.curation_api.model.entities.slotAnnotations.geneSlotAnnotations.GeneSymbolSlotAnnotation["evidence"]->java.util.ArrayList[0]->org.alliancegenome.curation_api.model.entities.InformationContentEntity["crossReferences"])
	at org.alliancegenome.curation_api.crud.controllers.GeneITCase.editGene(GeneITCase.java:192)`

src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
oblodgett: but the mantomany tells hibernate to build a query through the join table

src/main/java/org/alliancegenome/curation_api/model/entities/InformationContentEntity.java
oblodgett: so the api is trying to create a infocontent but what is getting passed in is a reference? there for the unknown fields?

markquintontulloch: Yeah

src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
markquintontulloch: Ok, makes sense - will revert to @ManyToMany

src/main/java/org/alliancegenome/curation_api/model/entities/InformationContentEntity.java
oblodgett: i am wondering if we should change the model to represent this? because this is like sending a gene to a biologicalentity it doesn't know about the specific gene field. plus i also maintain that this should work and not fail

markquintontulloch: I think the whole evidence part of the model needs a rethink.  There was some discussion about it a while ago but I forget the conclusion.  I'd be in favour of ditching InformationContentEntity and just using Reference directly, then using additional slots for other evidence types where necessary.

oblodgett: yes that seems most reasonable. however it would seem that the IT's are not using the default objectmapper if we are getting this error, which will cause us more problems in the future, so we need to sort out why its not being used

oblodgett: is the error happening on the API side or client side when the integration tests run?

markquintontulloch: Could that also explain the other issue I'm seeing with the allele/updateDetail endpoint?  Doing an update with alleleGeneAssociations doesn't delete the associations, yet the integration test that does exactly the same API calls works.

markquintontulloch: API side

oblodgett: Ya that could be... I think we could put some code in the initialization of the objectmapper to check that its being created

markquintontulloch: It does seem to be - I turned on the logging message in RestDefaultObjectMapper() and it showed up in the IT logs

chris-grove (approved):

PR #1410, opened 2024-01-22
SCRUM-3621/SCRUM-3637

adamgibs:


src/main/java/org/alliancegenome/curation_api/services/validation/AlleleValidator.java
```diff
@@ -621,14 +619,37 @@ private List<AlleleFunctionalImpactSlotAnnotation> validateAlleleFunctionalImpac
...
 
+		List<AlleleGeneAssociation> uiAssociations = uiEntity.getAlleleGeneAssociations();
+		List<AlleleGeneAssociation> dbAssociations = dbEntity.getAlleleGeneAssociations();
+
+		Set<AlleleGeneAssociation> associationsToDelete = new HashSet<>();
+
+		if (CollectionUtils.isNotEmpty(dbAssociations)) {
+			for (AlleleGeneAssociation dbAssociation : dbAssociations) {
+				Long dbAssociationId = dbAssociation.getId();
+				if (!containsAssociation(uiAssociations, dbAssociationId)) {
```
oblodgett: So instead of looping the whole set again and again in the containsAssociation it would be better to create a set and use that for lookup.

adamgibs: Do you mean like `uiAssociations.contains(dbAssociation)`?

src/main/java/org/alliancegenome/curation_api/services/validation/AlleleValidator.java
```diff
@@ -640,6 +661,16 @@ private List<AlleleGeneAssociation> validateAlleleGeneAssociations(Allele uiEnti
 				}
 			}
 		}
+		if (CollectionUtils.isNotEmpty(dbAssociations)) {
+			for (int ix = 0; ix < dbAssociations.size(); ix++) {
+				AlleleGeneAssociation ga = dbAssociations.get(ix);
+				if(associationsToDelete.contains(ga)){
```
oblodgett: This contains is doing a object compare and you are not guaranteed that two identical data contain objects will be equal they may have two different object ids but yet the same object and this contains will always be false.

oblodgett (changes requested):
We can go over this it you would like.

src/main/java/org/alliancegenome/curation_api/services/validation/AlleleValidator.java
adamgibs: Yeah, that's what I did at first. For some reason using the `uiAssociations.contains(dbAssociation)` up top was always returning null for me.

adamgibs:
> We can go over this it you would like.

Yeah, sure

oblodgett (approved):

PR #1414, opened 2024-01-23
Mergeback release v0.28.3 to alpha

mluypaert:
Autogenerated through GH-actions.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/constants/EntityFieldConstants.java
+++ b/src/main/java/org/alliancegenome/curation_api/constants/EntityFieldConstants.java
@@ -3,6 +3,7 @@ package org.alliancegenome.curation_api.constants;
 public final class EntityFieldConstants {
 
 	public static final String DATA_PROVIDER = "dataProvider.sourceOrganization.abbreviation";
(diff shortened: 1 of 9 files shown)


oblodgett:
@mluypaert why does this have a no-deploy on it we want to deploy this to alpha

mluypaert:
> @mluypaert why does this have a no-deploy on it we want to deploy this to alpha

@oblodgett Most mergebacks are for code that went from alpha to beta and production so does not need (re)deployment to alpha, so the `no-deploy` label is automatically attached. Patch releases are an exception on this, so you can simply remove the `no-deploy` from this PR before approval and merge to trigger alpha deployment in those cases.

oblodgett:
Which forces devs to read some documentation somewhere instead of this just being intuitive.

oblodgett (approved):

mluypaert:
> Which forces devs to read some documentation somewhere instead of this just being intuitive.

Well you can always update the release action that creates these mergeback PRs if you prefer an automated approach.

```
if $RELEASE =~ /^x.y.z$/ and z != 0
    deploy
else
    no-deploy
```

PR #1424, opened 2024-02-05
Changed the fields first alphabet to caps

VarunReddy1111:


oblodgett:
@VarunReddy1111 so the branch name here should be changed. Given that its two stories in one, the branch name shouldn't be a SCRUM branch rather `fix_typos` and just put both ticket numbers in the description / name.

oblodgett (changes requested):
Is this associated with these two stories? If not you shouldn't include them as the branch name. In jira it makes it look like these are both being built.

VarunReddy1111:
Thank you for the review @oblodgett . Will make the above mentioned changes

PR #1433, opened 2024-02-19
SCRUM-3661 Persistent Settings of Disease Annotation's - new annotation pop up form

VarunReddy1111:


src/main/cliapp/src/service/useGetUserSettings.js
```diff
@@ -10,7 +10,7 @@ export const useGetUserSettings = (key, defaultValue) => {
 	const [settings, setSettings] = useState(() => {
 		const stickyValue = localStorage.getItem(key);
 		const stickyObject = stickyValue !== null ? JSON.parse(stickyValue) : defaultValue;
-		if(key !== 'themeSettings') {
+		if(key !== 'themeSettings' && key !== 'DiseaseAnnotationsFormSettings') {
```
adamgibs: Maybe we should pass down a boolean like `isTable` and default it to false and then in places where this method is being used for a table, pass it in as true. Otherwise, this if check is just going to keep growing.

VarunReddy1111: Sure Adam, that's a good approach. Will modify as mentioned.

adamgibs (approved):

PR #1435, opened 2024-02-20
SCRUM-3515 LinkML v2.2.0 migration

markquintontulloch:


src/main/cliapp/src/containers/diseaseAnnotationsPage/DiseaseAnnotationsTable.js
```diff
@@ -503,19 +506,19 @@ export const DiseaseAnnotationsTable = () => {
 
 	const onConditionRelationHandleEditorValueChange = (props, event) => {
 		let updatedAnnotations = [...props.props.value];
-		if (typeof event.value === "object") {
+		if (typeof event.value === "diseaseAnnotationObject") {
```
adamgibs: This is one of the lines that's causing the error. Looks like a find and replace error. Should check for a JavaScript "object" type.

src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js
```diff
@@ -322,7 +324,7 @@ export const NewAnnotationForm = ({
 	const isExperimentEnabled = () => {
 		return (
 			//only enabled if a reference is selected from suggestions and condition relation table isn't visible
-			typeof newAnnotation.singleReference === "object"
+			typeof newAnnotation.singleReference === "diseaseAnnotationObject"
```
adamgibs: This is one of the lines that's causing the error. Looks like a find and replace error. Should check for a JavaScript "object" type.

src/main/cliapp/src/containers/diseaseAnnotationsPage/DiseaseAnnotationsTable.js
markquintontulloch: Got a bit find-and-replace happy! Fixed now

src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js
markquintontulloch: Also fixed now

src/main/java/org/alliancegenome/curation_api/model/entities/base/AuditedObject.java
```diff
@@ -17,27 +19,42 @@
...
 @Data
 @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
-@MappedSuperclass
 @ToString(exclude = { "createdBy", "updatedBy" })
 @AGRCurationSchemaVersion(min = "1.2.0", max = LinkMLSchemaConstants.LATEST_RELEASE)
+@MappedSuperclass
+@Schema(name = "AuditedObject", description = "POJO that represents the AuditedObject")
 public class AuditedObject extends BaseEntity {
 
+	@Id
```
oblodgett, line 50: @markquintontulloch so is this going to be the base class for everything? As in all objects will be getting a "id" field which will be the primary key across the whole DB? Just that AuditedObject will be a MappedSuper vs table.

src/main/java/org/alliancegenome/curation_api/model/entities/base/SubmittedObject.java
```diff
@@ -0,0 +1,79 @@
...
+import com.fasterxml.jackson.annotation.JsonView;
+
+import jakarta.persistence.ManyToOne;
+import jakarta.persistence.MappedSuperclass;
+import jakarta.persistence.Transient;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
+import lombok.ToString;
+
+@MappedSuperclass
```
oblodgett, line 30: So this being a mapped super will mean that the fields of AO, CO, and SO are all down on the child classes.

markquintontulloch: Yeah, that’s intended - too slow otherwise

src/main/java/org/alliancegenome/curation_api/model/entities/base/AuditedObject.java
markquintontulloch: Yeah. We need IDs on the biological entities as some don’t have curies or MOD IDs.  Ontology terms don’t really need them, but I think this gives us more flexibility for the future

src/main/java/org/alliancegenome/curation_api/model/entities/GenomicEntity.java
```diff
@@ -36,20 +30,21 @@ public class GenomicEntity extends BiologicalEntity {
 
 	@IndexedEmbedded(includePaths = {"referencedCurie", "displayName", "referencedCurie_keyword", "displayName_keyword"})
 	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
-	@ManyToMany
+	@OneToMany(cascade = CascadeType.ALL, orphanRemoval=true)
```
oblodgett, line 33: So this relationship type should be ManyToMany given that there is a join table. Which does not support the orphanRemoval. So typically what would be done here is another class is created: GenomicEntityCrossReference which represents the entity in the join table. Then this class has a `List<GenomicEntityCrossReference>` with a OneToMany relationship.

adamgibs:
@markquintontulloch As far as I can tell, the UI code looks good now. I'll let Olin go ahead and approve it when he's ready.

src/main/java/org/alliancegenome/curation_api/dao/AffectedGenomicModelDAO.java
```diff
@@ -16,13 +17,16 @@ protected AffectedGenomicModelDAO() {
...
-	public List<Long> findReferencingDiseaseAnnotations(String agmCurie) {
-		Query jpqlQuery = entityManager.createQuery("SELECT ada.id FROM AGMDiseaseAnnotation ada WHERE ada.subject.curie = :agmCurie");
-		jpqlQuery.setParameter("agmCurie", agmCurie);
-		List<Long> results = (List<Long>) jpqlQuery.getResultList();
+	public List<Long> findReferencingDiseaseAnnotations(Long agmId) {
+		List<Long> results = new ArrayList<>();
 		
-		jpqlQuery = entityManager.createNativeQuery("SELECT diseaseannotation_id FROM diseaseannotation_biologicalentity db WHERE diseasegeneticmodifiers_curie = :agmCurie");
-		jpqlQuery.setParameter("agmCurie", agmCurie);
+		Query jpqlQuery = entityManager.createQuery("SELECT ada.id FROM AGMDiseaseAnnotation ada WHERE ada.subjectBiologicalEntity.id = :agmId");
```
oblodgett: Does a findByParam not work here? AGMDiseaseAnnotationDAO.findByParams()
map.put("subjectBiologicalEntity.id", value);

oblodgett: Or are you only wanting id's back? We need to create a method with the same interface that returns ids vs full objects.

src/main/java/org/alliancegenome/curation_api/jobs/util/SlackNotifier.java
```diff
@@ -22,7 +22,7 @@
 @ApplicationScoped
 public class SlackNotifier {
 	
-	@ConfigProperty(name = "net")
+	@ConfigProperty(name = "net", defaultValue="\"\"")
```
oblodgett, line 25: I am guessing this is a change that came in from another PR? I think this should be `, defaultValue=""`

oblodgett: Or just add `NET = dev` in your application.properties file.

src/main/java/org/alliancegenome/curation_api/model/entities/AGMDiseaseAnnotation.java
```diff
@@ -21,59 +20,88 @@
...
 
-	@IndexedEmbedded(includeDepth = 2)
+	@IndexedEmbedded(includePaths = {
+			"curie", "modEntityId", "modInternalId", "curie_keyword", "modEntityId_keyword", "modInternalId_keyword",
+			"geneSymbol.formatText", "geneSymbol.displayText", "geneSymbol.formatText_keyword", "geneSymbol.displayText_keyword",
+			"geneFullName.formatText", "geneFullName.displayText", "geneFullName.formatText_keyword", "geneFullName.displayText_keyword",
+			"geneSystematicName.formatText", "geneSystematicName.displayText", "geneSystematicName.formatText_keyword", "geneSystematicName.displayText_keyword",
+			"geneSynonyms.formatText", "geneSynonyms.displayText", "geneSynonyms.formatText_keyword", "geneSynonyms.displayText_keyword",
+			"geneSecondaryIds.secondaryId", "geneSecondaryIds.secondaryId_keyword", "name", "name_keyword", "symbol", "symbol_keyword"
+	})
```
oblodgett, line 63: @cmpich this is the code I am referring to that will drop out the constructs.

src/main/java/org/alliancegenome/curation_api/services/associations/alleleAssociations/AlleleGeneAssociationService.java
```diff
@@ -47,10 +49,14 @@ public class AlleleGeneAssociationService extends BaseAssociationDTOCrudService<
 	@Inject
 	AlleleDAO alleleDAO;
 	@Inject
+	AlleleService alleleService;
+	@Inject
 	NoteDAO noteDAO;
 	@Inject
 	GeneDAO geneDAO;
 	@Inject
+	GeneService geneService;
```
oblodgett: It doesn't look like the gene or allele services are getting used?

src/main/java/org/alliancegenome/curation_api/jobs/util/SlackNotifier.java
markquintontulloch: it was put in to avoid endlessly looping errors when developing locally - it's to avoid entering the block of code that starts with `if(!systemName.get().equals("\"\"") && !slackToken.get().equals("\"\"")) {`
Does adding `NET=dev` do that?

src/main/java/org/alliancegenome/curation_api/model/entities/GenomicEntity.java
markquintontulloch: Would that not be a departure from the LinkML model though?  So should we just go back to manual cleanup of xrefs?

src/main/resources/db/migration/v0.31.0.1__linkml_v2.1.0_update.sql
```diff
@@ -0,0 +1,1173 @@
...
+ALTER TABLE organization DROP COLUMN IF EXISTS uniqueid;
+
+DROP TABLE IF EXISTS paperhandle;
+DROP TABLE IF EXISTS paperhandle_aud;
+
+ALTER TABLE agmdiseaseannotation DROP COLUMN IF EXISTS predicate;
+ALTER TABLE allelediseaseannotation DROP COLUMN IF EXISTS predicate;
+ALTER TABLE genediseaseannotation DROP COLUMN IF EXISTS predicate;
+
+ALTER TABLE synonym ALTER COLUMN name SET DATA TYPE text;
```
oblodgett: I think I already have this change in via an earlier commit but probably not an issue.

src/main/resources/db/migration/v0.31.0.1__linkml_v2.1.0_update.sql
```diff
@@ -0,0 +1,1173 @@
...
+ALTER TABLE uberonterm ADD CONSTRAINT uberonterm_pkey PRIMARY KEY (id);
+ALTER TABLE variant ADD CONSTRAINT variant_pkey PRIMARY KEY (id);
+ALTER TABLE vtterm ADD CONSTRAINT vtterm_pkey PRIMARY KEY (id);
+ALTER TABLE wbbtterm ADD CONSTRAINT wbbtterm_pkey PRIMARY KEY (id);
+ALTER TABLE wblsterm ADD CONSTRAINT wblsterm_pkey PRIMARY KEY (id);
+ALTER TABLE wbphenotypeterm ADD CONSTRAINT wbphenotypeterm_pkey PRIMARY KEY (id);
+ALTER TABLE xbaterm ADD CONSTRAINT xbaterm_pkey PRIMARY KEY (id);
+ALTER TABLE xbedterm ADD CONSTRAINT xbedterm_pkey PRIMARY KEY (id);
+ALTER TABLE xbsterm ADD CONSTRAINT xbsterm_pkey PRIMARY KEY (id);
+ALTER TABLE xcoterm ADD CONSTRAINT xcoterm_pkey PRIMARY KEY (id);
```
oblodgett, line 971: Wondering if these should be "pk" vs "pkey"

oblodgett (commented):
How long does the migration take?

src/main/java/org/alliancegenome/curation_api/dao/AffectedGenomicModelDAO.java
markquintontulloch: IDs only, but we already have that method (findFilteredIds) - will replace this and similar in other DAOs

markquintontulloch:
> How long does the migration take?

I think it was around 20 mins locally

src/main/java/org/alliancegenome/curation_api/services/associations/alleleAssociations/AlleleGeneAssociationService.java
markquintontulloch: Good spot - will remove

src/main/java/org/alliancegenome/curation_api/model/entities/GenomicEntity.java
oblodgett: It jsut that via hibernate a OneToMany would imply that there is a table called CrossReference that has a GenomeEntityFK back to this table... that is not true in that there is a join table in between the two tables which really should be considered as a ManyToMany relationship. This works, but is not the ideal hibernate way of doing it.

src/main/java/org/alliancegenome/curation_api/jobs/util/SlackNotifier.java
oblodgett: Well the NET=dev sets it to a value as I would guess maybe your file is missing it?

oblodgett:
> > How long does the migration take?
> 
> I think it was around 20 mins locally

NICE!!!

src/main/java/org/alliancegenome/curation_api/dao/AffectedGenomicModelDAO.java
oblodgett: Ya not a big deal just trying to use the least amount of JQL as possible.

src/main/resources/db/migration/v0.31.0.1__linkml_v2.1.0_update.sql
markquintontulloch: "pkey" is the default suffix that's used on all our primary keys where we don't specify a name in the table definition

src/main/resources/db/migration/v0.31.0.1__linkml_v2.1.0_update.sql
markquintontulloch: yeah, will remove

markquintontulloch:
That may be optimistic remembering, but it certainly wasn't hours!

src/main/resources/db/migration/v0.31.0.1__linkml_v2.1.0_update.sql
oblodgett: Oh ok... I thought it was "pk" as the others are "uk" and "fk" but whatever.

oblodgett (approved):

PR #1443, opened 2024-02-27
SCRUM-3770 Speed up ID retrieval for filtered ID lists

markquintontulloch:


oblodgett (approved):
This is some crazy code... good job on breaking it out.

PR #1444, opened 2024-02-27
SCRUM 3694 Adding the GENO Term ontology into persistent store

VarunReddy1111:


VarunReddy1111:
Few changes need to be done in UI. Will modify and create another one

oblodgett (approved):

PR #1459, opened 2024-03-13
SCRUM-3794

adamgibs:


src/main/cliapp/src/components/Editors/NotEditor.js
```diff
@@ -0,0 +1,31 @@
+import React, { useState } from 'react';
+import { Dropdown } from "primereact/dropdown";
+
+export function NotEditor({ props, value, editorChange }) {
+	const [selectedValue, setSelectedValue] = useState(value);
+	const textString = value ? "NOT" : "";
+	const options = [
+		{ label: "NOT", value: true },
+		{ label: "null", value: false }
```
oblodgett: So the idea is in the pull down you will either select "NOT" or you will not select anything at all. In other words there is 1 option in the pull down its either selected or its not. If its selected the it will show "NOT". And on the back end if the field is undefined then its set to false. If its true then "NOT" will be displayed.

adamgibs: Okay, sure

src/main/cliapp/src/constants/FilterFields.js
```diff
@@ -594,7 +594,7 @@ export const FILTER_CONFIGS = Object.freeze({
   isExtinctFilterConfig:                    { filterComponentType: "dropdown", fieldSets: [FIELD_SETS.isExtinctFieldSet] },
   obsoleteFilterConfig:                     { filterComponentType: "dropdown", fieldSets: [FIELD_SETS.obsoleteFieldSet] },
   internalFilterConfig:                     { filterComponentType: "dropdown", fieldSets: [FIELD_SETS.internalFieldSet] },
-  negatedFilterConfig:                      { filterComponentType: "dropdown", fieldSets: [FIELD_SETS.negatedFieldSet] },
+  negatedFilterConfig:                      { filterComponentType: "dropdown", fieldSets: [FIELD_SETS.negatedFieldSet], options: [ { label: "NOT", value: "true" }, { label: "null", value: "false" } ] },
```
oblodgett, line 597: Does this need to change also?

adamgibs: This is in place with the assumption that users may want to filter for false as opposed to just having no filter, or filtering for true.

oblodgett (approved):

PR #1461, opened 2024-03-15
add getSourceUrl() convenience method to encapsulate the businesss lo…

cmpich:
…gic for generating the URL


src/main/java/org/alliancegenome/curation_api/model/entities/DataProvider.java
```diff
@@ -45,4 +47,16 @@ public class DataProvider extends AuditedObject {
 	@JsonView({ View.FieldsOnly.class, View.ForPublic.class })
 	private CrossReference crossReference;
 
+	public String getSourceUrl() {
+		List<String> urlExpceptionHandler = List.of("MGI", "SGD", "OMIM");
+		if (crossReference != null) {
+			String urlTemplate = crossReference.getResourceDescriptorPage().getUrlTemplate();
+			if (urlExpceptionHandler.contains(sourceOrganization.getAbbreviation())) {
+				// remove the prefix in the template as the prefix is already in the curie.
+				urlTemplate = urlTemplate.replace(sourceOrganization.getAbbreviation() + ":", "");
```
oblodgett, line 57: @cmpich this does not use the regex

PR #1467, opened 2024-03-20
SCRUM 2755 Mod specific fields are made required in new annotation pop up form

VarunReddy1111:


src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js
```diff
@@ -113,10 +115,11 @@ export const NewAnnotationForm = ({
 		);
 
 		areUiErrors.current = false;
+		validateModFormFields(newAnnotation, modFormFields, uiErrorMessages, setUiErrorMessages, areUiErrors);
 		validateFormBioEntityFields(newAnnotation, uiErrorMessages, setUiErrorMessages, areUiErrors);
 		if (areUiErrors.current) {
-			newAnnotationDispatch({type: "UPDATE_ERROR_MESSAGES", errorType: "errorMessages", errorMessages: uiErrorMessages});
-			newAnnotationDispatch({ type: "SET_IS_ENABLED", value: false });
+			//newAnnotationDispatch({type: "UPDATE_ERROR_MESSAGES", errorType: "errorMessages", errorMessages: uiErrorMessages});
```
adamgibs, line 121: This shouldn't be commented out. It's needed to update the store.

src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js
```diff
@@ -487,11 +491,11 @@ export const NewAnnotationForm = ({
 						</div>
 					</div>
 
-					{selectedFormFields?.includes("Asserted Genes") && (
+					{(modFormFields.includes("Asserted Genes") ? required = requiredfield : (required = '', selectedFormFields?.includes("Asserted Genes"))) && (
```
adamgibs, line 494: Not saying this should happen in this PR, but at some point these inputs need to be moved into their own components similar to the allele detail page. In part to avoid having more and more inline complex logic.

src/main/cliapp/src/containers/diseaseAnnotationsPage/NewAnnotationForm.js
```diff
@@ -113,10 +115,11 @@ export const NewAnnotationForm = ({
 		);
 
 		areUiErrors.current = false;
+		validateModFormFields(newAnnotation, modFormFields, uiErrorMessages, setUiErrorMessages, areUiErrors);
 		validateFormBioEntityFields(newAnnotation, uiErrorMessages, setUiErrorMessages, areUiErrors);
 		if (areUiErrors.current) {
-			newAnnotationDispatch({type: "UPDATE_ERROR_MESSAGES", errorType: "errorMessages", errorMessages: uiErrorMessages});
-			newAnnotationDispatch({ type: "SET_IS_ENABLED", value: false });
+			//newAnnotationDispatch({type: "UPDATE_ERROR_MESSAGES", errorType: "errorMessages", errorMessages: uiErrorMessages});
+			newAnnotationDispatch({ type: "SET_IS_ENABLED", value: true });
```
adamgibs, line 123: Also, won't changing this boolean break the logic for validateBioEntityFields? If there are uiErrors with the bioEntityFields this should not be enabled.

adamgibs (approved):

PR #1471, opened 2024-03-22
SCRUM-3844 Possible fix for failing ontology loads

markquintontulloch:
@oblodgett I think this should fix the issue of failing loads that we're seeing on beta now that the scheduled loads have been reenabled (not sure how to test scheduled stuff locally).  If you look at the logs then you'll see that the failing loads are getting kicked off multiple times within the space of a few seconds.  My theory is that the loop is taking > 1 second to run so the lastCheck time doesn't get updated before the next scheduled check, and the load gets kicked off again.


oblodgett:
I kinda assumed this bug was taken care of by fixing the ontology loads but it would appear that it is not. I think we need to debug the scheduling code, so i would suggest we turn off the scheduled load till we can figure this out. i don't mind taking this on.

PR #1486, opened 2024-04-01
SCRUM-3711 refactor

markquintontulloch:
@oblodgett There's almost certainly a cleaner way of doing this.  The complication is that not all annotation service classes inherit from the same base service class.  The AGMDiseaseAnnotationService, AlleleDiseaseAnnotationService, and GeneDiseaseAnnotationService classes all inherited from BaseDTOCrudService, whereas DiseaseAnnotationService and all the phenotype annotation service classes inherit from BaseEntityCrudService.  No matter how I try to refactor I always end up with duplicate methods somewhere or unimplemented methods in multiple classes.  There's probably something really basic that I'm missing though.


oblodgett:
@markquintontulloch not sure which it should be for this conflict.

oblodgett (approved):

oblodgett:
@markquintontulloch was this a change in the code... `org.hibernate.query.SemanticException: Could not resolve attribute 'resourceDescriptor.prefix' of 'org.alliancegenome.curation_api.model.entities.ResourceDescriptorPage'` I don't remember seeing it.

markquintontulloch:
@oblodgett No, that's been there a while.  I keep meaning to address it as presumably that means we would get a 500 error rather than a 400.

PR #1490, opened 2024-04-02
Refactor to remove some classes SCRUM-3711.2

oblodgett:


oblodgett:
@markquintontulloch this is still wrong... I will continue working on it... its due to the get(String) method in that ontology terms are curied objects where as genes are submitted objects... will need to think about this more.

oblodgett:
Ok @markquintontulloch i think this is ready. Created separate interfaces for the different operations so that submitted object can look up by identifier ontology terms by curie and notes by id. Moved most of the functionality to the base service.

markquintontulloch (approved):
It's going to take a little while to get my head around everything that's going on here, but it looks like a much neater solution for a lot of things - nice

PR #1494, opened 2024-04-03
Increase reporting interval for bulk loads

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/jobs/executors/OrthologyExecutor.java
```diff
@@ -86,7 +86,7 @@ private void runCleanup(BulkLoadFileHistory history, String dataProvider, List<O
 
 		history.setTotalDeleteRecords((long)pairsToRemove.size());
 		
-		ProcessDisplayHelper ph = new ProcessDisplayHelper(1000);
+		ProcessDisplayHelper ph = new ProcessDisplayHelper(10000);
```
oblodgett: The default here could be used as well?

oblodgett: I think the default is 30000

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/jobs/executors/OrthologyExecutor.java
markquintontulloch: Sure

PR #1498, opened 2024-04-04
Added better messaging to error

oblodgett:


markquintontulloch (approved):
This is an improvement, and I'm fine to go with this.  However, I don't think those messages will mean much to some of the DQMs - I'd be inclined to go with:
```
if (term == null) {
	ecResponse.addErrorMessage("conditionClassId", ValidationConstants.INVALID_MESSAGE + " (" + dto.getConditionClassId() + " not found)");
} else if(term.getSubsets().isEmpty() || !term.getSubsets().contains(OntologyConstants.ZECO_AGR_SLIM_SUBSET) {
	ecResponse.addErrorMessage("conditionClassId", ValidationConstants.INVALID_MESSAGE + " (" + dto.getConditionClassId() + " not part of " +  OntologyConstants.ZECO_AGR_SLIM_SUBSET + " subset)");
} else {
	experimentalCondition.setConditionClass(term);
}
```

PR #1502, opened 2024-04-06
SCRUM-3808 Implement loading of GenomicEntity xrefs from DQM submissions

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/interfaces/crud/AffectedGenomicModelCrudInterface.java
```diff
@@ -32,4 +40,37 @@ public interface AffectedGenomicModelCrudInterface extends BaseSubmittedObjectCr
 	@Path("/bulk/{dataProvider}/agms")
 	@JsonView(View.FieldsAndLists.class)
 	public APIResponse updateAGMs(@PathParam("dataProvider") String dataProvider, List<AffectedGenomicModelDTO> agmData);
+
+	@Override
+	@GET
+	@Path("/{identifierString}")
+	@JsonView(View.AffectedGenomicModelView.class)
```
oblodgett, line 47: @markquintontulloch I would be curious if we could put the `@JsonView(View.AffectedGenomicModelView.class)` at the class level and then not have to override each of the methods. Or only have to override the `updateAGMs`

markquintontulloch: @oblodgett I've had a play around with it and the answer is "I don't think so" - at least, I haven't been able to get it to work if it is possible.  I have a vague recollection of us exploring this previously and coming to the same conclusion.  It sure would be nice if we could, most of our interface classes end up with everything being overridden just because of the view.

oblodgett (approved):

PR #1523, opened 2024-04-23
SCRUM-3967 Use curie to get/delete references

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/interfaces/base/BaseCurieObjectCrudInterface.java
```diff
@@ -0,0 +1,25 @@
...
+	BaseCreateControllerInterface<E>,
+	BaseReadCurieControllerInterface<E>,
+	BaseUpdateControllerInterface<E>,
+	BaseDeleteCurieControllerInterface<E>,
+	BaseSearchControllerInterface<E>,
+	BaseFindControllerInterface<E>,
+	BaseReindexControllerInterface
+{
+
+}
```
oblodgett, line 25: I thought we had this class already... but looks like we don't.

oblodgett (approved):

PR #1526, opened 2024-04-23
SCRUM-3960 Change to make obsolete column default filtering to false …

VarunReddy1111:
…for ontologies and few other tables

### DIFF
--- a/src/main/cliapp/src/service/TableStateService.js
+++ b/src/main/cliapp/src/service/TableStateService.js
@@ -239,13 +239,17 @@ export function getModFormFields(table) {
 	return modFormFields;
 };
 
-export function getDefaultTableState(tableName, columns, defaultColumnWidth, defaultVisibleColumns) {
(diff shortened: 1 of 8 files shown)


src/main/cliapp/src/service/TableStateService.js
```diff
@@ -239,13 +239,17 @@ export function getModFormFields(table) {
 	return modFormFields;
 };
 
-export function getDefaultTableState(tableName, columns, defaultColumnWidth, defaultVisibleColumns) {
+export function getDefaultTableState(tableName, columns, defaultColumnWidth, obsoleteDefaultFilter, defaultVisibleColumns) {
+
```
adamgibs: I know this ticket is specifically for the obsolete filter, but I think it would be simpler and more flexible to allow the calling table to pass in a `defaultFilters` object and set that in the filter field on the returned object.  You can set the default value of the parameters to an empty object if one is not passed in.

adamgibs (approved):

PR #1530, opened 2024-05-01
SCRUM-3968 Compact styling of data tables along with refactoring of respective data tables

VarunReddy1111:


src/main/cliapp/src/containers/constructsPage/ConstructsTable.js
```diff
@@ -379,26 +200,40 @@ export const ConstructsTable = () => {
...
+				underline={true}
+			/>,	
 			sortable: { isInEditMode },
 			filterConfig: FILTER_CONFIGS.constructComponentsFilterConfig,
 		},
 		{
 			field: "constructGenomicEntityAssociations.constructGenomicEntityAssociationObject.symbol",
 			header: "Component Associations",
-			body: genomicComponentsTemplate,
+			body: (rowData) => <ListDialogTemplate
```
adamgibs, line 222: I think the `GenomicEntityListTemplate` could potentially be used for this? I think it has some of the logic you're using in the `getComponentsAssociationTextString` function.

VarunReddy1111: Yes Adam, this is similar logic and I've tried using this one, but this template cannot be used in this case.

adamgibs (approved):

PR #1540, opened 2024-05-23
Change update message for DTO loads

markquintontulloch:
@oblodgett Can we change this message?  "GeneService_ClientProxy updating GeneDTO" is a bit cryptic and doesn't identify the data provider


oblodgett:
Sure, however there is not always a data provider and trying to keep it generic. but ya this looks good. `objectList.size();` should be pulled out of the message and added a param to the startProcess method.

oblodgett:
plus add the objectList could to the startProcess method

oblodgett (approved):

PR #1541, opened 2024-05-23
SCRUM-4062 Set obsolete/internal to false for FMS loads

markquintontulloch:


oblodgett:
@markquintontulloch I guess the bigger question is why were these set to true in the first place... as in did the "ids" not get added to the idList and therefore got turned off? I think I will try this with a single annotation. The default on these should be false, lets hold off on the PR until we can figure out what's the issue.

markquintontulloch:
@oblodgett Pretty sure they got obsoleted due to failing gene loads

oblodgett (approved):

PR #1543, opened 2024-05-24
SCRUM 4088 - Added Controller and Interface for Paralogy

VarunReddy1111:


src/main/java/org/alliancegenome/curation_api/jobs/executors/ParalogyExecutor.java
```diff
@@ -94,5 +93,15 @@ private void runLoad(BulkLoadFileHistory history, String dataProvider, ParalogyI
 		ph.finishProcess();
 
 	}
+	
+	public APIResponse runLoad(String dataProvider, ParalogyIngestFmsDTO paralogyData) {
```
oblodgett, line 97: @VarunReddy1111 you shouldn't need this as the super class should be able to take care of it.

oblodgett, line 97: We might not need this method as we can use the super class one.... maybe we need it due to having to deal with the "Pair" of genes.

VarunReddy1111: Yes, I believe we need to have this method as we have overridden the base `runLoad` method in order to deal with "Pair" of genes.

VarunReddy1111: I can exclude this for now, but for the following tickets, I would need this method as its used by the test cases.

oblodgett: Ya... you are right sorry.

oblodgett (approved):

PR #1559, opened 2024-06-06
SCRUM-4146 Fix orthology cleanup

markquintontulloch:


oblodgett (approved):

markquintontulloch:
@oblodgett Using the generic runLoad and runLoadApi methods now.  Can't currently use any of the generic cleanUp methods because the service class doesn't inherit from the appropriate base service classes.  I think I see a way to consolidate a lot of the cleanup code for orthology and a lot of other stuff, but will do that as a separate PR as the refactor is outside of the scope of this ticket and I'd like to get SCRUM-4150 in first as that will simplify things.

PR #1568, opened 2024-06-17
SCRUM-4018 fix for migration error

adamgibs:


adamgibs:
Previous deployment failed due to migration issues. I'm not sure if this will fix it or not since I can't recreate the failures locally.

oblodgett (approved):

PR #1578, opened 2024-06-21
SCRUM-4186 Migration for OMIM to MIM switch

markquintontulloch:


oblodgett:
Shouldn't the code be able to handle this kind of update?

markquintontulloch:
I can’t think of a simple way of doing that.  The Resource Descriptor load has no way of knowing this is an update rather than a creation and deletion.  This seems like a fairly extreme edge case - one of the organisations that we have manually loaded changing its prefix.  For that reason, I’m inclined to just make the change via migration.

oblodgett (approved):

PR #1583, opened 2024-06-25
SCRUM-3708: replace modId per AGRId in in GeneExpression uniqueId

abecerra:
Solution to comment:
https://agr-jira.atlassian.net/browse/SCRUM-3708?focusedCommentId=43024

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/helpers/annotations/GeneExpressionAnnotationUniqueIdHelper.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/helpers/annotations/GeneExpressionAnnotationUniqueIdHelper.java
@@ -8,12 +8,12 @@ import org.alliancegenome.curation_api.services.helpers.UniqueIdGeneratorHelper;
 public class GeneExpressionAnnotationUniqueIdHelper {
 
 	//	UniqueID =
-	//	assayId | evidenceCrosseferenceId | geneId | stageTermId | stageName | whereExpressedStatement | anatomicalStructureTermId | cellularComponentTermId
-	public String generateUniqueId(GeneExpressionFmsDTO geneExpressionFmsDTO) {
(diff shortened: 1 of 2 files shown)


src/main/java/org/alliancegenome/curation_api/services/helpers/annotations/GeneExpressionAnnotationUniqueIdHelper.java
```diff
@@ -8,12 +8,12 @@
 public class GeneExpressionAnnotationUniqueIdHelper {
 
 	//	UniqueID =
-	//	assayId | evidenceCrosseferenceId | geneId | stageTermId | stageName | whereExpressedStatement | anatomicalStructureTermId | cellularComponentTermId
-	public String generateUniqueId(GeneExpressionFmsDTO geneExpressionFmsDTO) {
+	//	assayId | evidenceCurie | geneId | stageTermId | stageName | whereExpressedStatement | anatomicalStructureTermId | cellularComponentTermId
+	public String generateUniqueId(GeneExpressionFmsDTO geneExpressionFmsDTO, String curie) {
```
markquintontulloch: I would rename this parameter as `referenceCurie` rather than just `curie` so that devs can see exactly what this refers to without having to trace back to the method call

markquintontulloch (changes requested):

src/main/java/org/alliancegenome/curation_api/services/helpers/annotations/GeneExpressionAnnotationUniqueIdHelper.java
markquintontulloch: I guess the comment does this already, but the parameter name makes it clearer when it is used later in the method

abecerra: Renamed

markquintontulloch (approved):

PR #1584, opened 2024-06-25
SCRUM-4183

adamgibs:


markquintontulloch (changes requested):
Looks good.  One thing - your SQTR entity class will need something like:
`@ToString(exclude = { "sequenceTargetingReagentGeneAssociations" }, callSuper = true)`
You'll also want to add the new field to the corresponding annotation on the Gene entity class.  Without that, you'll get stack overflow errors.
Might want to also consider implementing a specific view for these associations so that you can override the appropriate endpoints with that view and not retrieve all the data in the FieldsAndLists.class view.  Could leave that to the ticket where you implement the interfaces though.

adamgibs:
@markquintontulloch I've updated the ToString annotations. I'll add a new View class in later ticket

markquintontulloch (approved):

PR #1590, opened 2024-07-01
Added migration to add all the category tags into vocabulary

VarunReddy1111:


markquintontulloch (changes requested):
We shouldn't be entering empty strings into join tables.  If the term doesn't have a synonym then it just wouldn't be present in the vocabularyterm_synonyms table.

src/main/resources/db/migration/v0.35.0.9__add_htp_tags.sql
```diff
@@ -0,0 +1,218 @@
...
+INSERT INTO vocabularyterm_synonyms (vocabularyterm_id, synonyms) SELECT id,'' FROM vocabularyterm WHERE name = 'cell aging';
+INSERT INTO vocabularyterm (id, name, definition, vocabulary_id) SELECT nextval('vocabularyterm_seq'), 'amino acid metabolism', 'The cellular reactions and pathways used in the biosynthesis and catabolism of amino acids.',id FROM vocabulary WHERE vocabularylabel = 'data_set _category_tags';
+INSERT INTO vocabularyterm_synonyms (vocabularyterm_id, synonyms) SELECT id,'' FROM vocabularyterm WHERE name = 'amino acid metabolism';
+INSERT INTO vocabularyterm (id, name, definition, vocabulary_id) SELECT nextval('vocabularyterm_seq'), 'amino acid utilization', 'Utilization of different amino acids, combinations of amino acids, or altered levels of amino acids (e.g. amino acid limitation) during nutritional shifts.',id FROM vocabulary WHERE vocabularylabel = 'data_set _category_tags';
```
oblodgett: is this vocabulary label correct? it has a space in it

oblodgett (changes requested):

src/main/resources/db/migration/v0.35.0.9__add_htp_tags.sql
VarunReddy1111: No, There should be no space in it as per our naming convention. It was created by Chris through UI. I will change this.

src/main/resources/db/migration/v0.35.0.9__add_htp_tags.sql
```diff
@@ -1,218 +1,129 @@
...
-INSERT INTO vocabularyterm_synonyms (vocabularyterm_id, synonyms) SELECT id,'' FROM vocabularyterm WHERE name = 'amino acid metabolism';
-INSERT INTO vocabularyterm (id, name, definition, vocabulary_id) SELECT nextval('vocabularyterm_seq'), 'amino acid utilization', 'Utilization of different amino acids, combinations of amino acids, or altered levels of amino acids (e.g. amino acid limitation) during nutritional shifts.',id FROM vocabulary WHERE vocabularylabel = 'data_set _category_tags';
-INSERT INTO vocabularyterm_synonyms (vocabularyterm_id, synonyms) SELECT id,'' FROM vocabularyterm WHERE name = 'amino acid utilization';
-INSERT INTO vocabularyterm (id, name, definition, vocabulary_id) SELECT nextval('vocabularyterm_seq'), 'anatomical structure', 'study that characterizes expression in one or more tissues',id FROM vocabulary WHERE vocabularylabel = 'data_set _category_tags';
+UPDATE vocabulary SET vocabularylabel = 'data_set_category_tags' WHERE vocabularylabel = 'data_set _category_tags';
```
markquintontulloch: How was this vocabulary created in the first place?  As far as I can tell it was manually created on alpha rather than in a migration.  If so, then this migration will fail when deployed to beta/production as the vocabulary that you are inserting terms for won't exist (unless created ahead of time).  It would be better to delete the vocabulary on alpha and recreate it here in the migration - that way no manual work is required prior to deployment to beta/production.

VarunReddy1111: Yes, It was created manually on alpha. I will delete it and create again in migration file. Thank you. I didnt think of this scenario.

markquintontulloch (approved):

oblodgett (approved):

PR #1591, opened 2024-07-02
SCRUM-4185 updated executor, service, and validator

adamgibs:
I'm able to run an SQTR FMS file and get both SQTRs and their gene associations into the database locally with this code, but I don't know if this is the best way of going about it.


src/main/java/org/alliancegenome/curation_api/jobs/executors/SequenceTargetingReagentExecutor.java
```diff
@@ -29,38 +35,107 @@ public void execLoad(BulkLoadFile bulkLoadFile) {
...
 				bulkLoadFile.setAllianceMemberReleaseVersion(sqtrIngestFmsDTO.getMetaData().getRelease());
 			}
 
 			BackendBulkDataProvider dataProvider = BackendBulkDataProvider.valueOf(fms.getFmsDataSubType());
 
 			List<Long> sqtrIdsLoaded = new ArrayList<>();
-			List<Long> sqtrIdsBefore = new ArrayList<>();
+			List<Long> sqtrIdsBefore = sqtrService.getIdsByDataProvider(dataProvider.name());
+			List<Long> sqtrGeneAssociationIdsLoaded = new ArrayList<>();
+			List<Long> sqtrGeneAssociationIdsBefore = sqtrService.getIdsByDataProvider(dataProvider.name());
```
markquintontulloch: Isn't this getting the IDs of the SQTRs rather than the associations?

markquintontulloch: I think we need to create association-specific DAO and Service classes that would be used for lookup of IDs by data provider, checking for existing associations, and handling cleanup.  The approach for the executor part looks good, but I think the rest of the stack should be a similar structure to the associations we have coming in from direct DQM loads (e.g. AlleleGeneAssociation).  You can think of it as two completely separate loads, just handled by the same executor.  First the SQTR load runs using classes in the SQTR stack, then the SQTR association load runs using classes from the SQTR association stack.

markquintontulloch (changes requested):

src/main/java/org/alliancegenome/curation_api/jobs/executors/SequenceTargetingReagentExecutor.java
adamgibs: @markquintontulloch Yeah, I wasn't sure if it made since break the association class into it's own service and validator and whatnot. I'll start working on that. Thanks!

src/main/java/org/alliancegenome/curation_api/jobs/executors/SequenceTargetingReagentExecutor.java
```diff
@@ -29,38 +35,107 @@ public void execLoad(BulkLoadFile bulkLoadFile) {
...
+
+		BulkLoadFileHistory history = new BulkLoadFileHistory(sqtrDTOs.size());
+		BackendBulkDataProvider dataProvider = BackendBulkDataProvider.valueOf(dataProviderName);
+		runLoad(history, dataProvider, sqtrDTOs, sqtrIdsLoaded, sqtrGeneAssociationIdsLoaded);
+		history.finishLoad();
+
+		return new LoadHistoryResponce(history);
+	}
+
+	private void runLoad(BulkLoadFileHistory history, BackendBulkDataProvider dataProvider, List<SequenceTargetingReagentFmsDTO> sqtrs, List<Long> sqtrIdsLoaded, List<Long> sqtrGeneAssociationIdsLoaded) {
```
oblodgett, line 104: if you can its best to use the generic methods unless these need to do some specific logic

adamgibs: @oblodgett That makes sense. In this particular case, this method is custom because it needs to call the load `loadSequenceTargetingReagents` and `loadSequenceTargetingReagentGeneAssociations` methods below.

markquintontulloch (approved):

PR #1607, opened 2024-07-11
SCRUM-4259 update

markquintontulloch:
A couple more places to ignore SQTR associations


adamgibs:
@markquintontulloch Ah, makes sense. Thanks!

adamgibs (approved):

PR #1619, opened 2024-07-17
SCRUM-4188

adamgibs:
Add integration tests for SQTR Gene Associations


oblodgett (commented):
Looks good but mark is better at the IT's than me.

markquintontulloch (approved):
Looks good to me

PR #1627, opened 2024-07-26
SCRUM-3952: add ExpressionPattern, TemporalContext and AnatomicalSite

abecerra:


oblodgett (changes requested):
You will need to rename your migration file as varun's file that just went in has the same number:
v0.36.0.6__expression_pattern.sql will need to be: v0.36.0.7__expression_pattern.sql

abecerra:
> You will need to rename your migration file as varun's file that just went in has the same number: v0.36.0.6__expression_pattern.sql will need to be: v0.36.0.7__expression_pattern.sql

File renamed and annotation ElementCollection removed

src/main/java/org/alliancegenome/curation_api/model/entities/AnatomicalSite.java
```diff
@@ -0,0 +1,116 @@
...
+@Table(indexes = {
+	@Index(name = "anatomicalsite_anatomicalstructure_index ", columnList = "anatomicalstructure_id"),
+	@Index(name = "anatomicalsite_anatomicalsubstructure_index", columnList = "anatomicalsubstructure_id"),
+	@Index(name = "anatomicalsite_cellularcomponentterm_index", columnList = "cellularcomponentterm_id")}
+)
+public class AnatomicalSite extends AuditedObject {
+	@IndexedEmbedded(includePaths = {"name", "name_keyword"})
+	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
+	@ManyToOne
+	@JsonView({ View.FieldsOnly.class, View.ForPublic.class })
```
oblodgett: I would leave the ForPublic off that is only used for the public indexer, sorry didn't catch this before.

oblodgett: We will add them back in as needed but the default should be without them.

abecerra: Removed

oblodgett (approved):

PR #1629, opened 2024-08-02
Scrum 4305

cmpich:

PR #1630, opened 2024-08-03
SCRUM-4305 set obsolete / internal variables explicitly when being up…

cmpich:
…dated. (#1629)

(cherry picked from commit bc8af6e7164f7ec3870915b904ea46e9aa1c2d2c)


oblodgett:
I think you want to go into production with this PR?

PR #1640, opened 2024-08-14
SCRUM-4300 

adamgibs:


src/main/java/org/alliancegenome/curation_api/services/base/BaseOntologyTermService.java
```diff
@@ -300,4 +300,15 @@ private void handleCrossReferences(OntologyTerm dbTerm, OntologyTerm incomingTer
 		}
 	}
 
+	public <T extends OntologyTerm> T findSubsetTerm(T childTerm, String subsetName) {
+		for (OntologyTerm term : childTerm.getIsaAncestors()) {
```
markquintontulloch, line 304: Is this always going to be an ancestor term that we want returned?  Or could it be the case that the term itself belongs to the subset and can just be returned?

src/main/java/org/alliancegenome/curation_api/model/entities/AnatomicalSite.java
```diff
@@ -53,6 +55,19 @@ public class AnatomicalSite extends AuditedObject {
 	@ManyToOne
 	@JsonView({View.FieldsOnly.class})
 	private GOTerm cellularComponentTerm;
+	
+	//celullar compoent ribbon -- slim
+	@IndexedEmbedded(includePaths = {"name", "name_keyword"})
+	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
+	@ManyToOne
+	@JsonView({View.FieldsOnly.class})
+	private GOTerm cellularComponentRibbonTerm;
```
markquintontulloch, line 64: Should these fields be added to the LinkML model, or is there a reason to exclude them?  If so, we'd need a new LinkML release and to update the @AGRCurationSchemaVersion annotation and the LinkMLSchemaConstants.java class.

src/main/java/org/alliancegenome/curation_api/services/validation/dto/fms/GeneExpressionAnnotationFmsDTOValidator.java
```diff
@@ -212,6 +212,12 @@ private ObjectResponse<AnatomicalSite> validateAnatomicalSite(GeneExpressionFmsD
 				if (cellularComponent == null) {
 					response.addErrorMessage("whereExpressed - cellularComponentTermId", ValidationConstants.INVALID_MESSAGE + " (" + geneExpressionFmsDTO.getWhereExpressed().getCellularComponentTermId() + ")");
 				} else {
+					GOTerm cellularComponentRibbon = goTermService.findSubsetTerm(cellularComponent, "goslim_agr");
+					if (cellularComponentRibbon == null) {
+						anatomicalSite.setCellularComponentOther(true);
+					} else {
+						anatomicalSite.setCellularComponentRibbonTerm(cellularComponentRibbon);
```
markquintontulloch, line 220: I think you also need to set cellularComponentOther to false here, otherwise if it's an update where the value was previously true then it would remain as true.

src/test/java/org/alliancegenome/curation_api/ExpressionBulkUploadFmsITCase.java
```diff
@@ -88,6 +91,8 @@ public void expressionBulkUploadAllFields() throws Exception {
 			.body("results[0].expressionPattern.whereExpressed.anatomicalStructure.curie", is(anatomicalStructureTermId))
 			.body("results[0].expressionPattern.whereExpressed.anatomicalSubstructure.curie", is(anatomicalSubstructureTermId))
 			.body("results[0].expressionPattern.whereExpressed.cellularComponentTerm.curie", is(cellularComponentTermId))
+			.body("results[0].expressionPattern.whereExpressed.cellularComponentRibbonTerm.curie", is(cellularComponentRibbonTermId))
+			.body("results[0].expressionPattern.whereExpressed.cellularComponentOther", is(false))
```
markquintontulloch, line 95: I tend to set booleans to the non-default value for initial loads - that way you can check it isn't just the default being applied - and then do an update test where it gets set to the default.  Maybe that doesn't make sense here since the value for this field depends on the presence of another field, but the tests here wouldn't catch the issue I mentioned above about specifically setting cellularComponentOther to false in updates.

src/test/java/org/alliancegenome/curation_api/base/BaseITCase.java
```diff
@@ -344,6 +344,25 @@ public GOTerm createGoTerm(String curie, String name, Boolean obsolete) {
 		return response.getEntity();
 	}
 
+	public GOTerm createGOTermWithSubsets(String curie, String name, List<String> subsets) throws Exception {
```
markquintontulloch: Perhaps we could refactor the existing createGoTerm method to take subsets and ancester as input parameters and then just have methods that pass in nulls as appropriate for when we don't need those.

markquintontulloch (changes requested):

src/main/java/org/alliancegenome/curation_api/services/base/BaseOntologyTermService.java
adamgibs: I think it will always be an ancestor. The [loader code](https://github.dev/alliance-genome/agr_loader/blob/2c3dcf03c12df45cc95ddc2f7abb41469836e40f/src/etl/expression_ribbon_etl.py#L43) that this is replacing  traverses  the PART_OF|/S_A relationship to get the ancestors.

src/main/java/org/alliancegenome/curation_api/model/entities/AnatomicalSite.java
adamgibs: Ah, right. I can add these to the model.

src/main/java/org/alliancegenome/curation_api/services/validation/dto/fms/GeneExpressionAnnotationFmsDTOValidator.java
adamgibs: Sure, that makes sense. I can add that.

src/test/java/org/alliancegenome/curation_api/ExpressionBulkUploadFmsITCase.java
adamgibs: Okay, maybe it would make sense to add a separate test specifically for that situation then.

src/test/java/org/alliancegenome/curation_api/base/BaseITCase.java
adamgibs: Yeah, I can definitely do that.

src/main/java/org/alliancegenome/curation_api/services/validation/dto/fms/GeneExpressionAnnotationFmsDTOValidator.java
abecerra: @adamgibs , do we have all the GoTerms in the  "goslim_agr" subset already loaded?

adamgibs: @abecerra Yeah, as far as I know. There are GOTerms with subsets that include "goslim_agr" currently.

adamgibs:
@markquintontulloch Okay, I think I've addressed most of the issues that have been brought up here. The exception is the new test for the `cellularComponentOther` `boolean`. I wasn't able to write a test that didn't either result in a `GOTerm` duplication error or creating a second expression annotation with different state, which defeats the  purpose of trying to update the same `boolean` value. The first issue arose trying to run the same annotation through but updating the `GOTerm` to have an ancestor the second time and thereby flipping the `boolean` to false. The second issue arose when running the same annotation through, but with a different `cellularComponentTermId` that could then have an ancestor added without the duplication error.

src/main/java/org/alliancegenome/curation_api/model/entities/AnatomicalSite.java
markquintontulloch: I've created a new LinkML release to include your changes (v2.6.1) so the `@AGRCurationSchemaVersion` annotation on this class needs to be updated and `LinkMLSchemaConstants.LATEST_RELEASE` needs to be bumped to 2.6.1

src/test/java/org/alliancegenome/curation_api/base/BaseITCase.java
```diff
@@ -1241,11 +1251,21 @@ public void loadGenes(List<String> modEntityIds, String taxonCurie, VocabularyTe
 	}
 
 	public void loadGOTerm(String curie, String name) throws Exception {
```
markquintontulloch: We can get rid of the loadGOTerm methods entirely since they're doing the same thing as createGoTerm, just not returning the term (this is an issue elsewhere in the tests too, but might as well sort out this one now)

markquintontulloch:
_@adamgibs I don't think we need to worry too much about the integration test for the `cellularComponentOther` update as it's a bit of an edge case and the test setup required is perhaps a bit overly convoluted.  Just need to update the LinkML versions and drop the loadGOTerm methods and I think this is good to go._

Ignore the above, I've pushed these changes since you're away

markquintontulloch (approved):

PR #1642, opened 2024-08-15
SCRUM-4336 Implement Reference caching for data loads

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/ReferenceService.java
```diff
@@ -45,6 +51,23 @@ public ObjectResponse<Reference> getByCurie(String curie) {
...
+			if (referenceCacheMap.containsKey(curieOrXref)) {
+				reference = referenceCacheMap.get(curieOrXref);
+			} else {
+				Log.debug("Reference not cached, caching reference: (" + curieOrXref + ")");
+				reference = findOrCreateReference(curieOrXref);
+				referenceCacheMap.put(curieOrXref, reference);
+			}
+		} else {
+			reference = findOrCreateReference(curieOrXref);
+			referenceRequest = new Date();
```
oblodgett, line 64: I would think you would want to put it in the map here also.

oblodgett: PS not sure what the date is doing here... typically the logic goes:
if(exists in map) return ref;
else { lookup and cache in map) return ref;

markquintontulloch: Same logic we’ve used on all other caches we’ve implemented. Doesn’t get cached until 2nd time it’s seen so that only common (well, more than 1 at least) objects get stored.  Date could just as well be a Boolean I suppose.

oblodgett (approved):

PR #1644, opened 2024-08-20
SCRUM-4174 (update) Remove Pseudogene: prefix from WB GFF entries

markquintontulloch:


oblodgett (approved):
This is going to kill the current loads?

markquintontulloch:
I'm not in a rush to merge it

PR #1645, opened 2024-08-22
Initial Code for flattening the tables

oblodgett:


oblodgett:
@markquintontulloch so not sure what is happening but when doing a `select * from slotannotation where type is null` I get like 1.4M records so I must be missing some slotannotations somewhere.

markquintontulloch:
> @markquintontulloch so not sure what is happening but when doing a `select * from slotannotation where type is null` I get like 1.4M records so I must be missing some slotannotations somewhere.

@oblodgett weird - I don't think you've missed any in the migration.  Perhaps we've had a previous issue in a migration that has resulted in entries in child tables being deleted but not in the parent slotannotation table?

oblodgett:
@markquintontulloch im going to redownload the DB and try it again.

src/main/java/org/alliancegenome/curation_api/dao/slotAnnotations/constructSlotAnnotations/ConstructComponentSlotAnnotationDAO.java
```diff
@@ -19,7 +19,7 @@ protected ConstructComponentSlotAnnotationDAO() {
 	}
 
 	public void deleteAttachedNote(Long id) {
-		Query jpqlQuery = entityManager.createNativeQuery("DELETE FROM constructcomponentslotannotation_note WHERE relatednotes_id = '" + id + "'");
+		Query jpqlQuery = entityManager.createNativeQuery("DELETE FROM slotannotation_note WHERE relatednotes_id = '" + id + "'");
```
oblodgett, line 22: @markquintontulloch the draw back to this kind of query is the hardcoding the table. I wondering if we could use a JSQL query. But for now.

markquintontulloch (approved):

src/main/java/org/alliancegenome/curation_api/dao/slotAnnotations/constructSlotAnnotations/ConstructComponentSlotAnnotationDAO.java
markquintontulloch: I feel like we should be able to do this with orphanRemoval = true and not need this at all

oblodgett: Ya ... I agree.

PR #1655, opened 2024-08-30
SCRUM-3952: fix to save exceptions in DB

abecerra:


src/main/java/org/alliancegenome/curation_api/response/APIResponse.java
```diff
@@ -50,7 +48,7 @@ public void addErrorMessages(String fieldName, Integer rowIndex, Object fieldErr
 	}
 
 	public boolean hasErrors() {
-		return StringUtils.isNotEmpty(errorMessage) || MapUtils.isNotEmpty(errorMessages);
+		return StringUtils.isNotEmpty(errorMessage) || MapUtils.isNotEmpty(errorMessages) || MapUtils.isNotEmpty(supplementalData);
```
oblodgett: @abecerra I don't think this is the answer to fix the issue. As you can see here the errorMessage or the errorMessages are empty. Those needs values in order for the exceptions to show correctly. The exceptions are getting saving to the database its just that errorMessage is empty.

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/response/APIResponse.java
abecerra: Without the ` MapUtils.isNotEmpty(supplementalData)` the integration tests break. 
With it the tests succeed and a local load of XBXL data now stores the exceptions correctly.

oblodgett: Sure... but why is that. Something is not setting the errorMessage correctly otherwise the tests wont fail.

oblodgett: The old code works for all the other integration tests?

oblodgett: Also supplementalData doesn't mean there is errors we use that to deliver over data to the front end.

abecerra: I think my problem is using code like:
 	`response.addErrorMessages("whenExpressed", temporalContextObjectResponse.getErrorMessages()); `
I will try with:
	`response.addErrorMessage("whenExpressed", temporalContextObjectResponse.errorMessagesString());`

oblodgett (approved):

PR #1665, opened 2024-09-10
Fix missing Indexes and Flatten Associations

oblodgett:


src/main/resources/db/migration/v0.37.0.49__flatten_assocations.sql
```diff
@@ -576,4 +581,71 @@ ALTER TABLE ONLY association ADD CONSTRAINT fkeot29m2x4peefkgwrqmu36ub1 FOREIGN
...
+CREATE INDEX biologicalentity_curie_index ON public.biologicalentity USING btree (curie);
+CREATE INDEX biologicalentity_modinternalid_index ON public.biologicalentity USING btree (modinternalid);
+CREATE INDEX cellularcomponentqualifiers_cellularcomponentqualifiers_index ON public.anatomicalsite_cellularcomponentqualifiers USING btree (cellularcomponentqualifiers_id);
+CREATE INDEX informationcontententity_curie_index ON public.informationcontententity USING btree (curie);
+CREATE INDEX ontologyterm_curie_index ON public.ontologyterm USING btree (curie);
+CREATE INDEX ontologyterm_isa_ancestor_descendant_isancestors_index ON public.ontologyterm_isa_ancestor_descendant USING btree (isaancestors_id);
+
+-- Missing Constrains
+
+ALTER TABLE ONLY public.anatomicalsite ADD CONSTRAINT fk4h252bd81x75fwxgyy6wa61u7 FOREIGN KEY (anatomicalstructure_id) REFERENCES public.ontologyterm(id);
```
markquintontulloch, line 607: Can we give these constraints non-cryptic names?

oblodgett: The thing is that hibernate makes that hash as a combination of the table and field name. Yes we could name them whatever we want, but it would be good to have that name in java. I am not sure where the FK definition goes in java, I would guess where the index is defined.

markquintontulloch (approved):

PR #1693, opened 2024-10-24
SCRUM-4440 add biogrid orc load

adamgibs:


oblodgett (approved):
This is just the first phase where all the load infrastructure is in place, and the next PR will be the actual update to the DB?

adamgibs:
@oblodgett Yep, this is just the initial exec code with history, skips and migration.

markquintontulloch (approved):

PR #1709, opened 2024-11-07
SCRUM-4190 make xenbase and human gaf file load

cmpich:


src/main/resources/db/migration/v0.38.0.2__gaf-load.sql
```diff
@@ -168,7 +168,7 @@ from bulkload
 where name = 'RGD GAF';
 
 insert into bulkurlload (id, bulkloadurl)
-select id, 'https://www.ebi.ac.uk/gxa/species/Homo_sapiens/sitemap.xml?allEntries=true'
+select id, 'https://download.alliancegenome.org/2.3.0/GAF/HUMAN/GAF_HUMAN_1.gaf'
```
oblodgett: You can't change a migration once its been applied. You will need to create a new migration. Unless you plan to change this manually (db update) on alpha.

oblodgett: Plus this is looking more and more like a FMS load vs a URL load. Given that they latest version of this file is: https://download.alliancegenome.org/7.4.0/GAF/HUMAN/GAF_HUMAN_0.gaf.gz this needs to be turned into a FMS load so that we will always get the latest version of the file.

PR #1714, opened 2024-11-08
SCRUM-4190

cmpich:


src/main/java/org/alliancegenome/curation_api/services/GeneOntologyAnnotationService.java
```diff
@@ -106,14 +109,24 @@ private Long getGOID(GeneOntologyAnnotationDTO uiEntity) {
...
 
-	private Species getSpecies(String orgAbbreviation) {
-		if (species != null) {
+	private List<Species> getSpecies(String orgAbbreviation) {
+		if (CollectionUtils.isNotEmpty(species)) {
 			return species;
 		}
+		if (orgAbbreviation.equalsIgnoreCase("XB")) {
+			species = new ArrayList<>();
+			species.add(getSingleSpecies("XBXL"));
```
markquintontulloch, line 118: The whole getSingleSpecies method could be replaced here with `speciesDAO.findByField("displayName", "XB").getSingleResult();`

src/main/resources/db/migration/v0.38.0.3__gaf-load.sql
```diff
@@ -0,0 +1,3 @@
+update bulkurlload set bulkloadurl = 'https://download.alliancegenome.org/2.3.0/GAF/HUMAN/GAF_HUMAN_1.gaf'
```
markquintontulloch: Need to bump the version of this file

oblodgett: This needs to be turned into a FMS load which will always pull the latest version of the file.

oblodgett (changes requested):

src/main/resources/db/migration/v0.38.0.3__gaf-load.sql
cmpich: Do we know from where this file gets downloaded from for the loader? Only the human file is referenced from the FMS in the YAML file while all other MODs files have a source URL given at GOC:
https://github.com/alliance-genome/agr_ferret/blob/master/src/datasets/GAF.yaml

cmpich: Ok, the human file comes from RGD and that MOD is submitting the generated file into the FMS.

abecerra (approved):

src/main/resources/db/migration/v0.38.0.11__gaf-load.sql
```diff
@@ -0,0 +1,3 @@
+update bulkurlload set bulkloadurl = 'https://download.alliancegenome.org/2.3.0/GAF/HUMAN/GAF_HUMAN_1.gaf'
```
markquintontulloch: Doesn't this migration need to be deleted now?

cmpich: Yes, that is being delete in the next script.

cmpich: Actually, I did remove this as it was run second.

markquintontulloch: I mean that this whole file can be deleted - it's not doing anything as it's updating an entry that you've already removed

markquintontulloch: Sorry, I see you've already removed it!

markquintontulloch (approved):

oblodgett (approved):

PR #1726, opened 2024-11-13
SCRUM-4513

adamgibs:


src/main/java/org/alliancegenome/curation_api/services/DataProviderService.java
```diff
@@ -92,6 +92,21 @@ public ObjectResponse<DataProvider> insertExpressionAtlasDataProvider(DataProvid
...
+	@Transactional
+	public ObjectResponse<DataProvider> insertBioGridOrcDataProvider(DataProvider entity, Long geneticEntityId) {
+		String referencedCurie = entity.getCrossReference().getReferencedCurie();
+		
+		DataProvider dbEntity = getDataProvider(entity.getSourceOrganization(), referencedCurie, entity.getCrossReference().getResourceDescriptorPage());
+		
+		// we only create new records, no updates
+		if (dbEntity == null) {
+			dataProviderDAO.persist(entity);
+			crossReferenceDAO.persistAccessionGeneAssociated(entity.getCrossReference().getId(), geneticEntityId);
```
markquintontulloch: I'm not quite following what's going on here.  This seems to be taking the dataProvider xref and linking it to a genomic entity but a) we don't reuse cross-references; and b) I'm not sure why this persistAccessionGeneAssociated() method uses native SQL queries as opposed to using Hibernate to add a cross-reference to the genomic entity.

adamgibs: @markquintontulloch Yeah, so perhaps the `dataProvider` part isn't needed? Originally, I used `persistAccessionGeneAssociated` directly to link a newly created `crossReference` with an already-existing `genomicEntity`. I started using `dataProviders` because I thought that maybe the newly created `crossReferences` needed to be linked to an organization, but maybe that isn't necessary. I'm also using the newly created `dataProvider` ids to track `idsBefore` and `idsAdded`. Maybe I should use `crossRef` ids for that instead? 
I'm using the `persistAccessionGeneAssociated` because it already existed for the expression atlas stuff and seemed to do exactly what I needed. I also created a new DAO method, `getGenesWithCrossRefs`, that uses raw SQL because that seemed like the most straight-forward way to replicate the loader code, which uses this neo query 
```
MATCH (g:Gene)-[crr:CROSS_REFERENCE]-(cr:CrossReference)
                   WHERE cr.globalCrossRefId IN $parameter
                   RETURN g.primaryKey, g.modLocalId, cr.name, cr.globalCrossRefId
```

oblodgett (approved):

markquintontulloch (approved):

PR #1735, opened 2024-11-21
Scrum 4436

cmpich:


src/main/java/org/alliancegenome/curation_api/services/associations/constructAssociations/ConstructGenomicEntityAssociationService.java
```diff
@@ -28,23 +24,28 @@
 import org.alliancegenome.curation_api.services.validation.associations.constructAssociations.ConstructGenomicEntityAssociationValidator;
 import org.alliancegenome.curation_api.services.validation.dto.associations.constructAssociations.ConstructGenomicEntityAssociationDTOValidator;
 
-import io.quarkus.logging.Log;
-import jakarta.annotation.PostConstruct;
-import jakarta.enterprise.context.RequestScoped;
-import jakarta.inject.Inject;
-import jakarta.transaction.Transactional;
+import java.time.OffsetDateTime;
```
devopsagr, line 27: 🚫 **[checkstyle]** <com.puppycrawl.tools.checkstyle.checks.imports.UnusedImportsCheck> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Unused import - java.time.OffsetDateTime.

src/main/java/org/alliancegenome/curation_api/model/entities/Allele.java
```diff
@@ -212,4 +205,15 @@ public class Allele extends GenomicEntity {
 		@Index(name = "allele_note_relatednotes_index", columnList = "relatedNotes_id")})
 	private List<Note> relatedNotes;
 
+
+	@IndexedEmbedded(includePaths = {
```
markquintontulloch, line 209: What's the reason for moving this from GenomicEntity to the child classes?  If this is necessary then shouldn't it also be on the various other GenomicEntity classes (Transcript, Exon, etc.).  Or is this to avoid joining all those tables, in which case do we really need a model change similar to the one for disease genetic modifiers?

src/main/java/org/alliancegenome/curation_api/model/entities/associations/constructAssociations/ConstructGenomicEntityAssociation.java
```diff
@@ -17,26 +17,13 @@
...
-import lombok.EqualsAndHashCode;
-import lombok.ToString;
+import java.util.List;
 
 @Entity
 @Data
 @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
-@ToString(callSuper = true)
-@AGRCurationSchemaVersion(min = "2.2.0", max = LinkMLSchemaConstants.LATEST_RELEASE, dependencies = { EvidenceAssociation.class })
+@ToString(callSuper = false)
```
markquintontulloch: Should this really be false?

src/main/java/org/alliancegenome/curation_api/services/InformationContentEntityService.java
```diff
@@ -28,4 +28,8 @@ public InformationContentEntity retrieveFromDbOrLiteratureService(String curieOr
 
 		return ice;
 	}
+
+	public InformationContentEntity retrieveFromDbOrLiteratureServicea(String pmid) {
```
markquintontulloch: Can this be renamed to something more descriptive to indicate why it is different from `retrieveFromDbOrLiteratureService()`

src/main/java/org/alliancegenome/curation_api/services/ReferenceService.java
```diff
@@ -56,16 +59,40 @@ public Reference retrieveFromDbOrLiteratureService(String curieOrXref) {
...
+		if (shallowReferenceCacheMap.containsKey(pmid)) {
+			reference = shallowReferenceCacheMap.get(pmid);
+		} else {
+			Log.debug("Reference not cached, caching reference: (" + pmid + ")");
+			if (shallowReferenceCacheMap.isEmpty()) {
+				shallowReferenceCacheMap = referenceDAO.getShallowReferenceMap();
+				reference = shallowReferenceCacheMap.get(pmid);
+			} else {
+/*
+					reference = findOrCreateReference(curieOrXref);
```
markquintontulloch: So what happens if the reference isn't in our database?

src/main/java/org/alliancegenome/curation_api/model/entities/Allele.java
cmpich: Yes, we did this to avoid joining all other genomic entities.

src/main/java/org/alliancegenome/curation_api/services/ReferenceService.java
cmpich: adding the logic to create the reference (from the literature system)

src/main/java/org/alliancegenome/curation_api/services/InformationContentEntityService.java
cmpich: Wasn't used any longer. Removed it.

src/main/java/org/alliancegenome/curation_api/model/entities/Allele.java
markquintontulloch: This needs a LinkML model change then

src/main/java/org/alliancegenome/curation_api/services/validation/dto/associations/EvidenceAssociationDTOValidator.java
```diff
@@ -12,17 +11,17 @@
...
 @RequestScoped
 public class EvidenceAssociationDTOValidator extends BaseDTOValidator {
 
-	@Inject InformationContentEntityService informationContentEntityService;
+	@Inject
+	InformationContentEntityService informationContentEntityService;
 
 	public <E extends EvidenceAssociation, D extends EvidenceAssociationDTO> ObjectResponse<E> validateEvidenceAssociationDTO(E association, D dto) {
 		ObjectResponse<E> assocResponse = validateAuditedObjectDTO(association, dto);
-		association = assocResponse.getEntity();
```
markquintontulloch, line 25: why has this been removed?

src/main/java/org/alliancegenome/curation_api/services/associations/constructAssociations/ConstructGenomicEntityAssociationService.java
```diff
@@ -59,9 +59,11 @@ public ObjectResponse<ConstructGenomicEntityAssociation> upsert(ConstructGenomic
 			return null;
 		}
 		dbEntity = constructGenomicEntityAssociationDAO.persist(dbEntity);
+/*
```
markquintontulloch, line 62: Don't we need this code?

src/main/java/org/alliancegenome/curation_api/services/validation/dto/associations/EvidenceAssociationDTOValidator.java
markquintontulloch: Replaced by my PR

cmpich: It's not needed as the changes are done on that object as it is passed into the method.

src/main/java/org/alliancegenome/curation_api/services/associations/constructAssociations/ConstructGenomicEntityAssociationService.java
cmpich: It does not needed to be done as this is only adding the association to the construct object but the association already knows about it. It would only be needed if you want to see it on the construct object which we don;t in this load.

markquintontulloch: Are you sure about that? Don’t we list these in the Constructs table (or if not, won’t we be doing so in the future)?

markquintontulloch (approved):

PR #1747, opened 2024-12-03
SCRUM-4618: add MaTI host info and link

abecerra:


oblodgett (changes requested):
what would be better is to use the environment variable that is already configured for the mati host

oblodgett (commented):
@ConfigProperty(name = "mati.url")

abecerra:
> @ConfigProperty(name = "mati.url")

Done

oblodgett (approved):

PR #1754, opened 2024-12-06
SCRUM-4634: production release/v0.36.0

abecerra:


VarunReddy1111 (approved):

oblodgett (changes requested):
i thought we were doing alpha to beta today?

abecerra:
> i thought we were doing alpha to beta today?

Beta is deployed now!. 
This is to deploy on Monday if Chris signs up beta, but it could wait if needed

oblodgett:
ahh ok

oblodgett (approved):

PR #1758, opened 2024-12-10
Add related notes to Genes table

VarunReddy1111:


markquintontulloch (changes requested):
Backend stuff looks good, just needs integration tests updating/adding

markquintontulloch:
@VarunReddy1111 For completeness you should probably add integration tests for empty/missing required fields and invalid field values for the related notes.  Admittedly the actual validation code is tested elsewhere, but this may be subject to change and we may, for example, decide to implement a gene-specific set of VocabularyTerms for note type.

markquintontulloch (approved):

PR #1761, opened 2024-12-12
SCRUM-4691 Fix cleanup of interaction loads

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/jobs/executors/LoadFileExecutor.java
```diff
@@ -277,12 +277,13 @@ protected <E extends AuditedObject, T extends BaseDTO> boolean runLoad(BaseUpser
...
 		ProcessDisplayHelper ph = new ProcessDisplayHelper(10000);
 		ph.startProcess("Deletion/deprecation of: " + dataProviderName + " " + loadTypeString, idsToRemove.size());
-		//updateHistory(history);
+		updateHistory(history);
 		for (Long id : idsToRemove) {
 			try {
 				service.deprecateOrDelete(id, false, loadDescription, deprecate);
 				history.incrementCompleted(countType);
 			} catch (Exception e) {
+				e.printStackTrace();
```
oblodgett, line 296: Do we want this turned on?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/jobs/executors/LoadFileExecutor.java
markquintontulloch: I think so, otherwise there's no indication of failure in the logs

PR #1775, opened 2024-12-30
Add convenience methods

cmpich:
I need this method for the java repo to add PubModIDs to the DA doc.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
+++ b/src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
@@ -1,10 +1,11 @@
 package org.alliancegenome.curation_api.model.entities;
 
-import java.util.Collections;
-import java.util.List;
-import java.util.Optional;
-import java.util.stream.Collectors;
-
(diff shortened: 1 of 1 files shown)


src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
```diff
@@ -72,20 +60,46 @@ public class Reference extends InformationContentEntity {
...
+	 * Retrieve PUB MOD ID
+	 */
+	@Transient
+	@JsonIgnore
+	public String getPubModID() {
+		return getReferenceID(false);
+	}
+
+	@Transient
+	private String getReferenceID(boolean pubmedIdFirst) {
```
oblodgett, line 76: Seems like this method does two different functions based on boolean input. I would guess this should be two methods?

cmpich: Yes, that's why it's marked private. The two public methods calling this method are defined as 
getReferenceID()
and
getPubModID()

oblodgett (approved):

PR #1779, opened 2025-01-03
SCRUM-4710 disable scheduling of mass indexing

cmpich:


src/main/java/org/alliancegenome/curation_api/jobs/JobScheduler.java
```diff
@@ -238,7 +238,6 @@ public void cleanUpFileHistories() {
 		}
 	}
 
-	@Scheduled(cron = "0 0 0 ? * SUN")
```
oblodgett, line 241: If not removing the method can you just comment this vs removing it.

cmpich: Sure.

oblodgett (approved):

PR #1788, opened 2025-01-09
SCRUM-4657: add references to Variant FMS load

abecerra:


src/main/java/org/alliancegenome/curation_api/services/validation/dto/fms/VariantFmsDTOValidator.java
```diff
@@ -176,7 +167,29 @@ public Long validateVariant(VariantFmsDTO dto, List<Long> idsAdded, BackendBulkD
...
+			}
+			if (CollectionUtils.isNotEmpty(validatedReferences)) {
+				if (variant.getReferences() == null) {
+					variant.setReferences(new ArrayList<>());
+				} else {
+					variant.getReferences().clear();
+				}
+				variant.getReferences().addAll(validatedReferences);
+			}
+		}
```
markquintontulloch, line 193: What happens if `dto.getReferences()` is empty?  You're not clearing any existing references from the database.  There is code in the AlleleDTOValidator that can basically be copied and pasted here.

src/test/java/org/alliancegenome/curation_api/VariantFmsITCase.java
```diff
@@ -111,10 +111,10 @@ public void variantFmsBulkUpload() throws Exception {
...
-			body("entity.crossReferences[0].resourceDescriptorPage.name", is("homepage"));
-
+			body("entity.crossReferences[0].resourceDescriptorPage.name", is("homepage")).
+			body("entity.references[0].crossReferences[0].referencedCurie", is(pubmedId));
 	}
-	
+
 	@Test
 	@Order(2)
 	public void variantFmsBulkUploadUpdate() throws Exception {
```
markquintontulloch, line 120: Need to check that update of references works as expected in this test

src/test/java/org/alliancegenome/curation_api/VariantFmsITCase.java
```diff
@@ -196,7 +196,7 @@ public void variantFmsBulkUploadEmptyRequiredFields() throws Exception {
 		params.put("Locations", createCountParams(1, 0, 1, 0));
 		checkBulkLoadRecordCounts(variantFmsBulkPostEndpoint, variantFmsTestFilePath + "ER_05_empty_allele_id.json", params);
 	}
-	
+
 	@Test
 	@Order(5)
 	public void variantFmsBulkUploadInvalidFields() throws Exception {
```
markquintontulloch, line 202: Should have a test for invalid reference here

markquintontulloch (changes requested):
Should really also have integration tests for updates with missing/empty references list.  These would have caught the bug in the validation code. 

As adding the column to the UI, also need to add a UI test to the VariantsTable.test

src/main/java/org/alliancegenome/curation_api/services/validation/dto/fms/VariantFmsDTOValidator.java
abecerra: Changed as AlleleDTOValidator

markquintontulloch:
> Should really also have integration tests for updates with missing/empty references list. These would have caught the bug in the validation code.
> 
> As adding the column to the UI, also need to add a UI test to the VariantsTable.test

@abecerra These comments still need to be addressed

markquintontulloch (approved):

PR #1809, opened 2025-01-18
SCRUM-4764 clean out error response for each entity in AGM load

cmpich:


src/main/java/org/alliancegenome/curation_api/services/validation/dto/AffectedGenomicModelDTOValidator.java
```diff
@@ -42,10 +42,10 @@ public class AffectedGenomicModelDTOValidator extends BaseDTOValidator {
 	@Inject
 	AgmSecondaryIdSlotAnnotationDTOValidator agmSecondaryIdDtoValidator;
 
-	private final ObjectResponse<AffectedGenomicModel> agmResponse = new ObjectResponse<>();
+	private ObjectResponse<AffectedGenomicModel> agmResponse = new ObjectResponse<>();
 
 	public AffectedGenomicModel validateAffectedGenomicModelDTO(AffectedGenomicModelDTO dto, BackendBulkDataProvider dataProvider) throws ValidationException {
-
+		agmResponse = new ObjectResponse<>();
```
oblodgett, line 48: @markquintontulloch I'm not sure this is the way to go. There are other assumption being made on this object. Why does this need to be done?

markquintontulloch: I have no issue with this change - it's how it's done in all the other DTO validators and I think was this way previously for AGMs too.

markquintontulloch (approved):

src/main/java/org/alliancegenome/curation_api/services/validation/dto/AffectedGenomicModelDTOValidator.java
cmpich: Without this re-assignment this object keeps the errors from the previous entity.

PR #1823, opened 2025-02-05
Add cleanUp parameter documentation

chris-grove:
Added a section to the Submitting Data portion of the agr_curation README to explain how to turn off data load clean up using the ?cleanUp=false parameter passed to the API. I'd appreciate the developers reviewing my language and wording to check if I've made any mis-statements. Thanks!


markquintontulloch (commented):
Not sure what you mean by “outstanding duplicate entities”, but otherwise looks fine.  I don’t recall if we currently have any data types where we delete entries rather than deprecate, but that distinction is probably unimportant here anyway.

chris-grove:
@markquintontulloch 
Sounds good, I removed that language

markquintontulloch (approved):

PR #1825, opened 2025-02-07
Removed duplicate setting of reverse connect to objects

oblodgett:


src/main/java/org/alliancegenome/curation_api/services/associations/alleleAssociations/AlleleGeneAssociationService.java
```diff
@@ -60,8 +60,6 @@ public ObjectResponse<AlleleGeneAssociation> upsert(AlleleGeneAssociation uiEnti
 			return null;
 		}
 		dbEntity = alleleGeneAssociationDAO.persist(dbEntity);
-		addAssociationToAllele(dbEntity);
```
markquintontulloch, line 63: Failing integration tests suggest that this may be needed (or something needs to replace it).  As far as I know, we haven't yet implemented CRUD operations for the other associations types changed here.

markquintontulloch: Also, I'm not convinced our integration tests are sufficient to catch problems with updates in the bulk loads.  For example. the update test for AgmAlleleAssociations only checks that internal/obsolete are updated but doesn't try to update values for zygosity / created_by / updated_by / date_created / date_updated.

oblodgett: I guess in this case I would have thought that `validateAlleleGeneAssociationDTO` would do this work given that it runs a:
```java
association = validateAlleleGenomicEntityAssociationDTO(association, dto, VocabularyConstants.ALLELE_GENE_RELATION_VOCABULARY_TERM_SET);
```
I'll look into this more I guess.

markquintontulloch: Seems like the bulk load part is okay (based on integration tests), it's the non-bulk load CRUD stuff that seems to be failing

oblodgett: I guess I might not understand here but here is the test:

```java
	private String reference = "AGRKB:000000001";
	private String reference2 = "AGRKB:000000021";
	private String evidenceCodeCurie = "DATEST:Evidence0001";
	private String evidenceCodeCurie2 = "DATEST:Evidence0002";
```

```java
	body("entity.evidence[0].curie", is(reference)).
	body("entity.evidenceCode.curie", is(evidenceCodeCurie)).
```

However this is what is in the file:
```java
    "evidence_code_curie": "DATEST:Evidence0001",
    "evidence_curies": [
      "PMID:25920550"
    ],
```
The next test:
```java
	body("entity.evidence[0].curie", is(reference2)).
	body("entity.evidenceCode.curie", is(evidenceCodeCurie2)).
```
However the file has:
```java 
    "evidence_code_curie": "DATEST:Evidence0002",
    "evidence_curies": [
      "PMID:25920551"
    ],
```

But fails with:
```
JSON path entity.evidence[0].curie doesn't match.
Expected: is "AGRKB:000000021"
  Actual: AGRKB:000000001
```

I'm guessing the AGRKB:000000001 gets injected when the reference can't be found?

markquintontulloch: No, they're references created in earlier integration tests.  AGRKB:000000001 = PMID:25920550 and AGRKB:000000021 = PMID:25920551.  In bulk loads we're expecting the xref to be submitted rather than the AGR curie.  That test failing suggests that the update isn't working.

oblodgett: Never mind the association was not getting updated.... because of the early added return statement.

markquintontulloch (approved):

PR #1830, opened 2025-02-11
add AgmAgmAssociation load to AGM load

cmpich:


src/main/java/org/alliancegenome/curation_api/jobs/executors/BulkLoadJobExecutor.java
```diff
@@ -75,6 +75,7 @@ public void process(BulkLoadFileHistory bulkLoadFileHistory, Boolean cleanUp) th
 
 			if (loadType == AGM || loadType == FULL_INGEST) {
 				agmExecutor.execLoad(bulkLoadFileHistory, cleanUp);
+				agmAgmAssociationExecutor.execLoad(bulkLoadFileHistory, cleanUp);
```
oblodgett, line 78: @cmpich shouldn't this be under the AGM_ASSOCIATION section?

markquintontulloch: It should, and it is!  This PR should be closed.

PR #1833, opened 2025-02-12
SCRUM-4634: remove entry from migration

abecerra:
Trying to solve deployment  failure:

` Caused by: org.flywaydb.core.internal.command.DbMigrate$FlywayMigrateException: Script v0.38.0.22__agm_association_loads.sql failed
Feb 12 10:04:48 agr.curation.production.api.server : 	at org.alliancegenome.curation_api.main.Main.main(Main.java:11)
Feb 12 10:04:48 agr.curation.production.api.server : 	at io.quarkus.runtime.Quarkus.run(Quarkus.java:124)
Feb 12 10:04:48 agr.curation.production.api.server : ---------------------------------------------------
Feb 12 10:04:48 agr.curation.production.api.server : Error Code : 0
Feb 12 10:04:48 agr.curation.production.api.server : SQL State  : 23505
Feb 12 10:04:48 agr.curation.production.api.server : Location   : db/migration/v0.38.0.22__agm_association_loads.sql (/agr_curation/file:/agr_curation/agr_curation_api-runner.jar!/db/migration/v0.38.0.22__agm_association_loads.sql)
Feb 12 10:04:48 agr.curation.production.api.server :   Detail: Key (vocabularylabel)=(agm_relation) already exists.
Feb 12 10:04:48 agr.curation.production.api.server : Message    : ERROR: duplicate key value violates unique constraint "vocabulary_vocabularylabel_uk"
Feb 12 10:04:48 agr.curation.production.api.server : Statement  : Run Flyway with -X option to see the actual statement causing the problem`

### DIFF
--- a/src/main/resources/db/migration/v0.38.0.22__agm_association_loads.sql
+++ b/src/main/resources/db/migration/v0.38.0.22__agm_association_loads.sql
@@ -38,7 +38,6 @@ ALTER TABLE ONLY public.agmsequencetargetingreagentassociation ADD CONSTRAINT ag
 
 --create vocabulary
 
-INSERT INTO vocabulary (id, name, vocabularylabel) VALUES (nextval('vocabulary_seq'), 'AGM Relation', 'agm_relation');
 
 INSERT INTO vocabularyterm (id, name, vocabulary_id) SELECT nextval('vocabularyterm_seq'), 'contains', id FROM vocabulary WHERE vocabulary.vocabularylabel = 'agm_relation';
 


markquintontulloch (changes requested):
I have a slight concern that removing this from the migration may cause issues with the integration tests.  Would be safer to do an upsert, e.g. [Postgres upsert query](https://neon.tech/postgresql/postgresql-tutorial/postgresql-upsert)

markquintontulloch:
Okay, so failing integration tests prove my concern correct.  The integration tests start from an empty database and need to populate the required entities via either the test methods or migrations.  In this case the tests are expecting the vocabulary to be present from the migrations.  As such, the upsert query is definitely required.

markquintontulloch:
I think you're also going to run into issues with migration v0.38.0.28 where it tries to insert the vocabulary term `has_parental_population` so this will need to be changed to an upsert too

abecerra:
> Okay, so failing integration tests prove my concern correct. The integration tests start from an empty database and need to populate the required entities via either the test methods or migrations. In this case the tests are expecting the vocabulary to be present from the migrations. As such, the upsert query is definitely required.

Ok

PR #1834, opened 2025-02-12
SCRUM-4634: Promote curation R12 S3 beta to prod (as v0.36.1)

abecerra:


markquintontulloch (changes requested):
I think deployment will fail with migration v0.38.0.28 where it tries to insert the vocabulary term `has_parental_population` so this will need to be changed to an upsert too

abecerra:
Reopening, the migration v0.38.0.28 is not in branch beta yet

markquintontulloch:
> Reopening, the migration v0.38.0.28 is not in branch beta yet

Sorry, didn't spot that

markquintontulloch (approved):

PR #1845, opened 2025-02-19
Release/v0.36.2

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
```diff
@@ -58,7 +68,7 @@ public class Reference extends InformationContentEntity {
 	 * Retrieve PMID if available in the crossReference collection otherwise MOD ID
 	 */
 	@Transient
-	@JsonIgnore
+	@JsonView(View.ForPublic.class)
```
oblodgett, line 71: Does this need the view?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/model/entities/Reference.java
markquintontulloch: Makes life a lot easier - we have it for pubModID, so don't know why we don't have this too

PR #1855, opened 2025-02-26
Added code to use the API key at least once per week in order for it to not expire.

oblodgett:


src/main/java/org/alliancegenome/curation_api/auth/AuthenticationFilter.java
```diff
@@ -227,7 +234,7 @@ private AllianceMember getAllianceMember(GroupList groupList) {
 		for (Group group : groupList) {
 			String allianceMember = (String) group.getProfile().get("affiliated_alliance_member");
 			if (allianceMember != null) {
-				SearchResponse<AllianceMember> res = allianceMemberDAO.findByField("uniqueId", allianceMember);
+				SearchResponse<AllianceMember> res = allianceMemberDAO.findByField("abbreviation", allianceMember);
```
oblodgett, line 237: @markquintontulloch in the 70bf7f579c6974934942b82e8651a676eb34e683 commit we switch from UniqueId to Audited Object, and so would need to have changed this field also.

markquintontulloch (approved):

PR #1859, opened 2025-03-02
Proposal for dynamic views on the forpublic interface call

oblodgett:


oblodgett (commented):
This will need to get "bug fixed" / cherrypicked into production.

cmpich (approved):

PR #1861, opened 2025-03-03
SCRUM-4641 Adding filtering and display of referenced curie for Crossreferences

VarunReddy1111:


src/main/cliapp/src/containers/allelesPage/AllelesTable.js
```diff
@@ -1235,12 +1235,12 @@ export const AllelesTable = () => {
 			field: 'crossReferences.displayName',
 			header: 'Cross References',
 			sortable: true,
-			filterConfig: FILTER_CONFIGS.crossReferencesFilterConfig,
+			filterConfig: FILTER_CONFIGS.alleleCrossReferencesFieldSet,
 			body: (rowData) => (
 				<ObjectListTemplate
 					list={rowData.crossReferences}
 					sortMethod={crossReferencesSort}
-					stringTemplate={(item) => `${item.displayName} (${item.resourceDescriptorPage.name})`}
```
markquintontulloch, line 1243: Personally, for the cases where _displayName_ and _referencedCurie_ differ, I'd prefer something like '_displayName_ references _referencedCurie_ (_pageName_)' rather than '_displayName_ (_referencedCurie_) (_pageName_)', to make it obvious to the user which is displayed and which is referenced.  That shouldn't hold this up though, especially if Chris has specified the display requirements.

src/main/cliapp/src/constants/FilterFields.js
```diff
@@ -201,6 +201,10 @@ export const FIELD_SETS = Object.freeze({
 		filterName: 'crossReferencesFilter',
 		fields: ['crossReferences.displayName', 'crossReferences.resourceDescriptorPage.name'],
 	},
+	alleleCrossReferencesFieldSet: {
```
markquintontulloch: I suspect this should be a global change and not just for the Allele page - i.e. we should modify both the `crossReferenceFilterConfig` and `crossReferencesFilterConfig` and update the display in all tables where xrefs are displayed (may be worth creating xref-specific components rather than just using the `ObjectListTemplate`)

src/main/cliapp/src/containers/allelesPage/AllelesTable.js
VarunReddy1111: Sure Mark, will change this and also check with Chris

src/main/cliapp/src/constants/FilterFields.js
VarunReddy1111: Sure Mark, will do that

markquintontulloch (approved):

PR #1891, opened 2025-03-11
SCRUM-4945 Enable sorting on phylogenetic order

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/bridges/IntegerValueBridge.java
```diff
@@ -0,0 +1,17 @@
...
+	public String toIndexedValue(Integer value, ValueBridgeToIndexedValueContext context) {
+		String padding = "0000000000000";
+		if (value == null) {
+			return padding;
+		}
+		String valueString = value.toString();
+		
+		return padding.substring(valueString.length()) + valueString;
+	}
+}
```
oblodgett, line 11: @markquintontulloch here is another option.

`return String.format("%013d", value.toString());`

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/model/bridges/IntegerValueBridge.java
markquintontulloch: yes, that's neater - will change

PR #1892, opened 2025-03-11
SCRUM-4945 Fix format command

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/bridges/IntegerValueBridge.java
```diff
@@ -6,6 +6,6 @@
 public class IntegerValueBridge implements ValueBridge<Integer, String> {
 	@Override
 	public String toIndexedValue(Integer value, ValueBridgeToIndexedValueContext context) {
-		return String.format("%013d", value.toString());
+		return String.format("%113s", value.toString()).replace(' ', '0');
 	}
```
oblodgett, line 10: @markquintontulloch is this a "SmartAlpha" attempt because I think padding with 1's is going to have its own issues when other numbers are involved?

oblodgett (approved):

PR #1893, opened 2025-03-12
SCRUM-4944 Fix bug preventing saving of entries with references selected from autocomplete

markquintontulloch:


markquintontulloch:
@oblodgett - Is there a better way to do this?  This fixes the bug but seems hacky.  I don't think there's any unintended consequences, but could be missing something.

oblodgett:
Ya i'm trying to understand this also... lets pick it up tomorrow.

PR #1925, opened 2025-03-24
Release/v0.37.9

adamgibs:


markquintontulloch (approved):

oblodgett:
href might be what the UI is using for the main link, but we might want to revisit this name

adamgibs:
@oblodgett Yep, that's the name the public site docs have right now. I agree that it's not very descriptive.

PR #1928, opened 2025-03-25
Changes to DiseaseSummary Document

VarunReddy1111:


src/main/java/org/alliancegenome/curation_api/model/document/builders/DiseaseSummaryDocumentBuilder.java
```diff
@@ -1,6 +1,9 @@
 package org.alliancegenome.curation_api.model.document.builders;
 
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.List;
```
devopsagr, line 5: 🚫 **[checkstyle]** <com.puppycrawl.tools.checkstyle.checks.imports.UnusedImportsCheck> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Unused import - java.util.List.

src/main/java/org/alliancegenome/curation_api/model/document/es/DiseaseSummaryDocument.java
```diff
@@ -22,6 +24,6 @@ public class DiseaseSummaryDocument extends ESDocument {
 	private DOTerm doTerm;
 	private Set<OntologyTerm> parents;
 	private Set<OntologyTerm> children;
-	private HashMap<String, String> crossReferenceLinkUrls;
+	private List<Map<String, String>> crossReferenceLinkUrls;
 	private HashMap<String, String> sourceReferenceLinkUrls;
```
oblodgett: @VarunReddy1111 you will want to change sourceReferenceLinkUrls from HashMap to Map, I would also suggest converting it over to the same structure.

VarunReddy1111: As we hardcoded the sources, which probably should not give us any error that we got for crossReferencesLinks. In this case, do we really need to change this?

oblodgett: Correct however they add a bunch of fields to the mapping, so it would be better to not have them that way.

oblodgett (approved):

PR #1930, opened 2025-03-25
Release v0.37.11

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/model/entities/ontology/NCBITaxonTerm.java
```diff
@@ -23,7 +23,7 @@
 @Entity
 @Data
 @EqualsAndHashCode(onlyExplicitlyIncluded = true, callSuper = true)
-@ToString(callSuper = true)
+@ToString(callSuper = true, exclude = {"species"})
```
oblodgett, line 26: Ahh... sorry didn't catch this, makes total sense.

oblodgett: It would have helped if we were getting better messages in the API.

oblodgett (approved):

PR #1949, opened 2025-03-28
SCRUM-4932: consolidate annotations during loads from FMS 

abecerra:
Not merge yet, I need to analyze deletion of experiments

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/helpers/annotations/GeneExpressionAnnotationUniqueIdHelper.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/helpers/annotations/GeneExpressionAnnotationUniqueIdHelper.java
@@ -31,35 +31,22 @@ public class GeneExpressionAnnotationUniqueIdHelper {
 	// used to consolidate annotation DTOs before validating
 	public String generateHash(GeneExpressionFmsDTO geneExpressionFmsDTO) {
 		UniqueIdGeneratorHelper uniqueIdGeneratorHelper = new UniqueIdGeneratorHelper();
-		if (ObjectUtils.isNotEmpty(geneExpressionFmsDTO.getAssay())) {
-			uniqueIdGeneratorHelper.add(geneExpressionFmsDTO.getAssay());
-		}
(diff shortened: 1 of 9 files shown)


src/main/java/org/alliancegenome/curation_api/services/helpers/annotations/GeneExpressionAnnotationUniqueIdHelper.java
```diff
@@ -1,37 +1,75 @@
...
+			uniqueIdGeneratorHelper.add(consolidatedGeneExpressionFmsDTO.getWhereExpressed().getAnatomicalStructureTermId());
+			uniqueIdGeneratorHelper.add(consolidatedGeneExpressionFmsDTO.getWhereExpressed().getCellularComponentTermId());
+		}
+		return uniqueIdGeneratorHelper.getUniqueId();
+	}
+
+	// used to consolidate annotation DTOs before validating
+	public String generateHash(GeneExpressionFmsDTO geneExpressionFmsDTO) {
+		UniqueIdGeneratorHelper uniqueIdGeneratorHelper = new UniqueIdGeneratorHelper();
+		if (ObjectUtils.isNotEmpty(geneExpressionFmsDTO.getAssay())) {
```
markquintontulloch: I don't think ObjectUtils.isNotEmpty() works as you may think here - I've run into issues with this in the past.  You don't actually need some of these null checks anyway as the generator will handle that, you just need null checks when you're calling a field on an object that may be null.  Basically, this should look like the method above.

abecerra: Changed and added the null check, thanks for discovering it!

markquintontulloch (approved):

src/main/resources/db/migration/v0.38.0.36__gene_expression_crossreferences.sql
```diff
@@ -0,0 +1,21 @@
+
```
oblodgett, line 1: @abecerra this should have been v0.39.0.1

PR #1951, opened 2025-04-01
SCRUM-4937 adjust alleles and models output

cmpich:


oblodgett (approved):

oblodgett (changes requested):
Do you want this to go into production not alpha?

cmpich:
Yes, it needs to go into production.

PR #1952, opened 2025-04-01
SCRUM-4937 adjust alleles and models output

cmpich:


oblodgett (approved):

oblodgett (changes requested):
your branch name has to be `release/v0.37.17`

PR #1968, opened 2025-04-05
SCRUM-4073 create endpoint to feed indexer for models table

cmpich:


src/main/java/org/alliancegenome/curation_api/model/document/es/ModelDocument.java
```diff
@@ -0,0 +1,43 @@
...
+import org.alliancegenome.curation_api.model.entities.Organization;
+import org.alliancegenome.curation_api.model.entities.ontology.DOTerm;
+import org.alliancegenome.curation_api.view.View;
+
+import java.util.*;
+
+@Data
+@EqualsAndHashCode(callSuper = true)
+@JsonView(View.ModelDocumentView.class)
+public class ModelDocument extends ESDocument {
```
oblodgett: Ideally this would be called GeneModelDocument

src/main/java/org/alliancegenome/curation_api/view/View.java
```diff
@@ -150,4 +150,7 @@ public static Class<?> viewLookup(String name) {
 		}
 		return ForPublic.class;
 	}
+
+	public class ModelDocumentView {
```
oblodgett, line 154: And this view would match the class name as in GeneModelDocument.

src/main/java/org/alliancegenome/curation_api/model/document/es/ModelDocument.java
cmpich: Better would be: AffectedGenomicModelDocument as that's what this is.

cmpich: I think, AffectedGenomicModelDocument would be more in line with it.

src/main/java/org/alliancegenome/curation_api/view/View.java
cmpich: AffectedGenomicModelView would make sense but that already exists.

oblodgett: Adding the word view onto the end of the class name is redundant.

oblodgett (approved):

PR #1975, opened 2025-04-10
Switch curation from java 17 to java 21

oblodgett:


Dockerfile
```diff
@@ -10,7 +10,7 @@ WORKDIR /agr_curation/cliapp
 RUN make all build
 
 ### Stage 2: build API (and include UI components)
-FROM maven:3.8-eclipse-temurin-17 as BUILD_API_STAGE
+FROM maven:3.9.9-sapmachine-21 as BUILD_API_STAGE
```
abecerra: Why not `maven:3.9-eclipse-temurin-21`?

abecerra (approved):

Dockerfile
oblodgett: Sure, I didn't see that one on the list.

PR #1982, opened 2025-04-14
Fixed typo

oblodgett:


adamgibs (approved):

oblodgett:
@adamgibs seems odd that a change in a text string in the UI causes the backend integration tests to fail.

adamgibs:
@oblodgett This same test failed for me the other day. I reran it a couple of times and it eventually worked. I'm not sure why.

oblodgett:
Ya... looks like this list:
```java
body("entity.alleleVariantAssociations[0].relation.name", is("has_variant")).
body("entity.alleleVariantAssociations[0].alleleAssociationSubject.primaryExternalId", is("WB:AlleleWithVar1")).
body("entity.alleleVariantAssociations[1].relation.name", is("has_variant")).
body("entity.alleleVariantAssociations[1].alleleAssociationSubject.primaryExternalId", is("WB:AlleleWithVar2")).
```
might not be coming out in the same order each time. We can't be guaranteed order on lists like this.

PR #1986, opened 2025-04-17
SCRUM-4946 Enable multiple asserted alleles

markquintontulloch:


src/main/resources/db/migration/v0.39.0.1__make_asserted_allele_multivalued.sql
```diff
@@ -0,0 +1,25 @@
+CREATE TABLE agmphenotypeannotation_allele (
```
oblodgett, line 1: I think we want to keep this in the 0.38 series?

oblodgett (approved):

src/main/resources/db/migration/v0.39.0.1__make_asserted_allele_multivalued.sql
markquintontulloch: We're on v37 now on production.  I think we have a deployment of alpha scheduled that won't include this (SCRUM-5014), so when this hits production we should be on v39.  Unless we're skipping the R13 S2 deployment.

oblodgett: Not sure given that we did not deploy last sprint's stuff and we are on hold for the current sprint due to 8.1.0, maybe I am getting my facts mixed?

oblodgett (approved):

PR #1987, opened 2025-04-17
SCRUM-4817 consolidate models record by model and experimental condition

cmpich:


src/main/java/org/alliancegenome/curation_api/model/document/es/AffectedGenomicModelDocument.java
```diff
@@ -22,7 +22,7 @@ public class AffectedGenomicModelDocument extends ESDocument {
 	private AffectedGenomicModel model;
 	private List<String> associatedPhenotype = new ArrayList<>();
 
-	private Set<DOTerm> diseaseTerms;
+	private Set<DOTerm> diseaseTerms = new HashSet<>();
```
oblodgett: Typically we do this when its needed so that the document doesn't get a "null field".

oblodgett (approved):

PR #1990, opened 2025-04-21
SCRUM-4073 fix consolidation logic

cmpich:


src/main/java/org/alliancegenome/curation_api/model/entities/ExperimentalCondition.java
```diff
@@ -83,4 +83,8 @@ public class ExperimentalCondition extends UniqueIdAuditedObject {
 	@JsonView({View.FieldsOnly.class})
 	@Column(columnDefinition = "TEXT")
 	private String conditionFreeText;
+
+	public String getUniqueExperimentId() {
+		return conditionSummary;
+	}
```
oblodgett: This is quite confusing as it is basically remapping the name of a variable. This method is not needed, just get the string directly.

src/main/java/org/alliancegenome/curation_api/model/entities/ConditionRelation.java
```diff
@@ -66,6 +68,13 @@ public void addExperimentCondition(ExperimentalCondition experimentalCondition)
 		conditions.add(experimentalCondition);
 	}
 
+	public String getUniqueExperimentConditionId() {
+		if (CollectionUtils.isEmpty(conditions)) {
+			return "";
+		}
+		return conditions.stream().map(experimentalCondition -> conditionRelationType.getName() + ":" + experimentalCondition.getUniqueExperimentId()).collect(Collectors.joining(", "));
```
oblodgett: IT's are throwing a null pointer on this line:
```
com.fasterxml.jackson.databind.JsonMappingException: Cannot invoke "org.alliancegenome.curation_api.model.entities.VocabularyTerm.getName()" because "this.conditionRelationType" is null (through reference chain: org.alliancegenome.curation_api.model.entities.ConditionRelation["uniqueExperimentConditionId"])
```

cmpich: Which variable can be null? 
conditions
is not null after the fi-clause

oblodgett (approved):

oblodgett (commented):
Does this need to go out with 8.1.0?

oblodgett (changes requested):
Does this need to go out with 8.1.0?

cmpich:
No, this does not need to go out with 8.1.0

oblodgett (approved):

PR #1994, opened 2025-04-23
Htp dataset fix

adamgibs:


src/main/java/org/alliancegenome/curation_api/services/base/BaseOntologyTermService.java
```diff
@@ -227,24 +234,28 @@ private void handleSynonyms(OntologyTerm dbTerm, OntologyTerm incomingTerm) {
...
 					synonym = synonymDAO.persist(syn);
 				} else {
 					synonym = response.getSingleResult();
 				}
 				dbTerm.getSynonyms().add(synonym);
+			} else {
+				Synonym dbSynonym = response.getSingleResult();
+				if (!sameSynonym(dbSynonym, syn)) {
+					updateSynonym(dbSynonym, syn);
+				}
```
oblodgett: @adamgibs the other option here is to just always update.

adamgibs: So remove  the equality check and just update in the else statement always?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/services/base/BaseOntologyTermService.java
oblodgett: Ya always update is easier than trying to compute the diff. If this gets to resource intensive we can always revisit it.

PR #2012, opened 2025-04-29
SCRUM-5039: delete gene expression crossrefs for mgi wb and zfin

abecerra:


markquintontulloch (changes requested):
Why is this necessary?  Running the loads should clean up unattached xrefs - if not, then that needs fixing.  Also, using hard-coded ID values in a migration isn't a good idea as these could be theoretically be different on the 3 environments.

abecerra:
> Why is this necessary? Running the loads should clean up unattached xrefs - if not, then that needs fixing. Also, using hard-coded ID values in a migration isn't a good idea as these could be theoretically be different on the 3 environments.

At this point for these mods the cross-refs are duplicated in the annotations and the experiments, automatic deletions trigger a lot of foreign key errors

markquintontulloch (approved):

PR #2030, opened 2025-05-07
Release/v0.39.2

adamgibs:
Fix slow ontology tree

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,6 +40,9 @@ uirunalpha:
 uirunbeta:
 	export API_URL=https://beta-curation.alliancegenome.org; make -B -C src/main/cliapp run; unset API_URL
 
+uirunprod:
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -40,6 +40,9 @@ uirunalpha:
 uirunbeta:
 	export API_URL=https://beta-curation.alliancegenome.org; make -B -C src/main/cliapp run; unset API_URL
 
+uirunprod:
+	export API_URL=https://curation.alliancegenome.org; make -B -C src/main/cliapp run; unset API_URL
```
oblodgett: @adamgibs given that this would be against production I typically an reluctant to have this in the Makefile (oops i was pointed at prod)

oblodgett (dismissed):

Makefile
adamgibs: Sure, I went ahead and removed it.

oblodgett (approved):

PR #2034, opened 2025-05-08
Release/v0.39.3

oblodgett:


src/main/java/org/alliancegenome/curation_api/jobs/executors/GeneOntologyAnnotationExecutor.java
```diff
@@ -32,88 +35,91 @@
...
-		for (GeneOntologyAnnotationDTO modID : dtos) {
-			Long geneID = service.getGeneID(modID, abbr);
-			if (geneID != null) {
-				GeneOntologyAnnotation newGaf = service.insert(modID, abbr).getEntity();
-				if (newGaf != null) {
-					geneGoIdsLoaded.add(newGaf.getId());
-					bulkLoadFileHistory.incrementCompleted();
-				} else {
-					bulkLoadFileHistory.incrementSkipped();
+				annotation.setDbObjectId(prefix + annotation.getDbObjectId());
```
markquintontulloch, line 76: Isn't this going to be missing the colon?

src/main/java/org/alliancegenome/curation_api/jobs/executors/GeneOntologyAnnotationExecutor.java
```diff
@@ -32,88 +35,91 @@
...
-				String prefix = abbr;
-				if (abbr.equalsIgnoreCase("XB")) {
-					prefix = "Xenbase";
-				}
-				if (abbr.equalsIgnoreCase("HUMAN") || abbr.equalsIgnoreCase("MGI")) {
-					prefix = null;
+		Map<String, Set<String>> annotationMap = new HashedMap<>();
+
+		for (GeneOntologyAnnotationDTO annotation : gafData) {
+			if (annotation.getDb().equals(dataProvider.resourceDescriptor)) {
```
oblodgett, line 70: @markquintontulloch this check should work for the Xenbase and then later we use the curiePrefix for the id which is then XB: for everything in the file.

src/main/java/org/alliancegenome/curation_api/jobs/executors/GeneOntologyAnnotationExecutor.java
oblodgett: If you look at `dataProvider.curiePrefix` in the enum:
```java
	RGD("NCBITaxon:10116", "RGD:", 10116, "RGD", "RGD", "Rattus", false, false, false, false),
	MGI("NCBITaxon:10090", "MGI:", 10090, "MGI", "MGI", "Mus", true, false, true, false),
	SGD("NCBITaxon:4932", "SGD:", 4932, "SGD", "SGD", "Saccharomyces", true, false, false, false),
	HUMAN("NCBITaxon:9606", "HGNC:", 9606, "RGD", "RGD", "Homo", false, false, false, false),
```
I would agree that we shouldn't have the `:` as part of the prefix. But I am just trying to use what was in the enum.

markquintontulloch: Ah ok, sorry didn't realise the colon was part of the prefix already

src/main/java/org/alliancegenome/curation_api/jobs/executors/GeneOntologyAnnotationExecutor.java
markquintontulloch: Right, but the old code used Xenbase: for everything

oblodgett: yes the DB aka first field in the file is the dataProvider.resourceDescriptor which is the text "Xenbase".

markquintontulloch: Sorry again, was confused by old code setting fields in the DTO object rather than the database.

markquintontulloch (approved):

PR #2067, opened 2025-05-27
SCRUM-5142 remove indexing of AlleleGeneAssociation for Gene and Allele

cmpich:


oblodgett:
@cmpich it still needs to have the same release number

cmpich:
Git

PR #2081, opened 2025-06-16
SCRUM-4072 convert transgenic allele table: Start

cmpich:


src/main/java/org/alliancegenome/curation_api/dao/AlleleDAO.java
```diff
@@ -68,4 +80,49 @@ public Boolean hasReferencingPhenotypeAnnotations(Long alleleId) {
 		results = agmPhenotypeAnnotationDAO.findIdsByParams(agmPaParams);
 		return CollectionUtils.isNotEmpty(results);
 	}
+
+	public SearchResponse<Allele> findAllAllelesWithConstructs(Pagination pagination, HashMap<String, Object> params) {
```
oblodgett, line 84: Is the reason for this method because the generic method doesn't have a "is not empty" way of querying?

cmpich: Yes.

oblodgett: @cmpich can we add that in a generic way as in

`params.put("field", ""):` or `params.put("field", null);`

Then check on the other size and if its empty string or null added the correct condition?

oblodgett: In fact this code already exists:
```java
if (value == null) {
	restrictions.add(builder.isEmpty(root.get(key)));
```

So all you need to do is:
`params.put("alleleConstructAssociations", null);`
`findByParams(params);`

and you should get the right records

PR #2089, opened 2025-06-18
Added extra graphql endpoint

oblodgett:


src/main/java/org/alliancegenome/curation_api/controllers/crud/GraphQLController.java
```diff
@@ -0,0 +1,36 @@
...
+
+	@Inject GeneService geneService;
+	
+	@Query
+	public ObjectResponse<Gene> getGeneByIdentifier(String geneId) {
+		return geneService.getByIdentifier(geneId);
+	}
+
+	@Query
+	public SearchResponse<Gene> findGeneByParams(Integer page, Integer limit, Map<String, String> params) {
```
cmpich, line 31: Why do you not define the Map in the argument with Map<String, Object>? Then you don't need to convert it here.

oblodgett: GraphQL does not support <String, Object> only <String, String> because "Object" is not defined so it fails.

cmpich (approved):

src/main/java/org/alliancegenome/curation_api/controllers/crud/GraphQLController.java
cmpich: Ok.

PR #2099, opened 2025-06-19
SCRUM-5188 & SCRUM-5211 Fix ontology loads missing terms

markquintontulloch:


src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
```diff
@@ -187,7 +187,7 @@ public T traverse(OWLClass currentTreeNode, int depth, HashSet<String> requiredN
 				currentTerm.setAncestors(new HashSet<>(ancestors));
 			}
 
-			for (OWLClass childTermNode : reasoner.getSubClasses(currentTreeNode, true).entities().collect(Collectors.toList())) {
+			for (OWLClass childTermNode : reasoner.getSubClasses(currentTreeNode, false).entities().collect(Collectors.toList())) {
```
oblodgett, line 190: @markquintontulloch according to the doc:
If direct is true, a NodeSet such that for each class C in the NodeSet the set of reasoner axioms entails DirectSubClassOf(C, ce).
If direct is false, a NodeSet such that for each class C in the NodeSet the set of reasoner axioms entails StrictSubClassOf(C, ce).
Not sure what the difference between Strict and Direct but I thought it was children verse the whole tree?

markquintontulloch: My understanding is that 'true' retrieves direct subclasses only, whereas 'false' retrieves all subclasses.  We seem to have some ontologies (specifically PATO and FBDv) where we weren't reaching many (most) of the subclasses when set to 'true'.

markquintontulloch: Previously that manifested as the messed up tree displays, but now we're doing cleanup it's easier to see that we just weren't loading the terms.

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
oblodgett: PS Might have to change the part_of code, as it wanders down into the axioms, to find children but if this turns on ALL of those we might get dup relationships... I guess we will see.

oblodgett (approved):

PR #2107, opened 2025-06-27
Release/v0.39.18

cmpich:


src/main/java/org/alliancegenome/curation_api/model/document/es/AffectedGenomicModelDocument.java
```diff
@@ -50,15 +58,23 @@ public void addConditionRelations(Collection<ConditionRelation> relations) {
...
 		if (CollectionUtils.isNotEmpty(conditionRelationSet)) {
 			if (conditionRelations == null) {
-				conditionRelations = new HashSet<>();
+				conditionRelations = new ArrayList<>();
 			}
 			conditionRelations.addAll(conditionRelationSet);
 		}
 	}
+
+	public boolean isDiseaseAnnotations() {
```
oblodgett: i would guess these should be named "has" vs "is"?

cmpich: Agree.

adamgibs (approved):

abecerra (approved):

PR #2139, opened 2025-07-09
Release/v0.40.1

cmpich:


src/main/java/org/alliancegenome/curation_api/services/AlleleService.java
```diff
@@ -13,6 +13,7 @@
 import org.alliancegenome.curation_api.exceptions.ValidationException;
 import org.alliancegenome.curation_api.interfaces.base.BasePopularityInterface;
 import org.alliancegenome.curation_api.model.entities.Allele;
+import org.alliancegenome.curation_api.model.entities.associations.AlleleConstructAssociation;
```
devopsagr, line 16: 🚫 **[checkstyle]** <com.puppycrawl.tools.checkstyle.checks.imports.UnusedImportsCheck> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Unused import - org.alliancegenome.curation_api.model.entities.associations.AlleleConstructAssociation.

src/main/java/org/alliancegenome/curation_api/dao/associations/AlleleConstructAssociationDAO.java
```diff
@@ -12,4 +22,61 @@ protected AlleleConstructAssociationDAO() {
 		super(AlleleConstructAssociation.class);
 	}
 
+	public SearchResponse<AlleleConstructAssociation> findAllAssociations(Pagination pagination, HashMap<String, Object> params) {
```
markquintontulloch, line 25: Why can't we just use findByParams here?

markquintontulloch: Perhaps I don't understand what's going on here, but I'm wondering why we want to return cases where :params is null?

cmpich: I cannot use the findByParams as I need a query in which I want to retrieve all AlleleConstructAssociation objects that also have non-null ConstructGenomicEntityAssociation objects.

markquintontulloch: Ok, makes sense.  Could you replace the hard-coded field names with constants from the EntityFieldConstants package.  That way, if the LinkML model changes we only have one place to check - this has caught us out many times in the past.

PR #2143, opened 2025-07-11
Fix Bearer token authentication logic in AuthenticationFilter [SCRUM-5275]

christabone:
This commit fixes the authentication issue where valid Bearer tokens were being rejected with 401 errors. The condition for checking the Authorization header has been corrected to properly validate requests that include the 'Bearer ' prefix.

Changes:
- Fixed inverted logic in Bearer token check (removed negation)
- Fixed off-by-one error in token extraction to skip the space after 'Bearer'

This resolves the 401 authentication errors reported in SCRUM-5275.


adamgibs (approved):

oblodgett (commented):
.

PR #2144, opened 2025-07-14
SCRUM-5200 (update) Clean up DAs with incorrect data providers

markquintontulloch:
These need to be cleaned up manually as shouldn't have cases with OMIM for primary and secondary dataprovider.  Therefore cleanup methods not identifying these as previously loaded annotations.


markquintontulloch:
@claude please review the changes since the initial commit

oblodgett (approved):

PR #2147, opened 2025-07-17
Release/v0.40.1

cmpich:


src/main/java/org/alliancegenome/curation_api/dao/associations/AlleleConstructAssociationDAO.java
```diff
@@ -12,4 +25,69 @@ protected AlleleConstructAssociationDAO() {
 		super(AlleleConstructAssociation.class);
 	}
 
+	public SearchResponse<AlleleConstructAssociation> findAllAssociations(Pagination pagination, HashMap<String, Object> params) {
```
oblodgett, line 28: @cmpich you shouldn't need this method just use findByParams to get the data.

cmpich: I need only those AlleleConstructAssociations that have not null ConstructGenomicEntityAssociation objects. I cannot retrieve that subset with the findByParams method.

PR #2165, opened 2025-07-23
SCRUM-4073 add defulat constructor for inner class DOTermAssocation and getter and setter

cmpich:


src/main/java/org/alliancegenome/curation_api/model/document/es/AffectedGenomicModelDocument.java
```diff
@@ -106,30 +108,24 @@ public void addAssociatedPhenotype(List<String> phenotypeList) {
 		Collections.sort(associatedPhenotype, Comparator.comparing(String::toLowerCase));
 	}
 
+	@Getter
+	@Setter
```
oblodgett: If you use the `@Data` annotation it will include the default constructor.

oblodgett: Also if you use the @AllArgsConstructor, and also you can put those views at the class level then your class will only need to have the fields and nothing else.

oblodgett (approved):

PR #2170, opened 2025-07-24
SCRUM-4073b make public class out of inner class

cmpich:


src/main/java/org/alliancegenome/curation_api/model/document/es/DOTermAssociation.java
```diff
@@ -0,0 +1,20 @@
...
+
+@Data
+@AllArgsConstructor
+@JsonView({View.FieldsOnly.class, View.ForPublic.class, View.ModelDocumentView.class})
+public class DOTermAssociation {
+	@JsonView({View.FieldsOnly.class, View.ForPublic.class, View.ModelDocumentView.class})
+	private DOTerm disease;
+	@JsonView({View.FieldsOnly.class, View.ForPublic.class, View.ModelDocumentView.class})
+	private String associationType;
+	@JsonView({View.FieldsOnly.class, View.ForPublic.class, View.ModelDocumentView.class})
```
oblodgett: These jsonviews don't have to be on each field when its at the class level.

oblodgett (approved):

PR #2211, opened 2025-08-26
changes to the gene expression document builder

VarunReddy1111:


src/main/java/org/alliancegenome/curation_api/model/document/es/GeneExpressionDocument.java
```diff
@@ -22,12 +20,13 @@ public class GeneExpressionDocument extends ESDocument {
...
 	private String location;
 	private String stageName;
 	private MMOTerm assay;
-	private Organization dataProvider;
-	private List<CrossReference> crossReferences;
-	private String stageUberonTerm;
-
+	private String dataProvider;
+	private List<CrossReference> crossRefs;
+	private String stageTermId;
```
devopsagr, line 28: 🚫 **[checkstyle]** <com.puppycrawl.tools.checkstyle.checks.regexp.RegexpSinglelineJavaCheck> <sub>reported by [reviewdog](https://github.com/reviewdog/reviewdog) :dog:</sub><br>Line must not have trailing whitespace

src/main/java/org/alliancegenome/curation_api/model/document/builders/GeneExpressionDocumentBuilder.java
```diff
@@ -29,34 +48,111 @@ public GeneExpressionDocument buildDocument(GeneExpressionAnnotation annotation,
...
 		}
-		expressionDocument.setReference(annotation.getEvidenceItem());
-		expressionDocument.setDataProvider(annotation.getDataProvider());
-		expressionDocument.setGene(annotation.getExpressionAnnotationSubject());
+		expressionDocument.setStageTermId(stageUberonTerm);
+
+		AnatomicalSite whereExpressed = annotation.getExpressionPattern() != null ? annotation.getExpressionPattern().getWhereExpressed() : null;
+
+		if (whereExpressed != null) {
+			if (!whereExpressed.getAnatomicalStructureUberonTermOther()) {
```
oblodgett, line 76: Is this `getAnatomicalStructureUberonTermOther` a boolean?

VarunReddy1111: Yes

src/main/java/org/alliancegenome/curation_api/model/document/builders/GeneExpressionDocumentBuilder.java
```diff
@@ -29,34 +48,111 @@ public GeneExpressionDocument buildDocument(GeneExpressionAnnotation annotation,
...
 			}
 		} else {
-			expressionDocument.setCrossReferences(annotation.getCrossReferences());
+			expressionDocument.setCrossRefs(annotation.getCrossReferences());
+		}
+
+		Gene gene = new Gene();
+		gene.setPrimaryExternalId(annotation.getExpressionAnnotationSubject().getPrimaryExternalId());
+		gene.setGeneSymbol(annotation.getExpressionAnnotationSubject().getGeneSymbol());
+		gene.setTaxon(annotation.getExpressionAnnotationSubject().getTaxon());
```
oblodgett: Why not just use `annotation.getExpressionAnnotationSubject()` and not recreate the gene object?

VarunReddy1111: The gene brings back many fields which we dont use on the public site. These 3 are the fields that are used on the public site. Hence created a new gene object tagging only required fields.

oblodgett: Correct... thats why we use a JSONView for this specific document so that you can get just those three fields, but still use the existing objects

src/main/java/org/alliancegenome/curation_api/model/document/builders/GeneExpressionDocumentBuilder.java
```diff
@@ -29,34 +47,106 @@ public GeneExpressionDocument buildDocument(GeneExpressionAnnotation annotation,
...
 		} else {
-			expressionDocument.setCrossReferences(annotation.getCrossReferences());
+			expressionDocument.setCrossRefs(annotation.getCrossReferences());
+		}
+
+		String stageUberonTerm = null;
+		if (annotation.getExpressionPattern() != null && annotation.getExpressionPattern().getWhenExpressed() != null) {
+
+			List<VocabularyTerm> stageUberonTerms = annotation.getExpressionPattern().getWhenExpressed().getStageUberonSlimTerms();
+			if (isNotEmpty(stageUberonTerms) && isNotEmpty(stageUberonTerms.getFirst())) {
```
cmpich, line 65: What happens to possible alternate terms in this collection?

VarunReddy1111: Assuming that this list only has 1 term which is stage uberon term (When looked up in the fms files, it is just 1 element in the list)

cmpich: Ok, might be good to put a comment into the code or have a convenience method the retrieves just that one element from the collection, e.g.
getStageUberonSlimTerm()
But I am not sure if that is actually true. I think expression can happen in multiple stage ranges. Could we ask curators for that? Sue or Yvonne?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/model/document/builders/GeneExpressionDocumentBuilder.java
VarunReddy1111: Added the comment and confirmed with Chris on this. True that we can have multiple stages/stage terms, but we are only getting a single object and not a list from a FMS and we take this single object and store it in a list

PR #2212, opened 2025-08-27
refine Json setter and getter logic on TransgenicAlleleDocument

cmpich:


oblodgett:
@cmpich this is the work that the builder should do, and not have the business logic inside the dto / data object.

oblodgett (approved):

PR #2225, opened 2025-09-08
Scrum 4076 1

VarunReddy1111:


src/main/java/org/alliancegenome/curation_api/view/View.java
```diff
@@ -129,7 +129,7 @@ public static class PrivateOnlyView {
 	public static class ForPublic {
 	}
 
-	public static class GeneExpressionDocument extends ForPublic {
+	public static class GeneExpressionDocument {
```
oblodgett: If you can put this with the other document views.

oblodgett (changes requested):

oblodgett (approved):

PR #2237, opened 2025-09-18
added AlleleForPublic views

adamgibs:


cmpich (approved):
Is Allele, Construct and SubmittedObject the only objects that need the new JsonView?
I added a small change for my transgenic Alleles controller.

adamgibs:
@cmpich Yeah, I believe that should do. The AlleleForPublic inherits from ForPublic which is already annotated on the other needed fields. The only thing that I'm not sure about is related notes. I'm not sure if that should be fetched separately like the associations or annotated with the view.

PR #2259, opened 2025-10-14
Add targetedGenes field

cmpich:


src/main/java/org/alliancegenome/curation_api/model/input/Pagination.java
```diff
@@ -17,4 +17,9 @@ public long getOffset() {
 	public void increment() {
 		page += 1;
 	}
+
+	// if page = 0 and limit = 0 calculate the count of records
+	public boolean isCountCondition() {
```
oblodgett, line 22: I don't see code making use of this?

cmpich: It's coming in the next PR.

adamgibs (approved):

PR #2268, opened 2025-10-20
SCRUM-5398 add JSON views on RDPs for AGMs

adamgibs:


oblodgett (changes requested):
Hey Adam just wondering if you can refactor the "View" out of the name on the json views.

adamgibs:
@oblodgett On all of the views with "View" in the name or just this one?

oblodgett:
Just looking at the code we have a "Document" Section so these two view should be added to that section in the file. And the "View" should be removed from the name ModelDocumentView and TransgenicAllelesDocumentView.

adamgibs:
@oblodgett I've added the requested changes.

oblodgett (approved):

PR #2284, opened 2025-10-29
Release/v0.43.1

cmpich:


src/main/java/org/alliancegenome/curation_api/services/helpers/GenericOntologyLoadHelper.java
```diff
@@ -314,6 +314,21 @@ public Boolean getBoolean(OWLAnnotationValue owlAnnotationValue) {
 		return ((OWLLiteral) owlAnnotationValue).getLiteral().equals("true");
 	}
 
+	private Synonym findExisitingOrCreateNewSynonym(T term, String name) {
```
cmpich: typo:
findExistingOrCreateNewSynonym

src/main/java/org/alliancegenome/curation_api/model/entities/Reagent.java
```diff
@@ -65,4 +66,9 @@ public class Reagent extends SubmittedObject {
 	@JoinTable(indexes = @Index(name = "reagent_secondaryidentifiers_reagent_index", columnList = "reagent_id"))
 	private List<String> secondaryIdentifiers;
 
+	@FullTextField(analyzer = "autocompleteAnalyzer", searchAnalyzer = "autocompleteSearchAnalyzer", valueBridge = @ValueBridgeRef(type = BooleanValueBridge.class))
+	@KeywordField(name = "placeholder_keyword", aggregable = Aggregable.YES, sortable = Sortable.YES, searchable = Searchable.YES, valueBridge = @ValueBridgeRef(type = BooleanValueBridge.class))
+	@JsonView({View.FieldsOnly.class, View.ForPublic.class, View.TransgenicAllelesDocument.class})
+	@Column(columnDefinition = "boolean default false", nullable = false)
+	private Boolean placeholder;
```
oblodgett, line 73: @cmpich what does this field mean? i would like to see a different name

oblodgett (changes requested):

src/main/java/org/alliancegenome/curation_api/model/entities/Reagent.java
cmpich: It's the name the curators came up with. It means the construct name is not final and should not be displayed.

oblodgett: Sorry got mixed up, it looked like this was added to submittedobject, this is what i get for working from my phone.

oblodgett (approved):

PR #2286, opened 2025-10-30
Changes in gene expression document for consolidated documents

VarunReddy1111:


src/main/java/org/alliancegenome/curation_api/model/document/builders/GeneExpressionDocumentBuilder.java
```diff
@@ -110,8 +110,9 @@ public GeneExpressionDocument buildDocument(GeneExpressionAnnotation annotation,
 		}
 		expressionDocument.setTermIds(termIds);
 		if (annotation.getEvidenceItem() != null && annotation.getEvidenceItem() instanceof Reference reference) {
-			expressionDocument.setPubModID(reference.getPubModID());
-			expressionDocument.setReferenceId(reference.getReferenceID());
+			List<String> referenceIdList = new ArrayList<>();
+			referenceIdList.add(reference.getReferenceID());
+			expressionDocument.setReferenceId(referenceIdList);
```
cmpich: Wouldn't you have to add reference.getReferenceID()
to the list fo reference that exists on expressionDocument? This looks like you are create a list with a single reference for each reference and throughline 115 overwriting the previous one. 115 should be a method like expressionDocument.addReferenceIds(referenceIdList) which adds the new reference to the list of refs on the expressionDocument object.

src/main/java/org/alliancegenome/curation_api/controllers/document/GeneExpressionDocumentController.java
```diff
@@ -29,7 +29,7 @@ public SearchResponse<GeneExpressionDocument> findDocument(Integer page, Integer
 
 		Pagination pagination = new Pagination(page, limit);
 
-		SearchResponse<GeneExpressionAnnotation> resp = geneExpressionAnnotationService.findByParams(pagination, params);
+		SearchResponse<GeneExpressionAnnotation> resp = geneExpressionAnnotationService.findByParams(pagination, params, "expressionAnnotationSubject");
```
cmpich: Doesn't this need to be 
expressionAnnotationSubject.primaryExternalId?

VarunReddy1111: Yes, this ideally should be expressionAnnotationSubject.primaryExternalId, but the DAO is setup currently does not support nested paths. Hence, I chose to order it by gene IDs (expressionAnnotationSubject_id).

src/main/java/org/alliancegenome/curation_api/model/document/builders/GeneExpressionDocumentBuilder.java
VarunReddy1111: Here, earlier there would be a single String object to contain the referenceId of each annotation. Now, this is being changed to List<String>. But, this still is expected to be a singleton List as each annotation contains only 1 referenceId on curation system. The reason it is being changed to List<String> is to accomodate consolidated annotations on the public side where there might be more than 1 reference on the consolidated annotations.

cmpich: Oh, ok. the consolidation is not happening with this code change. IT's happening on the indexer / public site. 
In that case, you could write more succinctly:
expressionDocument.setReferenceId(List.of(reference.getReferenceID()))
making lines 113/114 obsolete.

VarunReddy1111: True, this can be changed. Thank you

src/main/java/org/alliancegenome/curation_api/controllers/document/GeneExpressionDocumentController.java
oblodgett: @VarunReddy1111 are you saying that we don't support nested paths in the order by? this makes sense but definitely a feature we should add.

cmpich (dismissed):

PR #2290, opened 2025-11-12
Consolidation changes for gene expression annotation

VarunReddy1111:


VarunReddy1111:
So, as discussed earlier I have changed the order by field use `expressionAnnotationSubject.primaryExternalId`. But, in this case the curation API response time was getting increased as the page number is increasing and leading to increase in indexing time. This is due to the join between the `geneexpressionannotation` table and `expressionannotationsubject(Gene)`table and as the page number is increased it has to scan all rows in the table and get to the offset  and extract according to pagination. (like, for page 250 and limit 1000 it has scan to 250000 records to get to the offset). Hence, switching back to use `expressionAnnotationSubject` which is ordering by geneIds. 

Also, tried using cursor in pagination with primaryExternalId, but it would not work for primaryExternalId as it is a string and not an Id.

cmpich (approved):

PR #2324, opened 2025-12-10
SCRUM-5650 set 19 MGI markers internal=true

cmpich:


oblodgett (approved):

oblodgett (commented):
wouldn't this be better done as a load?

cmpich:
We decided to do this directly in the database for 8.3.0. MGI may undo this in the future when time has come.

oblodgett (approved):

PR #2330, opened 2025-12-16
SCRUM-5650 do not include alleleGeneAssociations in allel_summary doc…

cmpich:
…s when the association is marked as internal=true


oblodgett (dismissed):

adamgibs:
@cmpich Maybe update the `buildAlleleOfGene `method in the `AlleleSummaryDocumentBuilder `as well? That filter line (line 57) I would guess isn't need if you filter in the DAO.

cmpich:
@adamgibs I am going to leave line 57 for now. I don't want to assume that the allele object in that class already has the internal values for the aga even though it is true in this case but I may call this class / method from somewhere else.

oblodgett (approved):

PR #2332, opened 2025-12-17
ZFIN-5614 peresist note of variant via cascadeType.ALL

cmpich:


src/main/java/org/alliancegenome/curation_api/services/validation/dto/fms/VariantFmsDTOValidator.java
```diff
@@ -225,7 +223,7 @@ public Long validateVariant(VariantFmsDTO dto, List<Long> idsAdded, BackendBulkD
 					String noteIdentity = NoteIdentityHelper.variantNoteFmsDtoIdentity(dto.getNotes().get(ix));
 					if (!noteIdentities.contains(noteIdentity)) {
 						noteIdentities.add(noteIdentity);
-						validatedNotes.add(noteDAO.persist(noteResponse.getEntity()));
+						validatedNotes.add(noteResponse.getEntity());
```
oblodgett, line 226: @cmpich is it possible that the method call created a new note and it needs to be persisted?

cmpich: I am not sure what you mean, @oblodgett 
When you persist an entity separately before adding it to a collection that uses a join table (@JoinTable), Hibernate doesn't automatically create the join table entry because:
  1. The note is already a managed entity with an ID
  2. When the parent (variant) is persisted, Hibernate sees the note is already persisted and doesn't realize it needs to create the join table link

oblodgett: `ObjectResponse<Note> noteResponse = variantNoteFmsDtoValidator.validateVariantNoteFmsDTO(dto.getNotes().get(ix));` may not persist a new entity. That is why it's persisted here, and you are removing the persist.

oblodgett (changes requested):

PR #2349, opened 2026-01-05
SCRUM-5611 add bearer auth to swagger and graphql

adamgibs:


src/main/java/org/alliancegenome/curation_api/config/RestApplication.java
```diff
@@ -16,7 +17,8 @@
 			+ "<h1>For flat file of API documentation click on the link below <a href=\"/openapi\"><h1>OpenAPI definition download</h1></a></h1>",
 		title = "Alliance of Genome Resources Curation API ",
 		version = "1.0 Alpha"
-	)
+	),
+	security = @SecurityRequirement(name = "APIToken")
```
oblodgett, line 21: @adamgibs the only potential issue I see with this is that we accept both Bearer and APIToken...

src/main/resources/application.properties.defaults
```diff
@@ -98,10 +98,12 @@ quarkus.oidc.token.issuer=https://cognito-idp.us-east-1.amazonaws.com/${cognito.
 quarkus.oidc.client-id=""
 quarkus.oidc.user-info-path=""
 
+quarkus.smallrye-openapi.security-scheme=api-key
```
oblodgett, line 101: This might be where we would want to use a cookie. Then swagger would just send it along for the ride.

adamgibs: Yeah, the problem is that quarkus OIDC can't parse the frontend cookie that we send back so it would just error out.

oblodgett (approved):

PR #2384, opened 2026-02-03
Refactored Variant Summary

oblodgett:


src/main/java/org/alliancegenome/curation_api/model/entities/PredictedVariantConsequence.java
```diff
@@ -80,120 +79,93 @@ public class PredictedVariantConsequence extends AuditedObject {
...
 	@FullTextField(analyzer = "autocompleteAnalyzer", searchAnalyzer = "autocompleteSearchAnalyzer", valueBridge = @ValueBridgeRef(type = BooleanValueBridge.class))
 	@KeywordField(name = "geneLevelConsequence_keyword", aggregable = Aggregable.YES, sortable = Sortable.YES, searchable = Searchable.YES, valueBridge = @ValueBridgeRef(type = BooleanValueBridge.class))
-	@JsonView({CurationView.FieldsOnly.class})
+	@JsonView({CurationView.FieldsOnly.class, CurationView.VariantDocument.class})
 	@Column(columnDefinition = "boolean default false", nullable = false)
 	private Boolean geneLevelConsequence = false;
 
-	@Transient
-	@JsonProperty(access = JsonProperty.Access.READ_ONLY)
-	@JsonView({CurationView.FieldsOnly.class, CurationView.VariantView.class})
```
cmpich, line 186: Where does this method / logic go if this method is deleted?

oblodgett: I can't find anywhere in the code where exons or introns are being set, and they are transient which means they are not coming from the DB either.

cmpich: Could you put back this method? I need that convenience method.

cmpich (dismissed):

PR #2429, opened 2026-02-12
SCRUM-5666 add resource descriptor object to variant doc

cmpich:


oblodgett (changes requested):
@cmpich this needs to go into production?

PR #2482, opened 2026-02-26
Update actions permission to write for auto-retry support

christabone:
## Summary
- Update `actions: read` → `actions: write` in Claude Code Review caller workflow
- Required for the auto-retry mechanism which uses `gh run rerun` API (see alliance-genome/.github#7)

## Test plan
- [ ] Merge alliance-genome/.github#7 first
- [ ] Merge this PR
- [ ] Verify auto-retry works on next silent drop

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -22,7 +22,7 @@ jobs:
       pull-requests: write
       issues: write
       id-token: write  # Required for OIDC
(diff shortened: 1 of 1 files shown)


christabone:
Closing — the `actions: write` permission is no longer needed. The auto-retry mechanism was simplified to use notification comments instead of `gh run rerun` (which can't work from within the same running workflow).

PR #2485, opened 2026-02-26
Release/v0.46.18

cmpich:


src/main/java/org/alliancegenome/curation_api/controllers/document/VariantDocumentController.java
```diff
@@ -33,6 +35,7 @@ public SearchResponse<VariantSummaryDocument> findDocuments(Integer page, Intege
 		if (resp.getResults() != null) {
 			VariantDocumentBuilder builder = new VariantDocumentBuilder();
 			for (Variant variant : resp.getResults()) {
+				Hibernate.initialize(variant.getRelatedNotes());
```
oblodgett, line 38: A different trick for this is to just call `variant.getRelatedNotes().size();` however if you have the JsonView on the notes they should be there.

oblodgett (approved):

PR #2492, opened 2026-02-27
SCRUM-5666 add JSON view to introns and exons

cmpich:


oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/model/entities/PredictedVariantConsequence.java
```diff
@@ -197,4 +197,18 @@ public String getIntronExonLocation() {
 		}
 		return "Exon " + exons + " : " + "Intron " + introns;
 	}
+
+	@Transient
+	@JsonProperty(access = JsonProperty.Access.READ_ONLY)
+	@JsonView({CurationView.FieldsOnly.class, CurationView.VariantSummaryDocument.class})
+	public String getSequenceFeatureName() {
```
oblodgett, line 204: It would be better to push display related functionality forward to the UI and not end up with all these small methods all over, that have no connection to the java code. It seems to me that you would want to do this on either input with another field or output (UI) but not in the middle.

cmpich: I agree, these convenience methods do spoil the entity classes but the logic what names certain entities have should not be hard-coded in the UI.

PR #2506, opened 2026-03-02
changes to variant summary document to mirror allelesummary document

VarunReddy1111:


cmpich (dismissed):

oblodgett:
@VarunReddy1111 I think we need to step back for a second on this and the variant variable in the variant summary needs to be a Variant Object. @cmpich I should have said this when you first created this class.

src/main/java/org/alliancegenome/curation_api/model/document/builders/VariantDocumentBuilder.java
```diff
@@ -21,6 +21,10 @@ public List<VariantSummaryDocument> buildVariantDocument(Variant variant) {
 					variant.getAlleleVariantAssociations().forEach(alleleVariantAssociation -> {
 						VariantSummaryDocument dto = new VariantSummaryDocument();
 						dto.setVariant(curatedVariantGenomicLocationAssociation);
+						Variant variantWrapper = new Variant();
+						variantWrapper.setVariantType(variant.getVariantType());
+						variantWrapper.setCuratedVariantGenomicLocations(List.of(curatedVariantGenomicLocationAssociation));
+						dto.setVariants(List.of(variantWrapper));
```
oblodgett: You also what to put the variant here not a wrapper variant

oblodgett (approved):

PR #2535, opened 2026-03-11
SCRUM-5834: Fix allele synonym/symbol/fullname editing in table inter…

adamgibs:
…face

The curation table dialogs for synonyms, symbols, and full names were broken after the PrimeReact 8→10 migration — row editing had stale closure races, rows swapped positions, and Keep Edits was unresponsive.

Split each dialog into Edit + ReadOnly components using the always-edit DataTable pattern (no per-row save/cancel) with batch validation on Keep Edits. Read-only dialogs are simple display tables.

Also fixes:
- rowIndex % rows bug in all handleXXXOpenInEdit functions across AllelesTable, GenesTable, and VariantsTable
- Person.emails/oldEmails LazyInitializationException on validate endpoints via @JsonProperty(access = WRITE_ONLY)
- Main table cell not re-rendering after dialog save (synonym count and Pending Edits warning) via PrimeReact editorCallback()

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/model/entities/Person.java
+++ b/src/main/java/org/alliancegenome/curation_api/model/entities/Person.java
@@ -12,7 +12,6 @@ import org.hibernate.search.mapper.pojo.mapping.definition.annotation.FullTextFi
 import org.hibernate.search.mapper.pojo.mapping.definition.annotation.KeywordField;
 
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonView;
 
 import jakarta.persistence.Column;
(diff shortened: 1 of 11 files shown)


src/main/java/org/alliancegenome/curation_api/model/entities/Person.java
```diff
@@ -66,13 +67,13 @@ public class Person extends Agent {
 
 	@KeywordField(aggregable = Aggregable.YES, sortable = Sortable.YES, searchable = Searchable.YES)
 	@ElementCollection
-	@JsonView({ CurationView.FieldsAndLists.class, CurationView.PersonSettingView.class })
+	@JsonProperty(access = JsonProperty.Access.WRITE_ONLY)
```
oblodgett: @adamgibs do we need a view on this? I would prefer to just take it off.

oblodgett: Here is what the review said: 

The previous annotation was @JSONVIEW({ CurationView.FieldsAndLists.class, CurationView.PersonSettingView.class }). Replacing it with @JsonProperty(access = WRITE_ONLY) is the right fix for the LazyInitializationException on the validate endpoint, but WRITE_ONLY suppresses serialisation in every response including the PersonSettingView that previously exposed these fields. The person_emails table exists and holds data but it will no longer be returned by any GET. Please confirm no front-end code reads emails back from the API before merging.

oblodgett (approved):

PR #2536, opened 2026-03-11
End point to retrieve all the variant names

VarunReddy1111:


src/main/java/org/alliancegenome/curation_api/interfaces/document/AlleleDocumentInterface.java
```diff
@@ -27,6 +28,10 @@ public interface AlleleDocumentInterface {
 	@JsonView(CurationView.AlleleSummaryDocument.class)
 	SearchResponse<Long> getAllIds();
 
+	@GET
+	@Path("/variantIds")
```
oblodgett: Here you call the endpoint variantIds however the method name is getNames choose one.

src/main/java/org/alliancegenome/curation_api/interfaces/document/AlleleDocumentInterface.java
```diff
@@ -27,6 +28,10 @@ public interface AlleleDocumentInterface {
 	@JsonView(CurationView.AlleleSummaryDocument.class)
 	SearchResponse<Long> getAllIds();
 
+	@GET
+	@Path("/variantNames")
+	ObjectListResponse<String> getAllVariantNames();
+
```
oblodgett: Sorry @VarunReddy1111 we have a VariantDocumentInterface I would think this method should go there then it would be /api/variant/allNames or something like that

oblodgett (approved):

PR #2557, opened 2026-03-17
SCRUM-5834: Fix validation error rendering and evidence autocomplete on allele detail page

adamgibs:


oblodgett (commented):
I didn't know the UI used these endpoints directly. Can we pull this out into a constant?

adamgibs:
@oblodgett I thought about that, but if we do that, we may as well create a constants file for all of the UI search endpoint strings, which will touch a lot of files.

oblodgett:
@adamgibs but are they all not just a single endpoint or are they all used more than once? I mean this is in 9 different places, is it that way for all of them?

adamgibs:
@oblodgett Yeah, they're scattered throughout the code. I think this would be a good change. It just may need a lot of testing to make sure nothing broke. May be better in it's own PR.

oblodgett (approved):
Wow I didn't realize it was that bad.

PR #2561, opened 2026-03-18
Release/v0.47.4

cmpich:


src/main/java/org/alliancegenome/curation_api/model/document/es/AVSParentDocument.java
```diff
@@ -15,4 +16,6 @@ public abstract class AVSParentDocument extends ESDocument {
 	protected Integer alterationTypeSortOrder;
 	protected Boolean hasPhenotype;
 	protected Boolean hasDisease;
+	protected Set<String> diseases;
```
oblodgett, line 19: Do more than one object need these?

cmpich: Yes. The DTO is used in the DAO class which then is used in the AlleleSummaryDocumentBuilder to ppopulate the AVSParentDocument that gets sent out for the API call.

cmpich:
Removed in lieu of v0.47.2

PR #2562, opened 2026-03-18
Release/v0.47.2

cmpich:


src/main/java/org/alliancegenome/curation_api/model/document/es/AVSParentDocument.java
```diff
@@ -15,4 +16,6 @@ public abstract class AVSParentDocument extends ESDocument {
 	protected Integer alterationTypeSortOrder;
 	protected Boolean hasPhenotype;
 	protected Boolean hasDisease;
+	protected Set<String> diseases;
+	protected Set<String> diseasesAgrSlim;
```
oblodgett: Are these on all three subobjects? And changes to the SSD cause drastic changes to the variant indexer, I would think these should just be on the AlleleSummaryDocument?

cmpich: Yes, for now I can move it into the one child class. I was not sure if it would be needed in the other child classes. We can add them when needed.

oblodgett (dismissed):

oblodgett (approved):

PR #2571, opened 2026-03-20
SCRUM-5858: Produce warnings instead of errors for unrecognized refer…

adamgibs:
…ences/evidence across all bulk loads

Unrecognized references and evidence curies in bulk loads now produce warnings instead of hard errors, allowing entities to still load. Applies to all entity types: allele, gene, construct, AGM, and variant.

Introduces a generic validateOptionalEntities() helper in BaseDTOValidator that takes a lookup function and always warns on missing entities. This replaces the old validateReferences() method and the inline evidence loop in SlotAnnotationDTOValidator, eliminating duplicated validation logic.

Warning propagation added to all entity DTO validators so sub-validator warnings bubble up to load results.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/validation/dto/base/BaseDTOValidator.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/validation/dto/base/BaseDTOValidator.java
@@ -447,7 +447,8 @@ public class BaseDTOValidator<E extends Object> {
 			return null;
 		}
 		if (noteResponse.hasWarnings()) {
-			response.addWarningMessages(noteResponse.getWarningMessages());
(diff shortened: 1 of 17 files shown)


src/main/java/org/alliancegenome/curation_api/services/validation/dto/base/BaseDTOValidator.java
```diff
@@ -50,7 +51,7 @@ public class BaseDTOValidator<E extends Object> {
 	@Inject
 	MiTermService miTermService;
 	@Inject
-	ReferenceService referenceService;
+	protected ReferenceService referenceService;
```
oblodgett: Shouldn't need protected here as the rest of the variables are not.

oblodgett (approved):

PR #2583, opened 2026-03-24
Release/v0.47.8

cmpich:


src/main/java/org/alliancegenome/curation_api/model/document/es/AlleleSummaryDTO.java
```diff
@@ -18,6 +18,9 @@ public class AlleleSummaryDTO {
 	private Boolean hasDisease;
 	private Set<String> diseases;
 	private Set<String> diseasesAgrSlim;
+	private Set<String> constructExpressedComponents;
```
oblodgett, line 21: One question I have had is why use this DTO class when you could just return the Document class directly?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/model/document/es/AlleleSummaryDTO.java
cmpich: Makes sense to remove the DTO class.

PR #2588, opened 2026-03-24
SCRUM-5825: Fix stopped load Slack notifications and sort dependencies alphabetically

VarunReddy1111:
## Summary
- Clear thread interrupt flag before sending Slack notification in `endLoad`, fixing `InterruptedIOException` that prevented stopped load notifications from being sent
- Sort dependent files alphabetically in the bulk load edit dialog

## Test plan
- [ ] Start a bulk load (e.g. SGD Gene), stop it, and verify Slack notification is sent
- [ ] Open the bulk load edit dialog and verify dependencies are listed in alphabetical order

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/jobs/processors/BulkLoadProcessor.java
+++ b/src/main/java/org/alliancegenome/curation_api/jobs/processors/BulkLoadProcessor.java
@@ -215,7 +215,6 @@ public class BulkLoadProcessor {
 		bulkLoadFileHistory.setErrorMessage(message);
 		bulkLoadFileHistory.setBulkloadStatus(status);
 		bulkLoadFileHistory.setLoadFinished(LocalDateTime.now());
-		Thread.interrupted();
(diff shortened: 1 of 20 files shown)


src/main/java/org/alliancegenome/curation_api/jobs/processors/BulkLoadProcessor.java
```diff
@@ -215,6 +215,7 @@ protected void endLoad(BulkLoadFileHistory bulkLoadFileHistory, String message,
 		bulkLoadFileHistory.setErrorMessage(message);
 		bulkLoadFileHistory.setBulkloadStatus(status);
 		bulkLoadFileHistory.setLoadFinished(LocalDateTime.now());
+		Thread.interrupted();
```
oblodgett: I don't think this is what we want to do.

oblodgett (changes requested):

oblodgett (approved):

PR #2592, opened 2026-03-25
Release/v0.47.11

cmpich:


src/main/resources/db/migration/v0.47.0.1__add_soterm_severity_order.sql
```diff
@@ -0,0 +1,48 @@
+ALTER TABLE ontologyterm ADD COLUMN IF NOT EXISTS severityorder integer;
```
oblodgett, line 1: We can't do this migration, it will prevent all the ones from alpha from running.

oblodgett (approved):

PR #2594, opened 2026-03-25
Add Load Order Dependencies page to curation frontend

adamgibs:
New page under "Data files and reports" that visualizes the load order dependencies from the README as an interactive SVG chart. Nodes are color-coded by category (Entity, Association, CrossReference, GFF, Orthology, VEP) with hover highlighting to trace dependency paths. Includes a warning about Gene loads clearing cross-references and a last-updated date for tracking data freshness.


cmpich (approved):

oblodgett (commented):
This is pretty cool... to bad it wasn't dynamic from the backend.

adamgibs:
> This is pretty cool... to bad it wasn't dynamic from the backend.

Yeah, I decided not to do that now to keep it simple. The data from the backend is MOD specific so the UI code would still need to turn that into the higher level load data. It does mean that this page will need to be manually updated when we add more loads or the load dependencies change, but that should be fairly straightforward. Also, we have to update the README in that case as well.

PR #2640, opened 2026-04-06
SCRUM-5928 migrate from CRA to Vite

adamgibs:
- Migrate curation frontend from Create React App to Vite (faster dev server, HMR, builds)
  - Replace axios with a lightweight native fetch wrapper (ApiClient.js)
  - Replace moment.js/react-moment with dayjs
  - Rename 327 .js files to .jsx (Vite requires .jsx extension for JSX)
  - Migrate Jest tests to Vitest
  - Migrate Sass @import to @use (fixes deprecation warnings)
  - Replace CRA-specific patterns: process.env → import.meta.env, global.* → native browser globals
  - Remove dead code: serviceWorker.js, gulpfile.js, setupProxy.js, IE11 polyfill, core-js imports
  - Update CI workflow for Vitest

  No changes to Dockerfile, Makefiles, or public/ assets. Build output structure unchanged.


src/main/cliapp/src/containers/dataLoadsPage/DataLoadsComponent.jsx
```diff
@@ -144,7 +143,7 @@ export const DataLoadsComponent = () => {
 			} else {
 				new_uri = 'ws:';
 			}
-			if (process.env.NODE_ENV === 'production') {
+			if (import.meta.env.PROD) {
```
oblodgett, line 146: curious about this change? We set this in amplify.

src/main/cliapp/src/AppTopbar.jsx
```diff
@@ -29,7 +29,7 @@ export const AppTopbar = (props) => {
 	} else {
 		new_uri = 'ws:';
 	}
-	if (process.env.NODE_ENV === 'production') {
+	if (import.meta.env.PROD) {
```
oblodgett, line 32: Same question as before.

oblodgett (approved):

src/main/cliapp/src/AppTopbar.jsx
adamgibs: @oblodgett I don't think the curation UI uses Amplify.

oblodgett: Right...

PR #2647, opened 2026-04-07
SCRUM-5829 rename variants to variantList

cmpich:


oblodgett (commented):
This needs to happen on the other documents also... I'll put the code info for those.

oblodgett (approved):

PR #2656, opened 2026-04-10
KANBAN-965: exclude bad VMA21 orthology pairs

christabone:
## Summary
This change blocks a known bad DIOPT orthology mapping between VMA21 genes and C. elegans `pdcd-2` from being reloaded into `agr_curation`. It also adds regression coverage for the stale-row cleanup path so future orthology reloads prove that pre-existing bad rows are removed when the excluded pair is skipped.

## Changes
- Added an explicit exclusion list for the eight known bad VMA21/pdcd-2 orthology pairs in the orthology FMS validator.
- Added an orthology fixture for the excluded MGI to WB pair.
- Extended the orthology integration test to verify that the excluded pair is skipped during ingest.
- Extended the orthology integration test to create a stale bad orthology row, run the executor-backed bulk load flow, and assert that cleanup deletes the stale row.

## Testing
- Added integration coverage in `IT_0602_OrthologyBulkUploadFmsITCase` for both skipped ingest and stale-row cleanup.
- Local test execution was not possible in this session because the environment does not provide `mvn` or a repo `mvnw` wrapper.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/services/validation/dto/fms/OrthologyFmsDTOValidator.java
+++ b/src/main/java/org/alliancegenome/curation_api/services/validation/dto/fms/OrthologyFmsDTOValidator.java
@@ -37,7 +37,11 @@ public class OrthologyFmsDTOValidator {
 	}
 
 	// KANBAN-965: temporary local exclusion for known bad VMA21/pdcd-2 DIOPT data.
-	// Remove this once the upstream DIOPT data is corrected and the bad pairs stop appearing.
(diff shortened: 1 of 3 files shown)


oblodgett:
@christabone does this need to happen for 9.0.0? if so you would need to branch off production, with a release branch. Curation is currently at v.0.47.36

christabone:
@oblodgett Yes, I was hoping for 9.0.0. I can move things around with the branches.

oblodgett:
Ok then you need to create branch `release/v0.47.37` off of production and PR it back into production. Once done the ortho loads will need to be run on prod, but I would coordinate with @chris-grove about that.

PR #2657, opened 2026-04-10
KANBAN-965: exclude bad VMA21 orthology pairs

christabone:
## Summary
This change blocks the known bad DIOPT orthology mapping between VMA21 genes and C. elegans `pdcd-2` from being reloaded into `agr_curation`. It also adds regression coverage for the stale-row cleanup path so future orthology reloads prove that pre-existing bad rows are removed when the excluded pair is skipped.

This replaces the earlier `alpha` PR approach from #2656. Per branch strategy feedback, this release-targeted fix is now based on `production` and opened against `production` on `release/v0.47.37`.

## Changes
- Added an explicit exclusion list for the eight known bad VMA21/pdcd-2 orthology pairs in the orthology FMS validator.
- Added an orthology fixture for the excluded MGI to WB pair.
- Extended the orthology integration test to verify that the excluded pair is skipped during ingest.
- Extended the orthology integration test to create a stale bad orthology row, run the executor-backed bulk load flow, and assert that cleanup deletes the stale row.
- Added comments marking the workaround and where to remove it once DIOPT corrects the upstream data.

## Testing
- Added integration coverage in `IT_0602_OrthologyBulkUploadFmsITCase` for both skipped ingest and stale-row cleanup.
- Local test execution was not possible in this session because the environment does not provide `mvn` or a repo `mvnw` wrapper.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/jobs/executors/OrthologyExecutor.java
+++ b/src/main/java/org/alliancegenome/curation_api/jobs/executors/OrthologyExecutor.java
@@ -12,9 +12,6 @@ import org.alliancegenome.curation_api.model.entities.bulkloads.BulkFMSLoad;
 import org.alliancegenome.curation_api.model.entities.bulkloads.BulkLoadFileHistory;
 import org.alliancegenome.curation_api.model.entities.orthology.GeneToGeneOrthologyGenerated;
 import org.alliancegenome.curation_api.model.ingest.dto.fms.OrthologyIngestFmsDTO;
-import org.alliancegenome.curation_api.model.ingest.dto.fms.OrthologyFmsDTO;
-import org.alliancegenome.curation_api.response.APIResponse;
-import org.alliancegenome.curation_api.response.LoadHistoryResponce;
 import org.alliancegenome.curation_api.services.orthology.GeneToGeneOrthologyGeneratedService;
 import org.apache.commons.lang3.StringUtils;
 
@@ -63,28 +60,5 @@ public class OrthologyExecutor extends LoadFileExecutor {
 			failLoad(bulkLoadFileHistory, e);
 			e.printStackTrace();
 		}
-	}
-
-	public APIResponse runLoadApi(String dataProviderName, List<OrthologyFmsDTO> orthologyData, Boolean cleanUp) {
-		List<Long> orthoPairIdsLoaded = new ArrayList<>();
-		BulkLoadFileHistory history = new BulkLoadFileHistory(orthologyData.size());
-		history = bulkLoadFileHistoryDAO.persist(history);
-
-		BackendBulkDataProvider dataProvider = BackendBulkDataProvider.valueOf(dataProviderName);
-		List<Long> orthoPairIdsBefore = new ArrayList<>();
-		if (Boolean.TRUE.equals(cleanUp)) {
-			orthoPairIdsBefore = generatedOrthologyService.getAllOrthologyPairIdsBySubjectGeneDataProvider(dataProvider);
-			log.debug("runLoadApi: Before: total " + orthoPairIdsBefore.size());
-		}
-
-		boolean success = runLoad(generatedOrthologyService, history, dataProvider, orthologyData, orthoPairIdsLoaded);
-		if (success && Boolean.TRUE.equals(cleanUp)) {
-			runCleanup(generatedOrthologyService, history, dataProvider.name(), orthoPairIdsBefore, orthoPairIdsLoaded, "orthology");
-		}
-
-		history.finishLoad();
-		updateHistory(history);
-		updateExceptions(history);
-		return new LoadHistoryResponce(history);
 	}
 }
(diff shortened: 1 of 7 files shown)


oblodgett (dismissed):

src/test/java/org/alliancegenome/curation_api/IT_0602_OrthologyBulkUploadFmsITCase.java
```diff
@@ -176,4 +284,146 @@ public void orthologyBulkUploadUpdateEmptyNonRequiredFields() throws Exception {
...
+	private Path createGzipFromResource(String resourcePath) throws Exception {
+		Path gzipFile = Files.createTempFile("kanban-965-orthology-", ".json.gz");
+		gzipFile.toFile().deleteOnExit();
+		try (GZIPOutputStream outputStream = new GZIPOutputStream(Files.newOutputStream(gzipFile))) {
+			outputStream.write(Files.readString(Path.of(resourcePath)).getBytes(StandardCharsets.UTF_8));
+		}
+		return gzipFile;
+	}
+
+	private Set<Integer> getHistoryIdsForBulkLoad(Long bulkLoadId) {
```
oblodgett: @christabone I don't think you need to do all this bulk load history stuff as the bulk load via API uses a different path. We don't test the actual upload of a file. We just test the raw data through the API. It hits the same path on the backend so all backend code gets tested.

christabone: Moved this fully onto the raw orthology bulk API in `b7340098c`. The endpoint now accepts `cleanUp=true`, so the test no longer touches DQM/manual-load history or the S3-backed upload path. It seeds the stale MGI/WB row, posts the excluded fixture through `/api/orthologygenerated/bulk/MGI/orthologyfile?cleanUp=true`, asserts the skipped record plus cleanup delete count, and then waits for that stale row to disappear.

src/main/java/org/alliancegenome/curation_api/jobs/executors/OrthologyExecutor.java
```diff
@@ -61,4 +64,27 @@ public void execLoad(BulkLoadFileHistory bulkLoadFileHistory) {
 			e.printStackTrace();
 		}
 	}
+
+	public APIResponse runLoadApi(String dataProviderName, List<OrthologyFmsDTO> orthologyData, Boolean cleanUp) {
```
oblodgett: Im not understanding why this is needed we generally don't worry about clean and such with integration tests. I just would assume you would fix the text to have 2 entries vs 1? and not have all this extra code?

oblodgett: Plus the:
      updateHistory(history);
      updateExceptions(history);
  Is only in the context of a file upload, when we hit the endpoints with data there is no "file". that's what we do with all the integration tests.

christabone: The regression here is specifically about a stale excluded orthology already existing in the DB and needing to be removed on the next MGI load. The test seeds that stale row on purpose, then verifies the cleanup path deletes it. If I just change the later assertions from 1 to 2, I’m effectively accepting the bad row persisting forever and no longer testing the regression. I added cleanUp=true as an opt-in on the raw orthology bulk API so the test can stay on the same API path you asked for, instead of using the DQM/manual-load path.

christabone: @oblodgett If you want, I can take that cleanup piece back out and keep the test focused only on skipping the excluded pair on load. The tradeoff is that we’d stop testing cleanup of an already-existing bad row.

oblodgett: So generally, this being a data cleanup issue we wouldn't run tests for it. Or we would just run it directly through the API via an update, and see that the API fails the update.

oblodgett: So the:
```
throw new KnownIssueValidationException("Skipping excluded orthology pair: "
				+ subjectGeneIdentifier + " <-> " + objectGeneIdentifier);
```
would fire if it was coming through a file load or via a upsert via the API. So typically we would test the upsert, and the test would be looking for the failure: which is the above exception, in order for the test to pass. Then none of the other code needs to change.

oblodgett: In other words we would test: `orthologyBulkUploadExcludedKnownIssuePair` and not `orthologyBulkExecutorCleanupDeletesExcludedKnownIssuePair`

oblodgett: We don't test the cleanup's so in this case, on the load all the id's of the one that get updated stay and everything else goes. We don't integration test this, as this gets tested all the time with our loads.

christabone:
@oblodgett All revamped now.

oblodgett (approved):

PR #2710, opened 2026-05-05
SCRUM-6042 do not serialized crossReferences on overlapGenes field on…

cmpich:
… the Variant object


oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/model/entities/associations/CuratedVariantGenomicLocationAssociation.java
```diff
@@ -123,6 +124,7 @@ public List<String> getHgvsP() {
 	@Transient
 	@JsonProperty(access = JsonProperty.Access.READ_ONLY)
 	@JsonView({CurationView.VariantSummaryDocument.class})
+	@JsonIgnoreProperties({"crossReferences"})
```
oblodgett, line 127: Why not remove the view vs ignoring it?

cmpich: Because we need the overlappingGene object.

PR #2723, opened 2026-05-07
SCRUM-6042 move to annotation based suppression of gene cross references

cmpich:


src/main/java/org/alliancegenome/curation_api/model/document/es/VariantSummaryDocument.java
```diff
@@ -24,41 +24,14 @@ public class VariantSummaryDocument extends AVSParentDocument {
 	}
 	private List<Variant> variantList;
 	private HashSet<String> geneSynonyms;
+
+	@JsonIgnore
 	private HashSet<String> geneCrossReferences;
+
```
oblodgett, line 30: If this isn't going to come out of curation. We don't need the field?

oblodgett:
Plus this needs to be a release branch

src/main/java/org/alliancegenome/curation_api/model/document/es/VariantSummaryDocument.java
cmpich: Sorry that needs to be there.

PR #2733, opened 2026-05-13
Add reference fuzzy search indexes

christabone:
## Summary
- add GIN `pg_trgm` indexes for reference fuzzy-search fields used by AI curation validation helpers
- cover reference short citations, AGRKB/reference curies, external reference curies, and cross-reference display names
- keep indexes partial where the validation lookup filters public/current curation records

## Validation
- `git diff --check origin/alpha..HEAD`
- pre-commit secret scan: Gitleaks passed
- pre-commit secret scan: TruffleHog passed


oblodgett:
@christabone we can bring this file in, but its not going to get run. Because the version is way behind the current version.

christabone:
@oblodgett Will it run eventually or do I need to change something? What do I need to do to get this into the curation database as a migration?

oblodgett:
It needs to be the latest version, im not sure what is but you should be able to see it in that dir... my guess is 49.0?

oblodgett (approved):

oblodgett:
Integration tests are failing:
```Caused by: org.flywaydb.core.api.FlywayException: Found more than one migration with version 0.47.0.3```

PR #2745, opened 2026-05-22
SCRUM-6081 Replace Species assembly free text with GenomeAssembly dropdown

VarunReddy1111:
## Summary
- Replace `assembly_curie` free text field on Species with a `@ManyToOne` relationship to `GenomeAssembly`
- Species table Assembly column now uses a taxon-filtered dropdown editor (matching Data Provider pattern)
- DB migration populates the new FK from existing `assembly_curie` data, then drops the old column
- Update `SpeciesValidator` to validate GenomeAssembly by DB lookup
- Update `AlleleDAO` native query to join GenomeAssembly instead of reading assembly_curie

## Test plan
- [ ] Verify Species table shows assembly values (primaryExternalId) for all species
- [ ] Click Edit on a species row — Assembly dropdown should show only assemblies matching that species' taxon
- [ ] Select a different assembly from dropdown and save — should persist correctly
- [ ] Clear the assembly and save — should allow null assembly
- [ ] Verify assembly filter works in the Species table
- [ ] Verify Allele summary documents still include assembly info


src/main/java/org/alliancegenome/curation_api/model/entities/Species.java
```diff
@@ -104,10 +105,17 @@ public class Species extends AuditedObject {
 	@JsonView({CurationView.FieldsOnly.class, CurationView.ForPublic.class, CurationView.GeneExpressionDocument.class, CurationView.TransgenicAllelesDocument.class})
 	private Integer phylogeneticOrder;
 
-	//@FullTextField(analyzer = "autocompleteAnalyzer", searchAnalyzer = "autocompleteSearchAnalyzer")
-	//@KeywordField(name = "assembly_keyword", aggregable = Aggregable.YES, sortable = Sortable.YES, searchable = Searchable.YES, normalizer = "sortNormalizer")
+	// //@FullTextField(analyzer = "autocompleteAnalyzer", searchAnalyzer = "autocompleteSearchAnalyzer")
```
cmpich: These comments could be removed, right?

oblodgett: I agree these comments can be removed

oblodgett (approved):

PR #2746, opened 2026-05-22
release/v0.49.3 — SCRUM-6093 fall back to AGRKB curie when no PMID/PubMod cross-reference

cmpich:
## Summary
- Cherry-picks SCRUM-6093 onto a hotfix branch off `production`.
- `Reference.getReferenceID(boolean)` now returns the reference's own AGRKB `curie` as a final fallback when no PMID and no MOD-prefixed cross-reference is present, instead of `null` (empty crossRefs) or the sorted-first cross-reference.

New display priority for both `getReferenceID()` and `getPubModID()`:
1. PMID (only `getReferenceID()`)
2. PubMod ID (MOD-prefixed: FB / MGI / RGD / SGD / WB / XB / ZFIN)
3. The reference's own AGRKB curie

## Test plan
- [ ] Build agr_curation locally with `-Dgpg.skip=true`
- [ ] Verify a Reference with empty `crossReferences` serializes `referenceID` / `pubModID` as its AGRKB curie (was `null`)
- [ ] Verify a Reference with only non-prefix cross-references (e.g. OMIM only) returns its AGRKB curie instead of the OMIM cross-ref
- [ ] Confirm references with PMID still return PMID; references with MOD-prefixed cross-refs still return those

### DIFF
--- a/src/main/resources/db/migration/v0.49.3.1__populate_sgd_agm_dataprovider_crossreferences.sql
+++ b/src/main/resources/db/migration/v0.49.3.1__populate_sgd_agm_dataprovider_crossreferences.sql
@@ -1,0 +1,36 @@
+-- SCRUM-5124: backfill dataProviderCrossReference for the 7 SGD strain AGMs.
+-- Their original load skipped dataprovidercrossreference_id, so the
+-- model_search_result indexer emits docs without modCrossRefCompleteUrl
+-- and the UI shows yeast strains without a link out to SGD.
+-- Creates one crossreference row per AGM pointing at the SGD "strain"
+-- resourcedescriptorpage, then links it back via biologicalentity.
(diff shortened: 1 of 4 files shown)


src/main/resources/db/migration/v0.49.3.1__populate_sgd_agm_dataprovider_crossreferences.sql
```diff
@@ -0,0 +1,36 @@
+-- SCRUM-5124: backfill dataProviderCrossReference for the 7 SGD strain AGMs.
```
oblodgett, line 1: This migration is going to conflict with varun's we need to talk about it at DS.

oblodgett (approved):

PR #2762, opened 2026-06-09
Comments in the ticket fixed

kthorat-prog:


oblodgett (approved):

src/main/cliapp/src/containers/alleleGeneAssociationsPage/AlleleGeneAssociationsTable.jsx
```diff
@@ -103,7 +149,7 @@ export const AlleleGeneAssociationsTable = () => {
 	const defaultFilters = { obsoleteFilter: { obsolete: { queryString: 'false' } } };
 
 	const initialTableState = useMemo(
-		() => getDefaultTableState('AlleleGeneAssociations', columns, DEFAULT_COLUMN_WIDTH, defaultFilters),
+		() => getDefaultTableState('AlleleGeneAssociations_v2', columns, DEFAULT_COLUMN_WIDTH, defaultFilters),
```
cmpich: Where does the _v2 come from?

kthorat-prog: The _v2 was added intentionally during this session. The table state key (AlleleGeneAssociations_v2) is used as a cache key in localStorage and server-side user settings. When you had cached settings from   
  before the new columns (Taxon, Notes, Evidence, etc.) were added, those cached settings didn't include the new columns in selectedColumnNames, so they were invisible.                                         
                                                                                                                                                                                                                 
  Bumping to _v2 forces all users to get a fresh default state (with all 14 columns in the correct order) instead of loading the stale cache. It's the standard way in this codebase to reset table settings for 
  all users when columns are added or reordered.

oblodgett: Ahh... this is not what we would want to do. We just need to reset the table in order to clear the cache.

kthorat-prog: I just fixed it and pushed the new changes. Please check "src/main/cliapp/src/containers/alleleGeneAssociationsPage/AlleleGeneAssociationsTable.jsx"

oblodgett (approved):

PR #2777, opened 2026-06-15
Scrum 6080

cmpich:


src/main/java/org/alliancegenome/curation_api/jobs/executors/gff/Gff3CDSExecutor.java
```diff
@@ -78,8 +77,8 @@ public void execLoad(BulkLoadFileHistory bulkLoadFileHistory) throws Exception {
 
 		String assemblyId = loadGenomeAssemblyFromGFF(gffHeaderData);
 
-		if (assemblyId == null) {
-			addException(bulkLoadFileHistory, new ObjectUpdateExceptionData(null, "GFF Header does not contain assembly", null));
+		if (!validateGffAssemblyMatchesSpecies(bulkLoadFileHistory, dataProvider, assemblyId)) {
+			return;
```
oblodgett, line 81: So in addException we have logic higher up that handles all the failLoad issues. I would not create a custom version of that. I would addException and then return. validateGffAssemblyMatchesSpecies return an error or something like that, but don't call failLoad directly.

oblodgett (changes requested):

oblodgett:
**`failLoad` should gather the exception, not re-derive its message**

This method is the whole-file abort handler, so its job is the same as everywhere else in this class: hand the structured exception data to `addException`. `runLoad` already does exactly that for this exception type:

```java
} catch (ObjectUpdateException e) {
    addException(history, e.getData());
}
```

The new `addException(...)` call here is the right move. But the `errorMessage` string-building block above it (the `getMessage()`/`getLocalizedMessage()`/cause-walk flattening) is the *reason* the bugs being fixed in this PR exist — the `"null"` recording and the broken `while (e.getCause() != null)` loop both come from trying to reconstruct a message out of the `Throwable` instead of trusting `ObjectUpdateException.getData()`, which already carries it.

So rather than patching the extraction logic, I'd lean on the structured data as the single source:

```java
protected void failLoad(BulkLoadFileHistory history, Exception e) {
    ObjectUpdateExceptionData data = (e instanceof ObjectUpdateException oue && oue.getData() != null)
        ? oue.getData()
        : new ObjectUpdateExceptionData(null, e.getMessage(), e.getStackTrace());
    addException(history, data);
    history.setBulkloadStatus(JobStatus.FAILED);
    updateHistory(history);
}
```

One thing to confirm before dropping the `errorMessage` assignment entirely: `errorMessage` drives the single banner on the history row, while `addException` populates the History exception table — they're separate in the UI. If curators rely on that banner, keep setting it, but derive it from `data` in one place rather than re-extracting from the `Throwable` with its own bug surface.

PR #2778, opened 2026-06-15
Add Team/Project tags to eb create environment

christabone:
## What
Adds `--tags Team=a-team,Project=curation-app` to the `eb create` target so the Elastic Beanstalk environment is tagged at creation.

## Why
These tags drive the `alliance-genome/agr_cost_analysis` tag enforcer (cost attribution by Team + ownership routing). The running resources were just tagged live (env `update-tags-for-resource` + instance tags), but `eb create` passed no `--tags`, so a future environment **recreate** would come up untagged again. This codifies it in IaC.

## Note
`eb create --tags` only applies at environment **creation**, so this does not change the currently-running environment (already tagged out-of-band) — it ensures future recreations carry the tags.


Makefile
```diff
@@ -97,7 +97,7 @@ eb-init:
 	eb init --region us-east-1 -p Docker curation-app
 
 eb-create:
-	@eb create ${ENV_NAME} --region=us-east-1 --cname="${ENV_NAME}" --elb-type application --envvars \
+	@eb create ${ENV_NAME} --region=us-east-1 --cname="${ENV_NAME}" --elb-type application --tags Team=a-team,Project=curation-app --envvars \
```
oblodgett, line 100: This command does not get run.

PR #2780, opened 2026-06-16
Scrum 6161

kthorat-prog:


src/main/java/org/alliancegenome/curation_api/model/entities/EvidenceAssociation.java
```diff
@@ -30,6 +33,7 @@ public class EvidenceAssociation extends Association {
 	@IndexedEmbedded(includeDepth = 2)
 	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
 	@ManyToMany
+	@Fetch(FetchMode.JOIN)
```
oblodgett, line 36: Typically we don't want to do this as it causes issues on the backend. Do you know the reason?

kthorat-prog: EvidenceAssociation.evidence is a @ManyToMany collection, which Hibernate fetches lazily by default — meaning it only loads the collection when something accesses it, not when the parent entity is loaded.   
                                                                                                                                                                                                                 
  The problem: BaseSQLDAO.searchByParams is not @Transactional. It uses Hibernate Search to find matching IDs in OpenSearch, then loads the entities via result.hits(). Once those entities are returned, the    
  Hibernate session is effectively done. When Jackson then tries to serialize the response, the evidence collection is still an uninitialized lazy proxy — the session is gone, so it can never be loaded.       
  Result: evidence serializes as null and the column shows nothing.                                                                                                                                              
                                                                                                                                                                                                                 
  @Fetch(FetchMode.JOIN) forces Hibernate to load evidence eagerly via a JOIN at the moment the entity itself is loaded (inside result.hits(), while the session is still active). By the time Jackson           
  serializes, the collection is already populated.                                                                                                                                                               
                                                                                                                                                                                                                 
  The alternative would have been to make searchByParams @Transactional, but that's a much wider change affecting every entity that uses it.

oblodgett (approved):

PR #2781, opened 2026-06-16
SCRUM-6080: fail GFF file load when header assembly != Species official assembly

cmpich:
## SCRUM-6080

On the FMS GFF file load, validate the header's `#!assembly` against the official assembly designated in the **Species table** (`Species.genomeAssembly`, curator-editable). The load **fails (imports nothing)** when:

1. the header has no `#!assembly` line,
2. no official assembly is designated for the provider's taxon, or
3. the header assembly does not match the official one.

**Strict match:** any non-matching assembly — new *or* old — fails. No auto-establishment of new assemblies (per product decision; upgrades happen by a curator updating the Species table).

### Changes
- `Gff3Executor.validateGffAssemblyMatchesSpecies(history, dataProvider, headerAssembly)` — centralizes the check; fails via `failLoad`.
- `SpeciesService.getByTaxonCurie(taxonCurie)` — SQL lookup of the species / official assembly.
- All four GFF executors call it in `execLoad` right after parsing `#!assembly`, returning (importing nothing) on failure.

### Scope
Validation is applied to the **FMS `execLoad` path only**; the per-entity bulk API endpoints (`runLoadApi`: `/api/transcript/bulk/...`, etc.) are **not** gated — the ticket targets GFF *file* loads, and gating the API path would break the pre-existing `IT_0608` loader tests (see below).

### Testing
The IT environment seeds **no species** (`v0.29.0.2__add_species.sql` inserts species only when the NCBITaxon term already exists, which it doesn't at Flyway-migrate time), so the strict-match path can't be exercised in IT without creating a species + assembly + official designation in-test (no `createSpecies` helper exists). Per decision, `IT_0608` is left at its original passing tests and the strict-match path is **not** covered by IT. Follow-up options: add an IT `createSpecies` helper, or a unit test of `validateGffAssemblyMatchesSpecies` with a mocked `SpeciesService`. `mvn compile` passes.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/jobs/executors/gff/Gff3Executor.java
+++ b/src/main/java/org/alliancegenome/curation_api/jobs/executors/gff/Gff3Executor.java
@@ -3,8 +3,7 @@ package org.alliancegenome.curation_api.jobs.executors.gff;
 import java.util.List;
 
 import org.alliancegenome.curation_api.enums.BackendBulkDataProvider;
-import org.alliancegenome.curation_api.exceptions.ObjectUpdateException.ObjectUpdateExceptionData;
-import org.alliancegenome.curation_api.exceptions.ValidationException;
+import org.alliancegenome.curation_api.exceptions.ObjectUpdateException;
 import org.alliancegenome.curation_api.jobs.executors.LoadFileExecutor;
 import org.alliancegenome.curation_api.model.entities.GenomeAssembly;
 import org.alliancegenome.curation_api.model.entities.Species;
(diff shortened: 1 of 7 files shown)


src/main/java/org/alliancegenome/curation_api/jobs/executors/gff/Gff3Executor.java
```diff
@@ -23,5 +30,45 @@ public String loadGenomeAssemblyFromGFF(List<String> gffHeaderData) throws Valid
...
+	/**
+	 * SCRUM-6080: a GFF load may only proceed if the assembly declared in the file
+	 * header matches the official assembly designated for the species in the Species
+	 * table ({@link Species#getGenomeAssembly()}). On any mismatch the whole load is
+	 * failed and no records are imported.
+	 *
+	 * @return true if the load may proceed; false if the load was failed (the caller
+	 *         must return without importing).
+	 */
+	protected boolean validateGffAssemblyMatchesSpecies(BulkLoadFileHistory history, BackendBulkDataProvider dataProvider, String headerAssembly) {
```
oblodgett: The point of our prior conversation today was to remove this method, or not use "failLoad" but rather use addException

oblodgett (approved):

PR #2787, opened 2026-06-17
SCRUM-6199: add references to transgenic_allele_summary documents

adamgibs:
Expose allele.references on the transgenic allele document so the
  transgenic_allele_summary ES docs carry paper associations for the
  reference page.

  - Add the TransgenicAllelesDocument JSON view to Allele.references, CurieObject.curie, and Reference.crossReferences/shortCitation/ getReferenceID/getPubModID so the reference shape serializes.
  - Pre-load references in AlleleDocumentController.findDocuments() via Hibernate.initialize (mirrors VariantDocumentController); the nested allele.references then rides through the indexer to ES.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/controllers/document/AlleleDocumentController.java
+++ b/src/main/java/org/alliancegenome/curation_api/controllers/document/AlleleDocumentController.java
@@ -64,6 +64,12 @@ public class AlleleDocumentController implements AlleleDocumentInterface {
 		if (response.getResults() != null) {
 			TransgenicAlleleDocumentBuilder builder = new TransgenicAlleleDocumentBuilder();
 			for (AlleleConstructAssociation association : response.getResults()) {
+				// Force-initialize the allele's references (size() loads the lazy collection) so they
+				// serialize onto the transgenic document. The deeper graph (crossReferences, resource
+				// descriptor pages) lazy-loads during JSON serialization, which depends on the request-scoped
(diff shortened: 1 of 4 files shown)


src/main/java/org/alliancegenome/curation_api/controllers/document/AlleleDocumentController.java
```diff
@@ -64,6 +65,11 @@ public SearchResponse<TransgenicAlleleDocument> findDocuments(Integer page, Inte
 		if (response.getResults() != null) {
 			TransgenicAlleleDocumentBuilder builder = new TransgenicAlleleDocumentBuilder();
 			for (AlleleConstructAssociation association : response.getResults()) {
+				// Pre-load the allele's references so they serialize onto the transgenic document. The
+				// deeper graph (crossReferences, resource descriptor pages) lazy-loads during JSON
+				// serialization, which depends on the request-scoped session staying open - do not make
+				// this method @Transactional or those fields will hit a LazyInitializationException.
+				Hibernate.initialize(association.getAlleleAssociationSubject().getReferences());
```
oblodgett: So typically you can `association.getAlleleAssociationSubject().getReferences().size();` and that will do the same thing without depending on hibernate specific code.

adamgibs: @oblodgett change has been made

oblodgett (approved):

PR #2795, opened 2026-06-25
SCRUM-6236 support chromosome RefSeq for GRCz12tu data in variants

cmpich:


VarunReddy1111 (approved):

oblodgett (commented):
Ack i didn't know about this enum, we need to get this in the database also

PR #2796, opened 2026-06-26
SCRUM-6220: fix and narrow disease-annotation autocompletes

adamgibs:
Fix the unresponsive/erroring autosuggest fields on the New Disease Annotation form (and the inline table editors), and constrain entity autocompletes to the curator's MOD species.

Unknown-field / OOM fixes:
- Make biologicalEntityAutocompleteConfig bridged-only (curie, primaryExternalId, modInternalId, crossReferences.*, and the BiologicalEntityTypeBridge flat fields name/symbol/synonyms/ secondaryIds). Per-type slot paths (geneSymbol.*, geneSystematicName.*, alleleSymbol.*, agm*.*) threw Hibernate Search "Unknown field" on the single-type /gene, /allele, /agm indexes, so allele/AGM lookups returned nothing.
- Add geneAutocompleteConfig (bridged + geneSystematicName) for the Gene-only fields; drop the old diseaseAnnotationSubjectAutocompleteConfig.
- Add ontologyCrossReferencesFieldSet (depth-1) for ontology autocompletes: the depth-2 crossReferences.resourceDescriptorPage.name path is not indexed on /doterm (includeDepth=1) and threw "Unknown field".
- Override /biologicalentity search() to @JsonView(FieldsOnly), matching the per-type search endpoints. FieldsAndLists serialized association collections with incomplete cross-subtype cycle-guards, causing an OOM.

Species/MOD narrowing:
- Add buildCuratorSpeciesFilter(): a hard taxon.curie constraint derived from the curator's cognito *Staff groups, AND'd against the typed query.
- Derive the MOD->taxa map at runtime from the Species table
- Apply to Subject, genetic-modifier (AGM/Allele/Gene) and asserted (Gene/Allele) fields, in both the form and the inline table editors. The "with" field stays human-pinned; Disease is excluded (no taxon).

SGD strain background: match by taxon.curie=NCBITaxon:559292 (keyword) instead of the brittle taxon.name full-text filter.

Add unit tests for the autocomplete configs and the species filter.

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/interfaces/crud/BiologicalEntityCrudInterface.java
+++ b/src/main/java/org/alliancegenome/curation_api/interfaces/crud/BiologicalEntityCrudInterface.java
@@ -1,12 +1,21 @@
 package org.alliancegenome.curation_api.interfaces.crud;
 
+import java.util.HashMap;
+
 import org.alliancegenome.curation_api.interfaces.base.BaseSubmittedObjectCrudInterface;
 import org.alliancegenome.curation_api.model.entities.BiologicalEntity;
+import org.alliancegenome.curation_api.response.SearchResponse;
+import org.alliancegenome.curation_api.view.CurationView;
 import org.eclipse.microprofile.openapi.annotations.tags.Tag;
 
+import com.fasterxml.jackson.annotation.JsonView;
+
 import jakarta.ws.rs.Consumes;
+import jakarta.ws.rs.DefaultValue;
+import jakarta.ws.rs.POST;
 import jakarta.ws.rs.Path;
 import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.QueryParam;
 import jakarta.ws.rs.core.MediaType;
 
 @Path("/biologicalentity")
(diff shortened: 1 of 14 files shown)


src/main/java/org/alliancegenome/curation_api/interfaces/crud/BiologicalEntityCrudInterface.java
```diff
@@ -15,4 +24,16 @@
...
+	// matching the per-type search endpoints (/gene, /allele, /agm, ...). On this
+	// cross-subtype index FieldsAndLists serializes association collections whose
+	// cycle-guards are incomplete across subtypes, producing a >2GB serialization /
+	// OutOfMemoryError. Callers needing associations use get-by-id.
+	@Override
+	@POST
+	@Path("/search")
+	@Tag(name = "Elastic Search Browsing Endpoints")
+	@JsonView({ CurationView.FieldsOnly.class })
+	SearchResponse<BiologicalEntity> search(@DefaultValue("0") @QueryParam("page") Integer page, @DefaultValue("10") @QueryParam("limit") Integer limit, HashMap<String, Object> params);
```
oblodgett, line 37: @adamgibs doesn't this method already exist?

adamgibs: @oblodgett It inherits a search method, but this is overriding it to switch the view. It's doing the same the gene, allele, and agm CrudInterfaces do.

oblodgett: But isn't the default view "FieldsOnly"

adamgibs: @oblodgett No, the default in the `BaseSearchControllerInterface `is `FieldsAndLists`.

oblodgett: ahh ok... so you were getting a stackoverflow I assume?

oblodgett (approved):

src/main/java/org/alliancegenome/curation_api/interfaces/crud/BiologicalEntityCrudInterface.java
adamgibs: @oblodgett Yep, exactly. All of the associations were coming along for the ride in the autocomplete

PR #2808, opened 2026-07-08
Scrum 6234

VarunReddy1111:


src/main/resources/db/migration/v0.52.0.1__add_antibodies.sql
```diff
@@ -0,0 +1,135 @@
+CREATE TABLE antibody (
```
oblodgett, line 1: I think this file needs to change adam put in a 51.0.1 into production yesterday?

oblodgett (approved):

PR #2814, opened 2026-07-15
SCRUM-6204: add typed findAllForPublic on ResourceDescriptorPageCrudI…

adamgibs:
…nterface

Consumed by agr_indexer's ResourceDescriptorService. The inherited findForPublic returns raw jakarta.ws.rs.core.Response which RESCU can't deserialize. New method returns typed SearchResponse under CurationView.ForPublic, so all prefix + synonym mappings resolve correctly.


oblodgett (commented):
Doesn't these endpoints already exist? They should, maybe they are not giving you the correct view?

adamgibs:
@oblodgett the base findForPublic returns Response we need a SearchResponse object returned

oblodgett:
Why does that matter? They both have the same fields SearchResponce ?

PR #2816, opened 2026-07-17
Release/v0.51.3

VarunReddy1111:


src/main/java/org/alliancegenome/curation_api/model/document/builders/GODocumentBuilder.java
```diff
@@ -45,7 +45,7 @@ public static GOSearchResultDocument buildSearchResultDocument(GOTerm goTerm, Re
 				String speciesAbbreviation = gene.getTaxon().getSpecies().getAbbreviation();
 				String geneDisplayString = geneSymbol + " (" + speciesAbbreviation + ")";
 
-				String taxonName = gene.getTaxon().getName();
+				String taxonName = gene.getTaxon().getSpecies().getFullName();
```
oblodgett: You may want to do the null check here also.

oblodgett (approved):

PR #2820, opened 2026-07-20
SCRUM-6268 fix cascade delete of shared entities referenced by HTPExpressionDatasetAnnotation

kthorat-prog:
## Summary
- Trim `HTPExpressionDatasetAnnotation.htpExpressionDataset` cascade from `ALL`+`orphanRemoval` to `PERSIST`/`MERGE`, since the linked `ExternalDataBaseEntity` is a shared lookup row (found-or-created by curie) that other annotations may still reference.
- Add `@OnDelete(action = OnDeleteAction.CASCADE)` to the `categoryTags`/`reference` `@ManyToMany` join fields, and add a matching migration to cascade the join-table FK constraints at the DB level.
- Net effect: deleting an annotation removes its own join rows but leaves shared datasets/tags/references intact for other annotations.

## Test plan
- [ ] Delete an `HTPExpressionDatasetAnnotation` with populated `categoryTags`/`reference`/`subSeries` and confirm no FK violation
- [ ] Confirm the shared `ExternalDataBaseEntity` dataset row survives deletion of one annotation when referenced elsewhere
- [ ] Run migration `v0.52.0.2` against a local DB

### DIFF
--- a/src/main/java/org/alliancegenome/curation_api/model/entities/HTPExpressionDatasetAnnotation.java
+++ b/src/main/java/org/alliancegenome/curation_api/model/entities/HTPExpressionDatasetAnnotation.java
@@ -55,13 +55,17 @@ public class HTPExpressionDatasetAnnotation extends AuditedObject {
 
 	@IndexedEmbedded(includeDepth = 1)
 	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
-	@OneToOne(cascade = CascadeType.ALL, orphanRemoval = true)
+	// SCRUM-6268: no cascade REMOVE/orphanRemoval - this is a shared lookup row (found/created by curie,
+	// see ExternalDataBaseEntityFmsDTOValidator) that HTPExpressionDatasetSampleAnnotation.datasetIds may
+	// still reference independently, so deleting this annotation must not cascade-delete it.
+	@OneToOne(cascade = { CascadeType.PERSIST, CascadeType.MERGE })
 	@JsonView({ CurationView.FieldsOnly.class })
 	private ExternalDataBaseEntity htpExpressionDataset;
 
 	@IndexedEmbedded(includeDepth = 1)
 	@IndexingDependency(reindexOnUpdate = ReindexOnUpdate.SHALLOW)
 	@ManyToMany
+	@OnDelete(action = OnDeleteAction.CASCADE)
 	@Fetch(FetchMode.JOIN)
 	@JoinTable(indexes = {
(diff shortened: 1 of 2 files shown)


kthorat-prog:
Two related problems, two fixes:                                                                                                                                                                               
                                                                                                                                                                                                                 
  1. htpExpressionDataset field (the @OneToOne to ExternalDataBaseEntity)                                                                                                                                        
  // before                                                                                                                                                                                                      
  @OneToOne(cascade = CascadeType.ALL, orphanRemoval = true)                                                                                                                                                     
  private ExternalDataBaseEntity htpExpressionDataset;                                                                                                                                                           
                                                                                                                                                                                                                 
  // after                                                                                                                                                                                                       
  @OneToOne(cascade = { CascadeType.PERSIST, CascadeType.MERGE })                                                                                                                                                
  private ExternalDataBaseEntity htpExpressionDataset;                                                                                                                                                           
  CascadeType.ALL + orphanRemoval = true meant deleting an annotation would delete its linked ExternalDataBaseEntity row too. But that ExternalDataBaseEntity is a shared lookup row — it's found-or-created by  
  curie (see ExternalDataBaseEntityFmsDTOValidator), so multiple HTPExpressionDatasetSampleAnnotation.datasetIds can point at the same row independently. Deleting one annotation would rip out a row other      
  records still reference. Trimming cascade to just PERSIST/MERGE keeps saves working but stops delete from propagating.                                                                                         
                                                                                                                                                                                                                 
  2. The @ManyToMany join tables (categoryTags, references, externalDatabaseEntity cross-refs)                                                                                                                   
                                                                                                                                                                                                                 
  These join tables had FK constraints back to htpexpressiondatasetannotation without ON DELETE CASCADE in the actual database schema — even though the Java entity's cascade annotations implied the join rows  
  should go away with the parent. Since categoryTags is required on every annotation, deleting any annotation that had tags/references/subSeries populated would throw a FK violation.                           
                                                                                                                                                                                                                 
  Added @OnDelete(action = OnDeleteAction.CASCADE) to those @ManyToMany fields, and a matching migration (v0.52.0.2__htp_dataset_annotation_join_fk_cascade.sql) that drops and re-adds the three FK constraints 
  with ON DELETE CASCADE at the DB level:                                                                                                                                                                        
  - htpexpressiondatasetannotation_categorytags                                                                                                                                                                  
  - htpexpressiondatasetannotation_reference                                                                                                                                                                     
  - htpexpressiondatasetannotation_externaldatabaseentity                                                                                                                                                        
                                                                                                                                                                                                                 
  Importantly, this cascade is scoped to just the join rows — it does not cascade to the actual VocabularyTerm, Reference, or ExternalDataBaseEntity targets, which (like the dataset lookup above) are shared   
  and must survive the annotation's deletion.

oblodgett (approved):

